PR #15, opened 2023-08-18
Check if exported log is same as existing log

mimi1vx:
If log and exported date are same skip exporting.


perlpunk (commented):
Can this be tested?

mimi1vx:
Manually yeah, simple unit test also ...

okurz (approved):

perlpunk (approved):

PR #17, opened 2023-09-28
Add some cosmetics to identify the ReviewRequest ID

logsnaath:
1.  Changes mtui prompt to include the review id.  Ex `S:M:xxxxx:yyyyyy/mtui-auto>`. This requires `python311-settproctitle`.
2.  Sets the process name to include the review id. This helps to identify the window/tab of most of the terminal emulators title.

Would like to get more insights and comments.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,10 @@ only-test:
 checkstyle:
 	black --check ./
 
+.PHONY: tidy-diff
+tidy-diff:
+	black --diff ./
+
 .PHONY: tidy
 tidy:
 	black ./
--- a/mtui/main.py
+++ b/mtui/main.py
@@ -1,6 +1,7 @@
 import logging
 import sys
 from subprocess import CalledProcessError
+import setproctitle
 
 from mtui.args import get_parser
 from mtui.config import Config
@@ -11,6 +12,7 @@ from mtui.systemcheck import detect_system
 
 from .argparse import ArgsParseFailure
 from .colorlog import create_logger
+from .utils import get_short_rrid
 
 
 def main():
@@ -33,7 +35,6 @@ def main():
 
 
 def run_mtui(sys, config, logger, Prompt, Display, args):
-
     if args.debug:
         logger.setLevel(level=logging.DEBUG)
 
@@ -45,6 +46,9 @@ def run_mtui(sys, config, logger, Prompt, Display, args):
 
     prompt = Prompt(config, logger, sys, Display)
     if args.update:
+        short_rrid = get_short_rrid(str(args.update.id))
+        # helps to set window/tab title with reviewid
+        setproctitle.setproctitle("mtui-" + short_rrid)
         if args.update.kind == "kernel":
             config.kernel = True
             config.auto = False
--- a/mtui/prompt.py
+++ b/mtui/prompt.py
@@ -14,6 +14,7 @@ import mtui.notification as notification
 from . import commands, messages
 from .argparse import ArgsParseFailure
 from .template.nulltestreport import NullTestReport
+from .utils import get_short_rrid
 
 logger = getLogger("mtui.prompt")
 
@@ -221,16 +222,17 @@ class CommandPrompt(cmd.Cmd):
 
     def set_prompt(self, session=None):
         self.session = session
-        session = ":" + str(session) if session else ""
+        session = str(session) + "/" if session else ""
         mode = "mtui"
         if self.config.auto and not self.config.kernel:
             mode += "-auto"
         elif self.config.kernel:
             mode += "-kernel"
-        self.prompt = f"{mode}{session}> "
+        self.prompt = f"{session}{mode}> "
 
     def load_update(self, update, autoconnect):
         tr = update.make_testreport(self.config, autoconnect=autoconnect)
         self.metadata = tr
         self.targets = tr.targets
-        self.set_prompt(None)
+        short_rrid = get_short_rrid(str(update.id))
+        self.set_prompt(short_rrid)
--- a/mtui/template/testreport.py
+++ b/mtui/template/testreport.py
@@ -85,6 +85,7 @@ class TestReport(metaclass=ABCMeta):
         self.reviewer = ""
         self.repository = None
         self.packages = {}
+        self.srpms = []
 
         self._attrs = [
             "products",
@@ -92,6 +93,7 @@ class TestReport(metaclass=ABCMeta):
             "packager",
             "reviewer",
             "packages",
+            "srpms",
             "bugs",
             "repository",
         ]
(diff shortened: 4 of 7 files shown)


okurz (changes requested):
Hi Logu, Happy to see your contributions. I am tempted to say please add test coverage but I don't know if the surrounding code is already covered by automatic tests at all. Anyway, please take a look. Maybe it's not that hard :) and second there is a duplication which you can avoid by extracting a method

logsnaath:
> Hi Logu, Happy to see your contributions. I am tempted to say please add test coverage but I don't know if the surrounding code is already covered by automatic tests at all. Anyway, please take a look. Maybe it's not that hard :) and second there is a duplication which you can avoid by extracting a method

I dont know which duplication you have mentioned, is it leveraging subprocess instead of setproctitle ?

mtui/main.py
```diff
@@ -45,6 +45,11 @@ def run_mtui(sys, config, logger, Prompt, Display, args):
 
     prompt = Prompt(config, logger, sys, Display)
     if args.update:
+        short_rrid = (
+            str(args.update.id).replace("SUSE", "S").replace("Maintenance", "M")
```
okurz, line 49: this is duplicating ...

mtui/prompt.py
```diff
@@ -221,16 +221,17 @@ def emptyline(self):
...
             mode += "-kernel"
-        self.prompt = f"{mode}{session}> "
+        self.prompt = f"{session}{mode}> "
 
     def load_update(self, update, autoconnect):
         tr = update.make_testreport(self.config, autoconnect=autoconnect)
         self.metadata = tr
         self.targets = tr.targets
-        self.set_prompt(None)
+        short_rrid = str(update.id).replace("SUSE", "S").replace("Maintenance", "M")
```
okurz, line 236: ... this

Makefile
```diff
@@ -9,6 +9,10 @@ only-test:
 checkstyle:
 	black --check ./
 
+.PHONY: tidy-diff
+tidy-diff:
+	black --diff ./
+
```
okurz, line 15: valid extension. But please make that a separate commit

mtui/template/testreport.py
```diff
@@ -430,6 +431,14 @@ def refhosts_from_tp(self, testplatform):
     def list_bugs(self, sink, arg):
         return sink(self.bugs, self.jira, arg)
 
+    def _get_result_links(self):
+        incident_id = self.repository.rstrip("/").split("/")[-1]
+        dboard_burl = "http://dashboard.qam.suse.de/incident/"
+        openqa_burl = "https://openqa.suse.de/tests/overview?build=:"
```
okurz, line 437: Please avoid specifying any SUSE internal URLs that are not relevant to this "upstream" open source project

mtui/template/testreport.py
```diff
@@ -443,6 +452,8 @@ def _show_yourself_data(self):
                 ("Build checks", self._testreport_url()[:-3] + "build_checks"),
                 ("Testreport", self._testreport_url()),
                 ("Repository", self.repository),
+                ("DashBoard Link", self._get_result_links()[0]),
+                ("OpenQA Link", self._get_result_links()[1]),
```
okurz, line 456:
```suggestion
                ("openQA Link", self._get_result_links()[1]),
```

mtui/template/testreport.py
```diff
@@ -443,6 +452,8 @@ def _show_yourself_data(self):
                 ("Build checks", self._testreport_url()[:-3] + "build_checks"),
                 ("Testreport", self._testreport_url()),
                 ("Repository", self.repository),
+                ("DashBoard Link", self._get_result_links()[0]),
```
okurz, line 455:
```suggestion
                ("Dashboard Link", self._get_result_links()[0]),
```

okurz (changes requested):

mtui/template/testreport.py
Martchus: I would also lowercase "Link" here, as well as in the line above. If this ends up rendered as link then maybe also drop " Link" completely. The other entries also don't have " Link".

mtui/template/testreport.py
logsnaath: So should I avoid this change entirely?
or get the base urls from some other ways like the user environment?
or have them in mtui/config.py where a few simillar urls are already defined?

okurz: reading from user environment or mtu/config.py both sounds good. I would just avoid specifying any SUSE internal URLs but for convenience maybe it makes sense to provide them but then please in a specific config file. https://github.com/openSUSE/mtui/blob/main/mtui/config.py#L114 looks similar so I guess it's ok if you define URLs there.

mtui/main.py
```diff
@@ -1,6 +1,7 @@
 import logging
 import sys
 from subprocess import CalledProcessError
+import setproctitle
```
mimi1vx, line 4: oh, nonexisting module

mtui/main.py
```diff
@@ -11,6 +12,7 @@
 
 from .argparse import ArgsParseFailure
 from .colorlog import create_logger
+from .utils import get_short_rrid
```
mimi1vx, line 15: why special function if this can be method of RequestReviewID class ?

mtui/main.py
```diff
@@ -45,6 +46,9 @@ def run_mtui(sys, config, logger, Prompt, Display, args):
 
     prompt = Prompt(config, logger, sys, Display)
     if args.update:
+        short_rrid = get_short_rrid(str(args.update.id))
+        # helps to set window/tab title with reviewid
+        setproctitle.setproctitle("mtui-" + short_rrid)
```
mimi1vx, line 51: btw how will `load_template` command change title ?

mtui/template/testreport.py
```diff
@@ -430,6 +431,14 @@ def refhosts_from_tp(self, testplatform):
     def list_bugs(self, sink, arg):
         return sink(self.bugs, self.jira, arg)
 
+    def _get_result_links(self):
+        incident_id = self.repository.rstrip("/").split("/")[-1]
+        dboard_burl = "http://dashboard.qam.suse.de/incident/"
+        openqa_burl = "https://openqa.suse.de/tests/overview?build=:"
+        dboard_lnk = dboard_burl + incident_id
+        openqa_lnk = openqa_burl + incident_id + ":" + self.srpms[0]
```
mimi1vx, line 439: rly ? this can't work

mtui/utils.py
```diff
@@ -333,3 +333,7 @@ def walk(inc):
             if isinstance(inc[key], (list, dict)):
                 inc[key] = walk(inc[key])
     return inc
+
+
+def get_short_rrid(id_str):
```
mimi1vx, line 338: why ? this should be method of RequestReviewID ...

mimi1vx:
I don't think this will get any update in future

PR #21, opened 2025-02-07
Set devel version to 13.4.0

mimi1vx:


setup.py
```diff
@@ -24,12 +24,13 @@
     maintainer="SUSE QA Maintenance",
     maintainer_email="qa-maintenance@suse.de",
     license="License :: Other/Proprietary License",
-    platforms=["Linux", "Mac OSX"],
```
okurz, line 27: Why remove this?

mimi1vx: Because it really never worked on MacOS 😃 
Dependency on rpm-python can't be really solved as on pypi is only placeholder empty package and real bindings depend on rpm distributed in Linux distributions and none of providers on mac doesn't deliver it

baierjan (approved):

Martchus (approved):

r-richardson (approved):

PR #28, opened 2025-03-10
Remove unneeded comparsions between targets

mimi1vx:
Not used + System class doesnt support rich comparsions so it cant really work

+ removed method wrapper around parse_host

### DIFF
--- a/mtui/target/target.py
+++ b/mtui/target/target.py
@@ -34,12 +34,6 @@ class Target:
         lock=TargetLock,
         connection=Connection,
     ) -> None:
-        """
-        :type connect: bool
-        :param connect:
-            introduced in order to run unit tests witout
-            having the target automatically connect
-        """
 
         self.config = config
         self.host, _, self.port = hostname.partition(":")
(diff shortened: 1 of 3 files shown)


mtui/target/target.py
```diff
@@ -94,17 +78,11 @@ def connect(self) -> None:
             logger.warning(self._lock.locked_by_msg())
 
         # get system
-        self.system = self._parse_system()
+        self.system = parse_system(self.connection)
```
baierjan, line 81: and this the misspelled `self.connection` I mentioned above or `self.Connection`?

okurz:
s/unneded comparsions/unneeded comparisons/

mtui/target/target.py
```diff
@@ -51,15 +45,9 @@ def __init__(
...
-        """
-        :param state:
-        :type state: str either "enabled" or "disabled"
-        :deprecated:
-        """
         self.timeout = timeout
         self.exclusive = exclusive
-        self.connction = None
-
+        self.connction: Connection
```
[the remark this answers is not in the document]
mimi1vx: one holds type , second instance ...

`Typer[Connection]` vs `Connection` .. I think original intent was to have possibility to change transport layer with minimal code changes in Target and other abstractions

mtui/target/target.py
mimi1vx: above is misspelled :D nice catch

baierjan (approved):

perlpunk:
s/Remove unneeded comparsions between targets/Remove unneeded comparisons between targets/

perlpunk:
Now the merged commit has `comparisions` :D

PR #33, opened 2025-03-21
New downgrade

mimi1vx:


mtui/actions/downgrade.py
```diff
@@ -0,0 +1,32 @@
+from string import Template
+
+from ..messages import MissingDowngraderError
+from ..utils import DictWithInjections
+
+
+def zypper() -> dict[str, Template]:
+    list_command_template = r"""
+for p in $packages; do \
+zypper -n se -s --match-exact -t package $$p; \
```
baierjan, line 10: `$$p` is that some sort of escaping? I suspect a typo.

mimi1vx: its escape for `$` in `string.Template`  which uses `$keyword` to fill template :D

baierjan: Hm, I see; and I guess the template system is different than in the original code, which explains why it wasn't escaped before. So now it is populated inside https://github.com/openSUSE/mtui/pull/33/files/c2e6656b3bf62a4eff4c79f1b10c04912418eefc#diff-2f7e2205eec4e38ab21036cf0cb2921a7db2c8de6449b1d36ec26a694015a613R240, so this one is OK, but!

mtui/actions/downgrade.py
```diff
@@ -0,0 +1,32 @@
...
+done \
+| grep -v "(System" \
+| grep ^[iv] \
+| sed "s, ,,g" \
+| awk -F "|" '{{ print $2,"=",$4 }}'
```
baierjan, line 15: What about this then, should that be also escaped?
```suggestion
| awk -F "|" '{{ print $$2,"=",$$4 }}'
```

mimi1vx: Not needed as `$123` isn't recognized as template because number cat be keyword

baierjan: Ah, ok then... to much auto-magic 😄

baierjan (approved):
I tried to find something, so at least here is a typo. I guess it is OK otherwise.

mtui/checks/downgrade.py
```diff
@@ -0,0 +1,43 @@
...
+            hostname,
+            stdin,
+            stdout,
+            stderr,
+            exitcode,
+        )
+        raise UpdateError("update stack locked", hostname)
+    if "(c): c" in stdout:
+        logger.critical(
+            "%s: unresolved dependency problem. please resolve manually:\n%s",
```
baierjan, line 27:
```suggestion
            "%s: unresolved dependency problem. Please resolve manually:\n%s",
```

PR #98, opened 2026-04-28
Use QEM Dashboard for auto workflow discovery

mimi1vx:
## Summary
- Replace SMELT and direct openQA auto workflow discovery with QEM Dashboard incident, aggregate, and job data.
- Keep kernel workflow on direct openQA discovery and keep openQA file URLs for log downloads.
- Include dashboard openQA results in manual exports and preserve legacy openQA result block spacing.

## Testing
- `uv run pytest`
- `uv run ruff check mtui/connector/qem_dashboard.py tests/test_qem_dashboard_connector.py`
- `uv run ruff check mtui/export/base.py mtui/commands/export.py tests/test_export_openqa.py`
- `uv run ruff check mtui/commands/simpleset.py mtui/commands/reloadoqa.py tests/test_simpleset.py`


okurz (commented):
I recommend to *not* use openQA less and qem dashboard more. Despite recent development work in QEM Development dashboard the scope of the dashboard is limited and some workflows are not covered there. Also, we are adding more features to openQA that can replace what previously the dashboard was used for.

mimi1vx:
> I recommend to _not_ use openQA less and qem dashboard more. Despite recent development work in QEM Development dashboard the scope of the dashboard is limited and some workflows are not covered there. Also, we are adding more features to openQA that can replace what previously the dashboard was used for.

For now .. everything we need is in dashboard + openQA response is too big for every job ( and wee need data about big bunch of jobs) and last .. openQA is slow and often overburdened by others.

baierjan (approved):
@okurz point is valid; however if I understood the code correctly, this particular change is in fact mainly about "using SMELT less" (because the GraphQL endpoint of SMELT might not be around for much longer).

okurz:
> > I recommend to _not_ use openQA less and qem dashboard more. Despite recent development work in QEM Development dashboard the scope of the dashboard is limited and some workflows are not covered there. Also, we are adding more features to openQA that can replace what previously the dashboard was used for.
> 
> For now .. everything we need is in dashboard + openQA response is too big for every job ( and wee need data about big bunch of jobs) and last .. openQA is slow and often overburdened by others.

Sounds like maybe this access of jobs over the API can be optimized?

mimi1vx:
> Sounds like maybe this access of jobs over the API can be optimized?

Yes, there is big room for improvements :D 

For now this will also remove some of OSD burden as it will not request - same data as qem-dasboard from OSD.

PR #108, opened 2026-05-05
Public-API test coverage for config, prompt, refhost, target, hostgroup and connection

mimi1vx:
## Summary

Adds public-API regression tests covering the previously undertested
modules in `mtui/`. Project test coverage rises from 58 % to 67 %; per
the standing 'current minus one' policy the Codecov project floor is
moved from 56 to 66 (patch target unchanged at 80).

## Per-module coverage

| Module                       | Before | After | Tests added |
| ---------------------------- | ------ | ----- | ----------- |
| `mtui/config.py`             | 88 %   | 99 %  | +10 |
| `mtui/prompt.py`             | 73 %   | 100 % | +21 |
| `mtui/refhost.py`            | 26 %   | 98 %  | +47 |
| `mtui/target/target.py`      | 56 %   | 99 %  | +50 |
| `mtui/target/hostgroup.py`   | 42 %   | 89 %  | +22 |
| `mtui/connection.py`         | 44 %   | 85 %  | +32 |

Total suite: 351 → 532 passing tests. No production code was
modified; this PR is tests + one CHANGELOG entry + one `codecov.yml`
floor bump.

## Behaviour locked in (worth a closer review)

While writing characterization tests for `mtui/config.py` two
pre-existing bugs surfaced. Both are now under regression tests with
inline `FIXME` comments so they are visible to the next person who
touches `Config`:

1. `_parse_config` runs `fixup(val)` outside its `try/except`, so a
   non-numeric value for `connection_timeout` (e.g.
   `connection_timeout = abc` under `[mtui]`) crashes
   `Config.__init__` with `ValueError` instead of falling back to the
   documented default of 300 seconds.
2. `_get_option`'s error-logging path (lines 298–301) calls
   `msg.format((*secopt, self.configfiles))`, which passes a single
   tuple to a format string expecting three positional args. The
   `format()` call itself raises before `logger.error` can run, so a
   bad value for any `getint`/`getboolean` option
   (`refhosts.https_expiration`, `template.smelt_threshold`,
   `mtui.chdir_to_template_dir`, `mtui.use_keyring`) is silently
   replaced by the default with no diagnostic in the logs. The
   default is correct; only the error message is missing.

Both are also documented under `### Known issues` in `CHANGELOG.md`.

A handful of smaller findings are noted inline (no FIXME, just locked
in):

- `Target.sftp_get`'s "trailing slash → folder mode" branch is
  unreachable when callers pass a real `Path` because
  `Path('/x/') == Path('/x')`. Only string callers exercise it.
- `Connection.sftp_open`'s `isinstance(sftp, SFTPClient)` guard makes
  the close-on-failure branch unreachable in any test that mocks the
  SFTP collaborator. The public `log + re-raise` contract is what's
  asserted.

## Validation

- `uv run ruff format --check .`
- `uv run ruff check .`
- `uv run ty check`
- `uv run pytest` (532 passed)
- `uv run --group doc sphinx-build -W -b html Documentation /tmp/x`
- `uv run python -m mtui --help` and `-V`

### DIFF
--- a/tests/test_refhost.py
+++ b/tests/test_refhost.py
@@ -1,6 +1,17 @@
 """Tests for the mtui refhost module - specifically the eval() fix."""
 
+import errno
 import re
+from pathlib import Path
+from types import SimpleNamespace
+from unittest.mock import MagicMock
+
+import pytest
+
+from mtui import refhost
+from mtui.messages import InvalidLocationError
+
+REFHOSTS_FIXTURE = Path(__file__).parent / "fixtures" / "refhosts.yml"
 
 
 class TestArchListParsing:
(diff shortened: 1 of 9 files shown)


tests/test_refhost.py
```diff
@@ -59,3 +70,407 @@ def test_no_brackets(self):
...
+        assert str(attr) == "x86_64"
+
+    def test_str_with_addons_sorted(self):
+        attr = refhost.Attributes()
+        attr.addons = [
+            {"name": "sdk", "version": {"major": 15, "minor": 5}},
+            {"name": "ha", "version": {"major": 15}},
+        ]
+        # Addons are sorted alphabetically.
+        assert str(attr) == "ha 15. sdk 15.5"
```
mjdonis, line 116: Maybe a bit of a nitpick: rendering for products vs addons is inconsistent when no minor version is present. Products render _without_ a dot at the end, e.g. "SLES 15" and you're even testing to make sure you follow that format, but addons _do_ have the dot, e.g. " HA 15." 

Why is the dot always added for the addons but not the products? Is there a reason or should they be following the same pattern?

mimi1vx: its caused by oversight in https://github.com/openSUSE/mtui/blob/35d0811fc83e924ab5a1024e2c7d71825c399f3a/mtui/refhost.py#L57-L64

and yep, it should be without dot when minor version isnt specified on SLE addons before 16

mjdonis (approved):

PR #112, opened 2026-05-09
fix(export): improve qem-dashboard openQA export results

mimi1vx:
## Summary

- Collapses passed qem-dashboard openQA jobs into grouped summary rows while listing only failed, incomplete, and timeout jobs individually with direct openQA links.
- Rewrites the auto-workflow `Install tests:` block from qem-dashboard data so passing install jobs are reported as `PASSED` instead of preserving stale `FAILED` output.
- Carries openQA install-job results through export so install rows render as `=> PASSED:`/`=> SOFTFAILED:` instead of `=> none:`.

## Why

Large maintenance updates can produce exported logs with hundreds of one-line openQA entries, while the install-test section could still be misleading: install logs were exported correctly, but the report showed `Installation tests ... FAILED` and rows like `=> none:`. Reviewers need the export to focus on actionable openQA failures and show the actual install-test status.

## Behaviour

- Aggregate openQA failures remain visible in the separate `Results from openQA jobs:` section; they do not make the install-test block fail.
- The `Install tests:` block is regenerated from install jobs only, replacing stale status and stale per-job rows on re-export.
- Existing openQA block markers remain unchanged, so `mtui/export/base.py::inject_openqa` continues replacing previous qem-dashboard results cleanly.
- Standard auto openQA export also carries install-job results to avoid regressions from the shared `URLs` structure.

## Tests

- Added/updated focused tests for qem-dashboard result propagation, standard auto openQA result propagation, and stale install-block replacement.
- Verified locally: `uv run pytest tests/test_qem_dashboard_connector.py tests/test_export_openqa.py tests/test_openqa_connector.py` (33 passed).
- Verified locally: `uv run ruff check ...` and `uv run ty check` passed.

## Changelog

Existing user-visible qem-dashboard export change remains documented under `[Unreleased] / Changed` in `CHANGELOG.md`.

### DIFF
--- a/mtui/export/auto.py
+++ b/mtui/export/auto.py
@@ -32,7 +32,10 @@ class AutoExport(BaseExport):
         auto = self.openqa.auto
         results = auto.results if auto is not None else None
         if not results:
-            return "FAILED"
+            # No results means the install tests have not run, are still
+            # running, or could not be fetched - which is distinct from a
+            # genuine FAILED outcome.
+            return "UNKNOWN"
         return (
             "PASSED"
             if all(result.result in {"passed", "softfailed"} for result in results)
--- a/mtui/export/auto.py
+++ b/mtui/export/auto.py
@@ -20,6 +20,70 @@ no_verify = ssl._create_unverified_context()  # noqa: SLF001
 class AutoExport(BaseExport):
     """An exporter for the automatic workflow."""
 
+    @staticmethod
+    def _install_job_line(result) -> str:
+        status = result.result.upper() if result.result else "UNKNOWN"
+        job_url = result.url.rsplit("/file/", 1)[0]
+        return (
+            f"{result.distri}_{result.version}_{result.arch} => {status}: {job_url}\n"
+        )
+
+    def _install_status(self) -> str:
+        auto = self.openqa.auto
+        results = auto.results if auto is not None else None
+        if not results:
+            return "FAILED"
+        return (
+            "PASSED"
+            if all(result.result in {"passed", "softfailed"} for result in results)
+            else "FAILED"
+        )
+
+    def install_results(self) -> None:
+        """Adds installation results to the template."""
+        status_line = (
+            "Installation tests done in openQA with following results: "
+            f"{self._install_status()}\n"
+        )
+        auto = self.openqa.auto
+        results = auto.results if auto is not None else []
+        result_lines = [self._install_job_line(result) for result in results or []]
+
+        try:
+            start = self.template.index("Install tests:\n") - 1
+        except ValueError:
+            start = len(self.template)
+            if "## export MTUI:" in self.template[-1]:
+                start -= 1
+            self.template[start:start] = [
+                "##############\n",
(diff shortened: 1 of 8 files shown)


tests/test_qem_dashboard_connector.py
```diff
@@ -157,3 +221,235 @@ def test_dashboard_auto_openqa_loads_incident_and_aggregate_jobs(mock_config):
...
+    jobs = [
+        _aggregate_job(4000, "mau-a", "passed", product="SLES-15-SP5"),
+        _aggregate_job(4001, "mau-b", "passed", product="SLES-15-SP5"),
+        _aggregate_job(4002, "mau-c", "failed", product="SLES-15-SP6"),
+    ]
+    out = "".join(dashboard._pretty_print(jobs))
+
+    assert "Aggregate jobs:" in out
+    # Shared BUILD is hoisted, not repeated per-row.
+    assert "  build: 20240101-1\n" in out
```
mjdonis: tests are fine but assertions in testcases for hoisted builds would be better if they actually verify that this is printed _only once_, something like
`assert out.count("  build: 20240101-1\n") == 1`

same for line 357

tests/test_qem_dashboard_connector.py
```diff
@@ -157,3 +221,235 @@ def test_dashboard_auto_openqa_loads_incident_and_aggregate_jobs(mock_config):
...
+    mixed_build = [
+        _aggregate_job(6100, "mau-x", "passed", product="A", build="20260101-1"),
+        _aggregate_job(6101, "mau-y", "passed", product="B", build="20260102-2"),
+    ]
+    out_mixed = "".join(dashboard._pretty_print(mixed_build))
+    # Hoist line absent.
+    assert "  build: 20260101-1\n" not in out_mixed
+    assert "  build: 20260102-2\n" not in out_mixed
+    # Per-row build present on at least one summary row (folded/full).
+    assert "build: 20260101-1" in out_mixed
```
mjdonis: similar to the other comment, since per-line format is "- build: <value>", it would be better if thse assertions use
`assert "- build: 20260101-1" in out_mixed`
`assert "- build: 20260102-2" in out_mixed`
instead of just `"build: 20260102-2" in out_mixed"` to avoid matching the hoisted format or any other unexpected output

mtui/export/auto.py
```diff
@@ -20,6 +20,70 @@
...
+        job_url = result.url.rsplit("/file/", 1)[0]
+        return (
+            f"{result.distri}_{result.version}_{result.arch} => {status}: {job_url}\n"
+        )
+
+    def _install_status(self) -> str:
+        auto = self.openqa.auto
+        results = auto.results if auto is not None else None
+        if not results:
+            return "FAILED"
```
mjdonis: why return FAILED when there are no results? I'm guessing that results being None means that the tests haven't run, are still running, or failed to fetch. Wouldn't it be better to set them to UNKNOWN or something like that?

mjdonis (approved):

PR #114, opened 2026-05-11
perf(qem-dashboard): fan out per-setting jobs fetches concurrently

mimi1vx:
## Summary

- Loading one incident from the QEM Dashboard issues `N+2` sequential GETs (one per incident setting + one per update setting). The API has no batch endpoint, so on large incidents this serial chain dominates wall time of `mtui` startup and any command that pulls fresh dashboard data (`reloadoqa`, `simpleset`, `export`).
- `DashboardAutoOpenQA._load_jobs` now uses `concurrent.futures.ThreadPoolExecutor` to run `incident_settings` and `update_settings` in parallel, then fans out every per-setting `jobs/incident/{id}` / `jobs/update/{id}` request concurrently in a second executor.
- Public surface of `QEMDashboardClient` and `DashboardAutoOpenQA` is unchanged — same requests issued, same payloads, same `dashboard.jobs` / `dashboard.pp` / `dashboard.results`.

## Implementation notes

- Futures are read back in **submission order**, so the resulting `jobs` list preserves the insertion order the sequential code produced. `_pretty_print` summaries depend on this ordering and stay byte-identical for unchanged inputs.
- `_get`'s existing swallow-and-return-`None` error handling is preserved; a single failed setting fetch no longer blocks or fails the rest of the batch.
- Worker count is the stdlib default (`min(32, cpu_count + 4)`), matching the existing concurrency pattern in `mtui/export/downloader.py` and `mtui/commands/{addhost,removehost,quit}.py` — no new config knob.
- No new dependencies; `requests` calls remain thread-safe (no shared `Session`).

## Tests

- Adjusted `test_qem_incident_uses_review_id_for_slfo_1_2` so the two now-concurrent settings URLs are asserted as an unordered set; the synchronous `incidents/…` call from `QEMIncident.__init__` is still asserted first.
- Added `test_dashboard_auto_openqa_fans_out_per_setting_fetches`: mocks three incident + three update settings, verifies every per-setting URL is hit exactly once, and pins the deterministic insertion order of `dashboard.jobs`.

## Verification

All local gates green:

```
uv run ruff format --check .   # clean
uv run ruff check .            # clean
uv run ty check                # clean (strict mode applies to mtui/connector/**)
uv run pytest                  # 548 passed
```

## Risks

- Increased burst load against the dashboard host. Stdlib's default worker pool can be aggressive on large machines; if the dashboard turns out to have rate limits we'd cap `max_workers` in a follow-up.
- No behavior change for callers; `_pretty_print` output stays stable because insertion order is preserved.

### DIFF
--- a/mtui/connector/qem_dashboard.py
+++ b/mtui/connector/qem_dashboard.py
@@ -16,6 +16,14 @@ logger = getLogger("mtui.connector.qem_dashboard")
 # per-group summary count to keep the report short and reviewable.
 FAILED_RESULTS: frozenset[str] = frozenset({"failed", "incomplete", "timeout_exceeded"})
 
+# (connect, read) timeout for every dashboard HTTP call. Bounds a stuck
+# socket so a broken network can't hang mtui startup indefinitely.
+_HTTP_TIMEOUT: tuple[float, float] = (5.0, 30.0)
+# Wall-clock cap per future in the parallel fan-out. Defense-in-depth on
+# top of the per-request timeout above; a stuck worker won't block the
+# whole batch.
+_FUTURE_TIMEOUT: float = 60.0
+
 
 class QEMDashboardClient:
     """Small read-only client for the QEM Dashboard API."""
@@ -29,6 +37,7 @@ class QEMDashboardClient:
                 f"{self.apiurl}/{path.lstrip('/')}",
                 params=params or None,
                 headers={"Accept": "application/json"},
+                timeout=_HTTP_TIMEOUT,
             )
             response.raise_for_status()
             return response.json()
@@ -111,8 +120,12 @@ class DashboardAutoOpenQA:
             update_settings_future = executor.submit(
                 self.client.update_settings, incident_number
             )
-            incident_settings = incident_settings_future.result()
-            update_settings = update_settings_future.result()
+            incident_settings = self._await_settings(
+                incident_settings_future, "incident_settings"
+            )
+            update_settings = self._await_settings(
+                update_settings_future, "update_settings"
+            )
 
         # Build a flat task list of (source, setting, fetcher) preserving
         # the original insertion order: all incident settings first, then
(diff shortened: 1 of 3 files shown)


mtui/connector/qem_dashboard.py
```diff
@@ -100,26 +101,57 @@ def __init__(
...
+        with concurrent.futures.ThreadPoolExecutor() as executor:
+            futures = [
+                executor.submit(_fetch, source, setting_id)
+                for source, _, setting_id in tasks
+            ]
+
+            jobs: list[dict[str, Any]] = []
+            for (source, setting, _), future in zip(tasks, futures, strict=True):
+                jobs.extend(
+                    self._normalize_job(job, source, setting) for job in future.result()
```
mjdonis: `future.result()` should probably be in a try/except block here, if fetching any of the futures fails (e.g. timeout, API error, or something of the sort) it would raise an exception and explode, and discard all successfully fetched results up to that point (and would not process the next futures either)

it would be a good idea to add a test for this case too

mimi1vx: I don't think so, its covered by https://github.com/openSUSE/mtui/blob/296d52a54b35f110e522d48ae772af5e82d30530/mtui/connector/qem_dashboard.py#L25 

which handles all error states, so when problems happens it simply returns None for job

mimi1vx: good catch, we should add here timeout :D on broken network it cause real problem

mjdonis: since this is wrapped in `_get`, the timeout should probably be resolved there in `requests.get()` (which afaik also has no default timeout), so I guess this is technically a pre-existent issue, but it would be a good idea to fix it

mimi1vx: should be done , added timeouts to both

mjdonis (approved):
lgtm, thank you for the changes

PR #119, opened 2026-05-14
refactor(commands): replace glob+globals plugin loader with __init_subclass__ registry

mimi1vx:
## Summary

Phase 5b / C5 of the [MTUI improvement plan](../PLAN.md): replace the
filesystem-glob + `dir()` reflection + `globals()`-injection plugin
loader in `mtui/commands/__init__.py` with a `Command.__init_subclass__`
registry. The loader still has to import every submodule (so subclass
hooks fire), but discovery is now driven by `pkgutil.iter_modules` and
the public surface collapses to a single `mtui.commands.registry:
dict[str, type[Command]]`.

- 55 commands registered, byte-identical surface to before (pinned by a
  characterization test introduced in the first commit of this PR).
- `BaseApiCall` (the abstract intermediate) is correctly skipped via
  `"command" not in cls.__dict__`.
- `quit` / `exit` / `EOF` (multi-class-per-module via subclassing) all
  register as before.
- Duplicate `command` strings now raise `CommandAlreadyBoundError` at
  class-creation time, not just at the first prompt instantiation.

## Changes

- `mtui/commands/_command.py` — add `Command.registry`,
  `__init_subclass__` hook, and relocate `CommandAlreadyBoundError` here
  so the hook can raise it.
- `mtui/commands/__init__.py` — rewrite using `pkgutil.iter_modules`;
  drop `cmd_list` and the `globals()` injection.
- `mtui/prompt.py` — iterate `commands.registry.values()` instead of the
  legacy `commands.cmd_list` + `getattr(commands, x)`; re-export
  `CommandAlreadyBoundError` for backwards compatibility.
- `tests/test_commands_registry.py` — new characterization test pinning
  the 55-entry surface and asserting the legacy attributes are gone.
- `tests/test_command_init_subclass.py` — new focused unit tests for the
  hook (skip-without-`command`, skip-when-only-inherited, duplicate-
  raises, alias-via-subclassing, registry identity).
- `CHANGELOG.md` — `[Unreleased] / Changed` entry documenting the
  module-attribute change.

## Commits

1. `test(commands): pin the loader's command surface (55 entries)` —
   characterization test, green against the pre-refactor `cmd_list`
   surface.
2. `refactor(commands): auto-register subclasses via
   Command.__init_subclass__` — adds the hook + relocates
   `CommandAlreadyBoundError`.
3. `refactor(commands): drop globals() loader for the Command.registry`
   — atomic swap of loader + consumer + characterization test (kept
   together so bisect blames a single commit).
4. `test(commands): cover Command.__init_subclass__ edge cases` — six
   focused unit tests with a `clean_registry` fixture.
5. `docs(changelog): note the Command.registry replacement of cmd_list`.

## Verification

All gates green:

- `uv run ruff format --check .` — 166 files clean
- `uv run ty check` — clean (no warnings; `error-on-warning` on)
- `uv run pytest` — **585 passed** (Phase 5b/C9 baseline 549 + 4 new
  characterization + 6 new init_subclass tests + 26 already on branch)
- `uv run python -m mtui --help` and `-V` — both work
- `rg cmd_list mtui/` and `rg "globals\(\)\[" mtui/commands/` — zero
  hits

## Notes for reviewers

- The duplicate-guard semantics shift slightly: today, a duplicate
  `command` string would only blow up the first time
  `CommandPrompt._add_subcommand` saw it; from this PR on, a duplicate
  declaration anywhere in the codebase raises at import time. This is
  strictly more correct (catches packaging-time mistakes earlier). The
  per-prompt local guard in `_add_subcommand` is left in place as
  belt-and-braces.
- A grep of the repo and tests confirms the only consumers of the
  legacy `cmd_list` / `globals()` surface were the two lines in
  `mtui/prompt.py`. The `Command` re-export from `mtui.commands` is
  preserved.
- The new characterization test scopes its snapshot to classes whose
  `__module__` starts with `mtui.commands.`, so test-only `Command`
  subclasses (e.g. `ConcreteCommand` in `tests/test_command_base.py`)
  do not pollute the snapshot when tests run in arbitrary order.

### DIFF
--- a/tests/test_commands_registry.py
+++ b/tests/test_commands_registry.py
@@ -1,0 +1,125 @@
+"""Characterization tests for the ``mtui.commands`` plugin registry.
+
+These tests pin the exact set of commands the package exposes so that the
+Phase 5b / C5 refactor (filesystem-glob + ``globals()`` -> ``__init_subclass__``
+registry) cannot silently change the public surface.
+
+The ``EXPECTED`` constant lists every ``(command_string, module, class_name)``
+triple the loader produces. The set must stay byte-identical across the
+refactor; if a real command is added or renamed in a separate change, update
+``EXPECTED`` in the same commit that introduces the change.
+"""
+
+from __future__ import annotations
+
+from mtui import commands
+
+# Snapshot taken on phase_five_b2 before the C5 refactor.
+EXPECTED: frozenset[tuple[str, str, str]] = frozenset(
+    {
+        ("EOF", "mtui.commands.quit", "DEOF"),
+        ("add_host", "mtui.commands.addhost", "AddHost"),
+        ("analyze_diff", "mtui.commands.showdiff", "AnalyzeDiff"),
+        ("approve", "mtui.commands.apicall", "Approve"),
+        ("assign", "mtui.commands.apicall", "Assign"),
+        ("checkout", "mtui.commands.checkout", "Checkout"),
+        ("comment", "mtui.commands.apicall", "Comment"),
+        ("commit", "mtui.commands.commit", "Commit"),
+        ("config", "mtui.commands.config", "Config"),
+        ("downgrade", "mtui.commands.downgrade", "Downgrade"),
+        ("edit", "mtui.commands.edit", "Edit"),
+        ("exit", "mtui.commands.quit", "QExit"),
+        ("export", "mtui.commands.export", "Export"),
+        ("get", "mtui.commands.sftpcmd", "SFTPGet"),
+        ("install", "mtui.commands.zypper", "Install"),
+        ("list_bugs", "mtui.commands.simplelists", "ListBugs"),
+        ("list_history", "mtui.commands.simplelists", "ListHistory"),
+        ("list_hosts", "mtui.commands.simplelists", "ListHosts"),
+        ("list_locks", "mtui.commands.simplelists", "ListLocks"),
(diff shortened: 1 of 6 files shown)


tests/test_commands_registry.py
```diff
@@ -0,0 +1,125 @@
...
+    actual = _current_surface()
+    assert actual == EXPECTED, {
+        "missing": sorted(EXPECTED - actual),
+        "unexpected": sorted(actual - EXPECTED),
+    }
+
+
+def test_registry_count_matches_snapshot() -> None:
+    """Guard against silent additions: command count is part of the contract."""
+    assert len(_current_surface()) == len(EXPECTED) == 55
```
mjdonis: The hardcoded 55 here should be dropped, comparing against `len(EXPECTED)` is enough and having the number there means we'd have to manually change it each time a command is added/removed

mimi1vx: Done in 55b642a — dropped both the trailing `== 55` from the count assertion and the matching "; 55 entries" suffix from the snapshot comment. The test now reads `len(_current_surface()) == len(EXPECTED)`, so adding/removing a command only requires editing `EXPECTED`.

mimi1vx: rebased and fixup into original test suite commit

mjdonis (approved):

PR #120, opened 2026-05-14
refactor(target): replace queue+thread with ThreadPoolExecutor

mimi1vx:
## Why

Three problems with the existing parallel-target machinery in `mtui/target/actions.py`:

1. **Module-level mutable state.** A single `queue: Queue[(callable, args)]` was shared across every concurrent `HostsGroup` operation in the process. It only happened to work because `mtui` never ran two operations concurrently — a global waiting for someone to step on it.

2. **Silent exception swallow.** `ThreadedMethod` ran the worker callable inside a bare `try/finally` that called `task_done()` but never `.result()`. A failing `Target.set_repo` / `sftp_*` / `run` took its worker thread down without ever surfacing to the caller, and the enclosing `update` / `prepare` / `downgrade` flow happily marched on as if it had succeeded. `pytest` had been quietly emitting `PytestUnhandledThreadExceptionWarning` for years.

3. **Reinvented the standard library.** `concurrent.futures.ThreadPoolExecutor` has solved "fan a callable out across N threads, surface exceptions, join on scope exit" since Python 3.2. The bespoke implementation was twice the code and missing the exception-propagation guarantee. (`mtui` started as a python-2.2-era script; this code dates back to that era.)

Replace the queue / `ThreadedMethod` / spinner machinery with a small `run_parallel()` helper on top of `ThreadPoolExecutor`. It builds one pool sized to the work, submits each `(callable, args)` pair, and iterates `as_completed` so the first worker exception re-raises into the caller. `ThreadedTargetGroup` collapses to a 5-line abstract base; `RunCommand`'s serial path becomes a plain for-loop with `prompt_user` between iterations; `HostsGroup._fanout_set_repo` is now a one-liner against the same helper.

Net effect: large LOC reduction in production modules, exceptions now surface where the user can see them, and the queue / `ThreadedMethod` / spinner seam disappears from `hostgroup`'s public surface.

## What changes for users

Two user-visible changes, both documented in `CHANGELOG.md` under `[Unreleased] / Fixed`:

> Exceptions raised inside parallel target operations (`Target.set_repo`, `Target.run`, the SFTP helpers used by `put` / `get` / `remove`) now surface to the caller and abort the enclosing `update` / `prepare` / `downgrade` / `install` / `uninstall` flow, instead of being silently lost in a worker thread while the flow pretended to succeed.

> Pressing `Ctrl+C` during a parallel `run` now drops queued hosts that have not started yet and unblocks in-flight workers by closing their SSH sessions, instead of waiting for the whole parallel batch to drain before the interrupt was honoured. Workers already executing a remote command will still finish their current command (Python cannot interrupt a foreign blocking call), but session shutdown unblocks them promptly.

> Long-running parallel operations (`run`, file uploads / downloads / deletes, `set_repo` fan-out used by `update`/`prepare`/`downgrade`) now show user-visible progress again via a `|/-\` spinner on stderr while work is in flight. The spinner is suppressed when stderr is not a TTY (log files and redirected output stay clean).

The exception-propagation change is the bug-fix half. The locks acquired by `update_lock` are still released because the surrounding `try/finally: self.unlock()` blocks remain in place; users now see the failure instead of a silent half-update.

## Verification

- `uv run ruff format --check .` ✓
- `uv run ruff check .` ✓
- `uv run ty check` ✓
- `uv run pytest` — **593 passed**
- `uv run python -m mtui --help` and `-V` ✓
- `rg "queue\.put|queue\.unfinished_tasks|queue\.join\(\)|ThreadedMethod\(" mtui/` returns zero hits — the hand-rolled global is gone

## Out of scope

- Async migration to `asyncssh`. Substantial rewrite; defer until other refactors stabilise.

### DIFF
--- a/mtui/target/actions.py
+++ b/mtui/target/actions.py
@@ -4,12 +4,11 @@ from __future__ import annotations
 
 import sys
 import threading
-import time
 from abc import ABC, abstractmethod
 from collections.abc import Callable, ValuesView
+from concurrent.futures import ThreadPoolExecutor, as_completed
 from contextlib import suppress
 from pathlib import Path
-from queue import Empty, Queue
 from threading import Lock
 from typing import Any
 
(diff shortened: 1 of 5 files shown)


mtui/target/actions.py
```diff
@@ -2,50 +2,37 @@
...
-from queue import Empty, Queue
 from threading import Lock
 from typing import Any
 
 from ..types import ExecutionMode
 from ..utils import prompt_user
 from . import Target
 
-queue: Queue[tuple[Callable[..., None], list[Any]]] = Queue()
+logger = getLogger("mtui.target.actions")
```
mjdonis: why the logger? I don't see it being used in this file

mtui/target/actions.py
```diff
@@ -197,81 +138,38 @@ def __init__(
...
-                thread.start()
-                queue.put((serial[target].run, [self.command, lock]))
-
-                while queue.unfinished_tasks:
-                    spinner(lock)
-
-                queue.join()
+                t.run(self._cmd_for(h), lock)
 
         except KeyboardInterrupt:
```
mjdonis, line 247: I think there's a problem here:

In the docstring for `run_parallel`, it says that "the first exception surfaces to the caller, the rest of the in-flight work is allowed to finish on the way out of the executor's ``with`` block."

So basically if I do run Ctrl+C and cause a `KeyboardInterrupt`, at the point where this block starts, the context manager aready exited and allowed any in-flight work to finish. 

That means that there's really no way to stop all refhosts from running a command using Ctrl+C, since we can't stop the workers that are mid-execution

mjdonis (commented):
I'm also concerned about removing the spinner since I don't see a replacement here.

While it didn't really indicate actual progress, it at least shows to the user "mtui is alive and waiting for operations to complete" vs complete silence that looks like a hang

While commands are running (especially with long operations, e.g. running update on a Product Increment or an MU with lots of packages), the user would now see no output at all, there would be no visual feedback to distinguish if mtui is working on something or not

mtui/target/actions.py
mimi1vx: Good catch — dropped the unused logger in 3264aff.

mtui/target/actions.py
mimi1vx: You're right — the `with ThreadPoolExecutor(...)` block was joining all in-flight workers before letting `KeyboardInterrupt` escape, so the `except` here was effectively dead code for long-running commands.

Fixed in 3264aff: `run_parallel` now manages the executor lifecycle explicitly and calls `ex.shutdown(wait=False, cancel_futures=True)` on any propagating exception, so:

- queued-but-not-started callables are cancelled outright;
- in-flight workers are *not* awaited (the caller returns immediately);
- `RunCommand`'s `except KeyboardInterrupt` now actually runs while workers may still be alive — its `t.connection.close_session()` loop unblocks any worker still parked inside `connection.run`, so they exit cleanly within one round-trip.

The one honest limitation: Python can't interrupt a foreign C-level blocking call, so a worker mid-`exec_command` will run to completion of *that single command*. The session-close just makes sure it returns promptly instead of hanging forever. The new docstring on `run_parallel` and the changelog entry call this out.

Added `test_run_parallel_keyboard_interrupt_cancels_queued_work` in tests/test_actions.py to pin the contract.

mimi1vx:
@mjdonis re. spinner removal — fair point, that was a real UX regression. Restored in deafd08:

- A `|/-\\` spinner thread is started by `run_parallel` whenever a `desc` is passed by the caller. It writes to stderr only when `sys.stderr.isatty()` is true, so log files, redirected output, and pytest stay clean. It erases its line on stop.
- In addition (because the spinner is invisible in non-interactive runs), each completed worker logs an INFO line of the form `<op>: <i>/<total> done`. So a long `update` against 20 hosts now leaves a 20-line audit trail in the mtui log even when stderr is not a TTY.
- All four parallel call sites pass a `desc`: `RunCommand` → `"run"`; `FileDelete` / `FileUpload` / `FileDownload` → their class name; `HostsGroup._fanout_set_repo` → `f"set_repo {operation}"`. So `update` / `prepare` / `downgrade` / `install` / `uninstall` / `put` / `get` / `remove` / `run` all show progress again.
- 3 new tests in `tests/test_actions.py` pin the contract: spinner is silent without a TTY, INFO lines fire when `desc` is given, INFO lines stay quiet when `desc` is omitted (so the test-only `run_parallel(work)` path stays log-clean).

mimi1vx:
Follow-up on the spinner restore: dropped the per-completion INFO log lines in b926374 after manual testing — they were too noisy on multi-host runs (e.g. an `update` against 20 hosts buried real diagnostics behind 20+ progress lines per phase). The TTY spinner is now the only progress channel; `run_parallel` itself emits no log records.

Net shape:
- TTY: spinner labelled with the operation (`run`, `set_repo add`, `FileUpload`, ...).
- Non-TTY (log files, redirected output, CI): silent — same as before this PR.

Test pinned at `tests/test_actions.py::test_run_parallel_emits_no_log_records` so any future "let's log progress too" attempt has to actively delete it.

baierjan (approved):

PR #121, opened 2026-05-18
refactor(connection): batch multi-step SFTP ops over a single session

mimi1vx:
## Summary

Today every `Connection.sftp_*` helper pays a full SFTP-channel
handshake per call -- one `client.open_sftp()`, one op, one `close()`.
Two of those helpers are themselves multi-step (`sftp_get_folder` does
listdir + N gets; `sftp_rmdir` does listdir + N removes + rmdir), and
`parsers/system.parse_system` calls 6+ helpers back-to-back against the
same connection. Every step pays the handshake.

This PR introduces a private `@contextmanager _sftp(self)` plus a public
`Connection.sftp_session()` and routes every public `sftp_*` method
through them, so a session opened once at CM entry is reused for the
whole `with` block. `sftp_get_folder` and `sftp_rmdir` are rewritten to
inline their child ops against the live client (`N+2` handshakes per
call drop to `1`), and `parse_system` is wrapped in one
`sftp_session()` (6+ handshakes drop to `1`).

## Why

- **Latency on slow target links.** Each SFTP open is a full SSH channel
  open + SFTP version negotiation roundtrip. On the typical maintenance
  SUT this is the dominant cost of `add_host` (which calls `parse_system`)
  and of `update`'s post-flight cleanup (`sftp_rmdir` on the staging
  repos). The fan-out is `O(targets)` so the savings compound.
- **Readability.** Eight near-identical `sftp = self.__sftp_reconnect();
  try: ...; finally: sftp.close()` blocks collapse into `with self._sftp()
  as sftp:`. The open/close lifecycle now lives in one place.
- **Public API for future callers.** Any caller that wants to perform
  short batches of SFTP ops against one host can now do so without
  reaching into private methods.

## Key design decisions

1. **Public `sftp_session()` in addition to the private CM.** Lets
   external callers (starting with `parse_system`) batch their own
   multi-step flows; the alternative of keeping the CM private would
   force each future caller to grow its own per-call helpers.
2. **No reconnect inside a session.** Today each `sftp_*` reconnects up
   to `RETRIES` times if the SSH session is dead. Under the long-lived
   session CM, reconnect happens **once** at CM entry; mid-block paramiko
   errors propagate to the caller. Auto-retrying the whole block would
   re-apply already-applied side effects, and matches paramiko's own
   model.
3. **`sftp_open` keeps its manual lifetime.** The returned `SFTPFile`
   holds a strong reference to the SFTP channel, so the file object
   stays usable after `client.close()`. Routing `sftp_open` through the
   CM would close the client on exit and break the handle. Documented
   inline.
4. **Surface, don't also fix.** While reading I noticed
   `Connection.__sftp_open`'s except-clause has a dead defensive branch
   (`if "sftp" in locals()` is never true after a failed assignment).
   Documented under `[Unreleased] / Known issues` and deferred to a
   follow-up fix PR rather than mixed into a refactor.

## Commits

```
f6aa040 refactor(connection): route public sftp_* methods through _sftp() context manager
b5d2a98 refactor(connection): reuse one SFTP session in sftp_get_folder and sftp_rmdir
17aa42b refactor(parsers/system): batch parse_system over one SFTP session
5755999 test(connection): pin sftp_session reuse and multi-step batch contract
8a4ce05 docs(changelog): note SFTP session batching and the surfaced __sftp_open dead branch
```

Each refactor commit is independently reviewable; tests are pinned in a
dedicated `test(...)` commit so bisect can attribute behaviour changes
cleanly.

## Compatibility

- **Public API.** Every existing `Connection.sftp_*` method keeps its
  signature, return type, and exception contract. Per-call handshake
  count is unchanged for the seven simple helpers (still 1 open + 1
  close per call, just through the CM).
- **`sftp_get_folder` / `sftp_rmdir`.** Same ops, same order, same
  exceptions. Only the SSH/SFTP handshake count drops from `N+2` to `1`.
  The `OSError` handling that `Connection.sftp_remove` previously did
  per file is preserved by wrapping the inlined `sftp.remove` in a
  `try/except OSError: logger.exception(...)`.
- **`parse_system`.** Same ops, same order. Paramiko's `sftp.listdir`
  raises `OSError` (caught as before) and `sftp.open` raises
  `FileNotFoundError` for missing files (because `OSError(ENOENT)`
  auto-promotes in Python 3). The `Path()` wrapping is dropped because
  paramiko's `sftp.*` methods take `str` directly; this also drops the
  now-unused `from pathlib import Path`.

## Verification

All four CI gates pass locally:

- `uv run ruff format --check .` -- 167 files clean
- `uv run ruff check .` -- clean
- `uv run ty check` -- clean (no warnings; `error-on-warning` is on)
- `uv run pytest` -- **597 passed** (593 baseline + 4 new in
  `tests/test_connection.py`)
- `uv run --group doc sphinx-build -W -b html Documentation /tmp/x` -- clean
- `uv run python -m mtui --help` and `-V` -- both succeed

Also confirmed:

- `rg "self\.__sftp_reconnect\(\)" mtui/connection.py` -- 3 hits (was
  8): inside `_sftp`, inside `sftp_open`'s preserved manual path, and
  inside `sftp_put`'s pre-existing reconnect-on-mkdir-fail loop.
- `rg "sftp\.close\(\)" mtui/connection.py` -- 3 hits (was 9): inside
  `_sftp`'s `finally`, inside `__sftp_open`'s cleanup, and inside
  `sftp_open`'s preserved manual cleanup.

### DIFF
--- a/mtui/connection.py
+++ b/mtui/connection.py
@@ -562,31 +562,25 @@ class Connection:
         """
         with self._sftp() as sftp:
             path = ""
-            # create remote base directory and copy the file to that directory
+            # create remote base directory and copy the file to that directory.
+            # Paramiko transport errors propagate per the ``_sftp`` contract;
+            # don't rebind ``sftp`` mid-loop or the CM cannot close the
+            # replacement client. ``OSError`` covers the common "directory
+            # already exists" case (EEXIST) and is the only exception we
+            # intentionally swallow.
             for subdir in str(remote).split("/")[:-1]:
                 path += subdir + "/"
-                created = False
-                while not created:
-                    try:
-                        sftp.mkdir(path)
-                        created = True
-                    except (
-                        AttributeError,
-                        paramiko.ChannelException,
-                        paramiko.SSHException,
-                    ):
-                        created = False
-                        sftp = self.__sftp_reconnect()
-                    except Exception:
-                        # Most commonly: directory already exists (sftp.mkdir
-                        # raises OSError on EEXIST). Treat as success but leave
-                        # a breadcrumb so unexpected failures are diagnosable.
-                        logger.debug(
-                            "sftp mkdir %s treated as success",
-                            path,
-                            exc_info=True,
-                        )
-                        created = True
+                try:
+                    sftp.mkdir(path)
+                except OSError:
+                    # Most commonly: directory already exists. Treat as
+                    # success but leave a breadcrumb so unexpected failures
+                    # are diagnosable.
+                    logger.debug(
+                        "sftp mkdir %s treated as success",
+                        path,
+                        exc_info=True,
+                    )
 
             logger.debug(
                 "transmitting %s to %s:%s:%s",
--- a/mtui/connection.py
+++ b/mtui/connection.py
@@ -637,7 +637,11 @@ class Connection:
                 remote,
                 local,
             )
-            files = self.sftp_listdir(remote)
+            # listdir directly on the live client so the whole batch
+            # (1 listdir + N gets) shares one SFTP session; previously
+            # listdir went through self.sftp_listdir, opening a second
+            # client.
+            files = sftp.listdir(str(remote))
             for file in files:
                 sftp.get(
                     f"{remote}/{file}",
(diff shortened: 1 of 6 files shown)


mtui/connection.py
```diff
@@ -530,48 +560,47 @@ def sftp_put(self, local: Path, remote: Path) -> None:
...
+                    try:
+                        sftp.mkdir(path)
+                        created = True
+                    except (
+                        AttributeError,
+                        paramiko.ChannelException,
+                        paramiko.SSHException,
+                    ):
+                        created = False
+                        sftp = self.__sftp_reconnect()
```
mjdonis: this here is opening a new connection that is not managed by the context manager, so `_sftp` will close the original connection when it exits, but the new one created here will never be closed

mimi1vx: Fixed :D thx for finding

mjdonis (approved):

PR #125, opened 2026-05-22
feat: add openqa_overview interactive command (port of oqa-search)

mimi1vx:
## Why this change was needed

QA engineers working through a maintenance update review routinely switch
between the mtui REPL and the standalone
[oqa-search](https://github.com/mjdonis/oqa-search) tool to look up
openQA job results and qam.suse.de build-check summaries for the current
MU. Having to switch tools breaks the review flow, requires oqa-search to
be installed separately, and means the looked-up data cannot feed back
into the testreport export automatically.

## What was done

Ported oqa-search's three core features into mtui as a new interactive
REPL command `openqa_overview`. After `load_template <RRID>`, the command:

1. **Single Incidents - Core** — queries openQA for PASSED/FAILED/RUNNING
   per SLE version and prints a direct overview URL for each.
2. **Aggregated Updates** — walks back `--days` days (default 5) in the
   aggregated job groups to find a build covering the incident.
3. **Build checks** — scrapes the qam.suse.de `build_checks/` HTML index
   and extracts test-result summary lines from each log, stripping OBS
   timestamp prefixes (`[ Ns]`) from the displayed output.

The structured payload is stored on `metadata.openqa.overview` so it
survives to the export pipeline without a second network round trip.

**`openqa_overview --export`** injects the rendered overview into the
testreport's `log` file under `regression tests:`. The block is
bracketed by `<!-- mtui openqa_overview begin -->` /
`<!-- mtui openqa_overview end -->` markers so re-exports replace the
prior block in place rather than accumulating duplicates. The regular
`export` command (all three workflows: auto/manual/kernel) also picks
up `metadata.openqa.overview` automatically when it is populated.

**`--no-fetch`** reuses the cached payload without hitting the network,
useful when you want to re-export after changing flags without waiting
for another full fetch.

TLS verification is disabled on the shared `requests.Session` because
openqa.suse.de, dashboard.qam.suse.de and qam.suse.de present
self-signed or internal-CA certificates that the standard system trust
store does not validate — this mirrors the effective behaviour of the
upstream oqa-search tool on SUSE developer machines.

## Commits

```
33e864a feat(types): add OpenQAOverviewResult dataclass and wire into OpenQAResults
48264fa feat(connector): add oqa_search connector porting upstream oqa-search
43abc0b feat(commands): add openqa_overview interactive REPL command
a1d8f84 feat(export): inject openqa_overview block into regression tests section
5fd75f3 test: add tests for oqa_search connector, command, injector, and exporter
d2d2d1c docs: document openqa_overview command and --export flag
```

Each commit is independently reviewable and bisectable.

## Key design decisions

- **No positional `update_id` argument**: the command uses the already-loaded
  testreport's `rrid`, consistent with `reload_openqa` and other commands.
- **Connector module, not extended existing connectors**: oqa-search's
  job-group enumeration and aggregated-updates date-walking do not map onto
  the incident-centric `AutoOpenQA`/`KernelOpenQA` design; a dedicated
  `mtui/connector/oqa_search.py` keeps them isolated.
- **Return data, never print**: the connector functions return typed
  dataclasses; the command layer applies colour and OBS-timestamp stripping
  for the REPL, and the injector uses the plain-text renderer for the log
  file. The stored `metadata.openqa.overview` payload retains raw lines.
- **Marker-bounded idempotent injection**: HTML-comment markers allow
  `overview_inject.py` to find and replace a prior block in place.
  Exactly one blank line is guaranteed above and below the block regardless
  of how many times `--export` is run.

## Verification

All local CI gates pass:

- `uv run ruff format --check .` — 173 files clean
- `uv run ty check` — clean (`error-on-warning` is active; `mtui/connector/**` runs under `all = "error"`)
- `uv run pytest` — **638 passed** (622 baseline + 17 connector + 10 command + 11 injector + 2 export × 2 new)
- `uv run --group doc sphinx-build -W -b html Documentation Documentation/.build/html` — build succeeded

Manually verified end-to-end:

```
mtui> load_template SUSE:Maintenance:44425:410560
mtui> openqa_overview
OpenQA:
#######
Single incidents - Core
12-SP3-TERADATA -> https://openqa.suse.de/tests/overview?...
PASSED
12-SP5 -> https://openqa.suse.de/tests/overview?...
PASSED
-------
Aggregated updates - Core
12-SP5 -> https://openqa.suse.de/tests/overview?...
PASSED
-------
Build checks:
#############
https://qam.suse.de/.../xz.SUSE_SLE-12_Update.x86_64.log
All 9 tests passed
```

## Checklist

- [x] `uv run ruff format --check .` is clean.
- [x] `uv run ruff check .` is clean.
- [x] `uv run ty check` is clean.
- [x] `uv run pytest` passes.
- [x] User-visible changes are recorded under `[Unreleased]` in `CHANGELOG.md`.
- [x] Documentation under `Documentation/iui.rst` updated (`openqa_overview` section added after `reload_openqa`).


mtui/connector/oqa_search.py
```diff
@@ -0,0 +1,823 @@
...
+    lines.append("")
+
+    # --- Single Incidents - Core ---
+    if single_incidents_rows:
+        lines.append("### Single Incidents - Core")
+        lines.append("")
+        for row in single_incidents_rows:
+            lines.extend(_render_version_row(row))
+        lines.append("")
+    elif not skip_aggregated and not aggregated_updates_rows:
```
mjdonis: I think there's a problem here:

if there's no single incidents results and the user passed `--no-aggregated` then it would not get into this elif as `not skip_aggregated` would resolve to `False`

in that case, since you are skipping the aggregated section due to the flag, then it wouldn't output anything, even though the user is expecting results from single incidents or the "no openQA builds" hint

mtui/connector/oqa_search.py
```diff
@@ -0,0 +1,823 @@
...
+
+    # --- Build checks ---
+    if build_checks_rows:
+        lines.append("### Build Checks")
+        lines.append("")
+        for entry in build_checks_rows:
+            lines.extend(_render_build_check(entry))
+    else:
+        lines.append("### Build Checks")
+        lines.append("")
```
mjdonis: nit: the build checks header is duplicated here, it could be moved outside of the if/else block since it's going to be added to the output anyways

mtui/commands/openqa_overview.py
```diff
@@ -0,0 +1,392 @@
...
+        ``build_checks`` re-appends ``/testreports/...``.
+        """
+        return reports_url.rstrip("/").removesuffix("/testreports")
+
+    @staticmethod
+    def _title(text: str) -> str:
+        """Cyan-ish title via mtui's color helpers (blue, since no cyan exists)."""
+        # mtui exposes green/red/yellow/blue; use blue for titles. The
+        # helpers honour colors_enabled() automatically.
+        return text  # titles stay plain so they survive piping into mtui logs
```
mjdonis: I'm a bit confused, what's the point of this? Is the intention to apply color here in some way in the future and this is a placeholder or is it just a no-op?

mjdonis:
new tests are nice, and some of them cover parts of the upstream code, but coverage is missing for lots of helper functions

e.g. `_is_valid_template`, `_is_name_matching`, `_filter_openqa_groups` and couple more don't have any test coverage, most of them are mocked in the new tests

we should port those tests from upstream

and also there's very few and basic tests for the build checks, there's a bunch of tests with real logs from upstream to test the heuristics and `summarize_test_results`

mimi1vx:
@mjdonis some tests added :D

mjdonis (approved):
lgtm, thanks!

PR #138, opened 2026-05-29
feat(locks): reap stale reference-host locks on connect

plusky:
## Summary

Reference hosts are shared, and their `/var/lock/mtui.lock` is cooperative/advisory. Today, when `add_host` (or any connect) hits a host that is already locked, mtui only logs a warning — even when the lock is days old and the owning session is long gone. Testers then have to manually `unlock -f` shared hosts before they can work.

This change makes mtui **force-remove a pre-existing lock that is older than a configurable threshold** on connect, regardless of which user or session created it (including exclusive/commented locks), since such a lock is almost always left over from a crashed or abandoned session. Fresh locks are still only reported, never removed.

The removal is logged at `WARNING` so it stays observable:

```
<host>: removing stale lock held by <user> (<N> h old)
```

## Configuration

New `[lock]` section, default-on with a one-day threshold:

| Option | Type | Default | Meaning |
| --- | --- | --- | --- |
| `reap_stale` | bool | `true` | Reap stale locks on connect; `false` restores warn-only behaviour |
| `stale_age` | int (seconds) | `86400` | Age beyond which a lock is stale; `0` also disables reaping |

A missing or malformed lock timestamp is treated as "leave alone".

## Implementation

- `TargetLock` gains `age_seconds()` (epoch math on the existing lockfile timestamp) and `reap_if_stale()` (honours the config, force-unlocks when stale). It now also stores the `config` reference.
- `Target.connect()` changes one line: `if self.is_locked() and not self._lock.reap_if_stale(): warn(...)`.
- Two new `ConfigOption`s in `mtui/config.py`.

## Tests / docs

- `tests/test_locks.py`: new `TestTargetLockReaping` (age computation, foreign + exclusive reap, fresh-lock kept, both disable switches, malformed-timestamp kept).
- `tests/test_target.py`: updated the existing locked-connect test for the new path + added a reaped-lock test.
- `CHANGELOG.md` (Added) and `Documentation/cfg.rst` (option docs + example).

All gates pass locally: `ruff format --check`, `ruff check`, `ty check`, `pytest` (707 passed), and `sphinx-build -W`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,20 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
 
 ## [Unreleased]
 
+### Added
+
+- Stale reference-host locks are now reaped automatically on connect.
+  When `add_host` (or any connect) finds a pre-existing `/var/lock/mtui.lock`
+  older than a configurable threshold, mtui force-removes it regardless of
+  owner — including exclusive (commented) locks — since such a lock is
+  almost always left over from a crashed or abandoned session. The removal
+  is logged at `WARNING` (`<host>: removing stale lock held by <user>
+  (<N> h old)`). Controlled by two new `[lock]` config options:
+  `reap_stale` (default `true`) and `stale_age` in seconds (default
+  `86400`, i.e. one day); set `reap_stale = false` or `stale_age = 0` to
+  restore the previous behaviour of only warning. Fresh locks are still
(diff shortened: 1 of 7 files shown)


mimi1vx (dismissed):

mimi1vx:
@copilot resolve the merge conflicts in this pull request

mimi1vx (approved):

PR #146, opened 2026-06-01
feat(commit): default message from export footer instead of editor

plusky:
## Summary

Running `commit` without `-m` previously left `svn ci` to open an editor prompting for a commit message. This commits the testreport non-interactively with a generated message that reuses the testreport **export footer**, so the SVN history mirrors what's recorded at the end of the report:

```
committed from MTUI:17.1.0, paramiko 3.5.1 on openSUSE Leap-16.0 (kernel: 6.12.0-160000.29-default) by mpluskal
```

Passing `-m` still uses the supplied message verbatim.

## Implementation

- `mtui/support/systemcheck.py` — `system_info()` gains a `prefix` parameter (default `"## export"`, so the export footer is unchanged). The `commit` command reuses it with `prefix="committed from"`.
- `mtui/commands/commit.py` — when no `-m` is given, build the message via `system_info(...)` (rstripped) and pass it as `svn ci -m <message>` instead of an empty message.

## Tests / docs

- `tests/test_systemcheck.py`: `system_info` default vs. custom prefix.
- `tests/test_cmd_commit.py`: default (no `-m`) reuses the export footer with the `committed from` prefix; explicit `-m` still passes through verbatim.
- `CHANGELOG.md` (Changed).

All gates pass locally: `ruff format --check`, `ruff check`, `ty check`, `pytest` (999 passed).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -172,3 +172,6 @@ tags
 
 # Coverage / profile artefacts
 *.profraw
+
+# Claude Code session artefacts
+.claude/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
 
 ### Changed
 
+- `commit` without `-m` no longer opens an editor to ask for a message.
+  It now commits non-interactively with a generated message that reuses
+  the testreport export footer, e.g. `committed from MTUI:<version>,
+  paramiko <version> on <distro>-<verid> (kernel: <kernel>) by <user>`.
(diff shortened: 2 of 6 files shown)


mimi1vx (changes requested):
remove Claude artefacts please , you can extend gitignore with that

plusky:
Thanks — removed the stray `.claude/scheduled_tasks.lock` that slipped in via `git add -A`, and extended `.gitignore` to ignore the `.claude/` directory. No Claude artefacts are tracked now. Force-pushed; please re-review when you have a moment.

mimi1vx (approved):

PR #183, opened 2026-06-18
feat(smelt): add --unassigned/--show-assignment to smelt_updates

plusky:
## What

Annotate the SLFO update queue (`smelt_updates`) with the current assignee:

- `--unassigned` — keep only updates nobody has picked up for `--group`.
- `--show-assignment` — add an assignee column.
- `--group` — review group for the lookup (default `qam-sle`).

Assignment is read from each PR's mtui assign/unassign comments via Gitea (`Gitea.assignee()`). The lookup is one Gitea call per row, computed **lazily** highest-priority-first and short-circuited by `--limit`, so the default listing (no flags) stays a single SMELT call, while:

```
smelt_updates --pending qam-sle-review --unassigned --limit 1
```

finds the top unassigned update in only a few calls. Without a Gitea token the flags are ignored with a hint instead of failing.

## Why

The SLFO feed has no per-tester assignment field — it lives in the PR's mtui comments. This makes "find the first unassigned update by priority" a single command instead of an out-of-tree script + custom Gitea comment walk.

## Depends on

Builds on #182 (`Gitea.assignee()` + `pr_api_url()`). Until #182 merges this PR's diff includes that commit; it will reduce to just the `smelt`/test changes once #182 lands.

## Testing

`tests/test_cmd_smelt.py` covers the plain listing (no assignment work), `--unassigned` filtering, lazy `--limit 1` (asserts probing stops at the first hit), `--show-assignment` column, the missing-token hint, and that `_assignee` builds the client from `external_url`. Full suite green; ruff clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,17 @@ All notable user-visible changes to MTUI are documented in this file.
 The format is based on [Keep a Changelog 1.1.0](https://keepachangelog.com/en/1.1.0/),
 and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- `smelt_updates` gains `--unassigned` and `--show-assignment` (with `--group`,
+  default `qam-sle`) to surface each SLFO update's current assignee, read from the
+  PR's mtui assign/unassign comments via Gitea. The lookup is lazy and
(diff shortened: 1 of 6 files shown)


mimi1vx:
missing CHANGELOG.md entry for user visible changes

mimi1vx (changes requested):
missing changelog entry for user visible changes

mimi1vx (approved):

PR #188, opened 2026-06-19
fix(connection): don't prompt for SSH password in non-interactive mode

mimi1vx:
## Problem

When SSH public-key authentication fails, `Connection.connect()` falls back to `getpass.getpass()` to ask for the root password. Under `mtui-mcp` there is no TTY and stdin is the JSON-RPC pipe, so the prompt is **invisible to the client and blocks the session indefinitely**.

Reported when a client (Claude) ran `add_host` against a refhost (`homer.qam.suse.cz`) that had no working key auth — mtui hung waiting for a password that could never be entered.

## Fix

- Add an `interactive: bool = True` parameter to `Connection.__init__`. When `False`, the key-auth-failure branch skips the password prompt and raises a new `NonInteractiveAuthRequired` instead of calling `getpass`, after logging a single actionable WARNING (set up working SSH key auth, verify with `ssh root@<host>`).
- `NonInteractiveAuthRequired` subclasses `paramiko.AuthenticationException`, so existing `except (paramiko.)AuthenticationException` / `except Exception` handlers in the connect path keep catching it — the host is cleanly reported unreachable rather than hanging.
- Thread `interactive` from `Target` into `Connection`. The two `Target(...)` construction sites in `TestReport` derive it from the prompter (`interactive=self._prompter is not None`): MCP sets `prompter=None` → non-interactive; the REPL always supplies a real `Prompter` → interactive.
- This covers **both** `add_host` paths: `connect_targets()` → `connect_target()` and `add_target()`.

Interactive REPL behaviour is unchanged (default `interactive=True` preserves the legacy `getpass` fallback).

## Tests

- `tests/test_connection.py`: non-interactive auth failure and bad-host-key both raise `NonInteractiveAuthRequired` **without** calling `getpass`; subclass-contract check.
- `tests/test_testreport.py`: `connect_target` / `add_target` build `Target` with `interactive=False` when no prompter (MCP) and `interactive=True` when a prompter is present (REPL).

## Verification

- `uv run ruff format --check .` — clean
- `uv run ty check` — clean
- `uv run pytest` — 1345 passed

### DIFF
--- a/mtui/test_reports/testreport.py
+++ b/mtui/test_reports/testreport.py
@@ -497,6 +497,7 @@ class TestReport(ABC):
                 self.packages,
                 timeout=self.config.connection_timeout,
                 prompter=self._prompter,
+                interactive=self._prompter is not None,
             )
             target.connect()
             new_system = str(target.system)
@@ -577,6 +578,7 @@ class TestReport(ABC):
                 hostname,
                 self.packages,
                 prompter=self._prompter,
+                interactive=self._prompter is not None,
             )
             self.targets[hostname].connect()
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,13 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
 
 ### Fixed
 
+- Running under MCP (`mtui-mcp`) no longer hangs when a refhost's SSH key
+  authentication fails. Previously the connect path fell back to an interactive
+  `getpass` root-password prompt, which has no TTY in MCP mode (stdin is the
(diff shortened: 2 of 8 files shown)


mtui/test_reports/testreport.py
```diff
@@ -497,6 +497,7 @@ def connect_target(
                 self.packages,
                 timeout=self.config.connection_timeout,
                 prompter=self._prompter,
+                interactive=self._prompter is not None,
```
mjdonis, line 500: one thing worth recording that this made me think of:

the prompter gets created anyway if you ran mtui with `--noninteractive --prerun`, so it would still fail in that case, but this isn't something introduced in this PR so not a blocker

PR #197, opened 2026-06-20
fix: update reports every host's outcome, not just the first failure

plusky:
## Problem

`HostsGroup.perform_update()` ran the post-update check in a plain loop:

```python
for t in self.data.values():
    t.check("updater")(t.hostname, t.lastout(), t.lastin(), t.lasterr(), t.lastexit())
```

The first host whose check raised `UpdateError` aborted the loop, so a **parallel
update that failed on several hosts surfaced only the first host's error** and the
rest went unreported. (Observed during a Product Increment run: `update` returned
a single `UpdateError(<one host>)` while 4 hosts had actually failed — the other
three silently stayed un-updated.)

## Fix

Check **every** host, collecting failures instead of aborting on the first:

- Log a per-host **success/failure summary** (`update succeeded on: …` /
  `update failed on <host>: <reason>`).
- A **single** failing host re-raises that host's **original** `UpdateError`
  unchanged (backward compatible — same error the caller used to get).
- **Multiple** failures raise an aggregated `UpdateError` naming all failed hosts
  with each reason.
- **No reboot** is attempted when any host failed (unchanged: the raise happens
  before `_reboot`). Hosts are still unlocked in the `finally`.

The caller's `UpdateError` handling (rollback + re-raise) is unaffected.

## Tests

- `test_perform_update_reports_all_host_failures` — two failing hosts → both
  logged, error names both, `_reboot` not called, both unlocked.
- `test_perform_update_single_failure_reraises_original` — one failing host →
  the exact original `UpdateError` is re-raised; passing host not reported failed.
- Existing happy-path / unlock-on-failure tests unchanged.

Gates: `ruff format`/`ruff check` clean, `pytest` green (1388 passed), no new `ty` diagnostics.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,13 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
 
 ### Fixed
 
+- A failed `update` now reports the outcome of **every** host instead of only the
+  first failure. The post-update check aborted on the first host that failed, so a
+  parallel update that broke on several hosts surfaced just one `UpdateError` and
+  the rest went unreported. mtui now checks all hosts, logs a per-host
+  success/failure summary, and raises an error naming **all** failed hosts (a
+  single-host failure still re-raises that host's original error unchanged). No
+  reboot is attempted when any host failed.
 - `mtui-mcp` now advertises `readOnlyHint=True` for the `openqa_jobs` tool (it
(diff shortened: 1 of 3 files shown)


mimi1vx:
@plusky broken CHANGELOG.md .. please fix

mimi1vx (approved):

PR #236, opened 2026-06-25
feat(teregen): consolidate report data sources via the TeReGen API

plusky:
## What

Make the **TeReGen report API** (`qam.suse.de/api/v1`) mtui's source of truth for the data previously read straight from SMELT, and add the write/queue capabilities on top.

### Data sources
- New `TeReGen` client (`data_sources/teregen.py`): `info` (main report endpoint), `metadata`, `status`, `checkers`, `updates`, `regenerate` + `wait_for_template`; best-effort (`None` on any failure), base URL from `[teregen] api`.
- `assign` surfaces an update's **priority/deadline from TeReGen**. These are read from the **main report endpoint** (`GET /reports/{id}`), where TeReGen now serves them live; `/metadata` is the on-disk file only.
- Removes the `Smelt` client, the `smelt_*` commands, and the `[smelt] url` config option.

### Commands (TeReGen-backed)
- `updates` — the update queue (`GET /updates`), sorted by priority; each row shows **priority, status, kind (SLFO/Maintenance/…), deadline** and the RRID. The queue merges gitea-sourced (SLFO/SL-Micro) with classic Maintenance updates in QAM testing. Filters: `--review-group`/`--status`/`--limit`. Replaces `smelt_updates`/`smelt_requests`.
- `checkers` — build-check result runs (`GET /reports/{id}/checkers`). Replaces `smelt_checkers`.
- `regenerate` — regenerate the loaded update's template (`POST /reports/{id}/regenerate`), wait for the Minion job, reload it. Flags `--force`, `--ignore-inconsistent`, `--no-wait`. The stale-template loader (Gitea-hash mismatch) now offers to regenerate via TeReGen, delete the local checkout and wait for the rebuild.

### Docs
- `Documentation/iui.rst` + `cfg.rst`: document the new commands and `[teregen] api`; drop the removed `smelt_*` / `[smelt] url`.

## Tests
Full suite green; new command tests (`updates`/`checkers`/`regenerate`), updated `apicall`/`updateid`/registry snapshots.

## Matches teregen
Aligned with teregen master: `/updates` merged+normalized queue (kind/deadline), live `priority`/`deadline` on `GET /reports/{id}`, `/metadata` file-only, and the `/regenerate` endpoint.

PR #239, opened 2026-06-26
feat(updates): expose and filter queue by assignee

plusky:
## What

Restores assignment visibility/filtering to the `updates` command. The old
`--unassigned`/`--show-assignment` were dropped when SMELT was removed
(#236); this brings the capability back on top of the TeReGen `/updates` API.

### TeReGen client (`mtui/data_sources/teregen.py`)

`updates()` gains four optional keyword params mapping 1:1 to the new
`GET /api/v1/updates` query params:

- `assignee=<user>` — filter to updates assigned to that user (any qam group).
- `unassigned=1` — filter to updates with no assignee.
- `with_assignment=1` — include assignment on each row without filtering.
- `no_cache=1` — bypass the server's short assignment cache.

All three booleans are only sent when `True`; `assignee` only when truthy.

### `updates` command (`mtui/commands/updates.py`)

New flags:

- `--assignee <user>` — filter to updates assigned to that user.
- `--mine` — shorthand for `--assignee <current session user>`.
- `--unassigned` — only updates with no assignee.
- `--show-assignment` — annotate every row with its assignee, no filtering.

`--mine`/`--assignee` are mutually exclusive, and `--unassigned` cannot be
combined with `--assignee`/`--mine`. Whenever any assignment flag is given the
client is called with `with_assignment=True`, and rows render
` assignee=<user>` (`assignee=unassigned` when none).

### Tests

Client-level tests assert the right query params are sent; command-level tests
cover `--show-assignment` rendering, `--mine` -> session user, `--unassigned`
pass-through, and the mutually-exclusive validation errors.

## Note

Depends on the TeReGen `/updates` assignment MR (assignee/unassigned/
with_assignment/no_cache params + assignees/assignee fields); will work once
TeReGen is deployed.

### DIFF
--- a/mtui/commands/updates.py
+++ b/mtui/commands/updates.py
@@ -43,11 +43,47 @@ class Updates(Command):
             default=0,
             help="cap the number of rows (0 = all)",
         )
+        # --assignee/--mine/--unassigned select at most one assignment view;
+        # argparse rejects any combination of the three for us.
+        assignment = parser.add_mutually_exclusive_group()
+        assignment.add_argument(
+            "--assignee",
+            default=None,
+            help="filter to updates assigned to this user (any qam group)",
+        )
+        assignment.add_argument(
+            "--mine",
+            action="store_true",
+            help="filter to updates assigned to the current session user",
+        )
+        assignment.add_argument(
+            "--unassigned",
+            action="store_true",
+            help="filter to updates with no assignee",
+        )
+        parser.add_argument(
+            "--show-assignment",
+            action="store_true",
+            help="show the assignee on each row without filtering",
+        )
 
     def __call__(self) -> None:
         """Fetch and print the update queue from TeReGen."""
+        assignee = self.args.assignee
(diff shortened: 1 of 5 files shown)


mtui/commands/updates.py
```diff
@@ -43,11 +43,53 @@ def _add_arguments(cls, parser: ArgumentParser) -> None:
...
+            action="store_true",
+            help="show the assignee on each row without filtering",
+        )
 
     def __call__(self) -> None:
         """Fetch and print the update queue from TeReGen."""
+        if self.args.mine and self.args.assignee:
+            self.println("error: --mine and --assignee are mutually exclusive")
+            return
+        if self.args.unassigned and (self.args.mine or self.args.assignee):
```
mimi1vx: I think there is better use mutuals exclusive group ?

mimi1vx (changes requested):

mimi1vx (approved):

PR #257, opened 2026-07-01
feat(commands): add request_review Slack review/approve command

plusky:
## What

New `request_review` command: commits the loaded update's testreport, posts its
`reports_url/<RRID>/log` URL to a configured Slack channel (`[slack]` config
section), then watches the thread — streaming replies and treating a 👍
reaction as review approval, driving the existing `approve` path with the
best-effort reactor recorded as reviewer. `--no-watch` posts and returns;
`--no-approve` watches and reports without approving.

`approve` and `reject` now require a recorded Slack review whose message still
carries a 👍; a REPL-only `-f/--force` bypasses the gate and is structurally
unavailable over MCP (schema-skipped **and** gated on `prompt.interactive`),
so an AI agent can never force-approve an update without a human review.

The Slack message reference (channel + ts) is persisted durably in the
testreport (`Slack Review: <channel>/<ts>` marker line, parsed back on load)
so the ack survives a reload and the gate re-checks it live.

## Design decisions

- Bot token + Web API polling (no new dependency; webhooks can't read
  reactions, Events/Socket Mode are daemon shapes).
- Reviewer name is a **best-effort** reactor: Slack does not guarantee
  `users[]` ordering; on a `users.info` failure (e.g. missing `users:read`
  scope) the raw member id is recorded instead — the ack itself is
  authoritative and is never lost to name-resolution problems.
- Fan-out: one Slack thread + one watch per loaded template, consistent with
  `approve`/`commit`.

## Robustness (from adversarial review)

- **Cooperative cancellation over MCP**: the session now binds a per-call
  `threading.Event` (`mtui.support.cancellation.current_cancel_event`,
  propagated into `asyncio.to_thread` via contextvars) and sets it on
  `job_cancel` / client disconnect; the Slack watch passes it into
  `wait_for_ack`. A cancelled background job's worker thread exits promptly
  instead of polling unobserved for hours and auto-approving on a late 👍.
- **Transient tolerance**: the watch only gives up as `unreachable` after 3
  *consecutive* failed poll cycles; a single network blip or sustained-429
  pair no longer aborts a multi-hour watch.
- **No dangling review requests**: the template's marker anchor
  (`Test Plan Reviewer:` line) is validated **before** posting to Slack, and a
  post-post marker failure is reported cleanly instead of crashing with the
  request already live in the channel.

Known interplay: with several templates loaded, Ctrl-C during one template's
watch ends that watch; aborting the whole fan-out on Ctrl-C is fixed
separately in #258 (`except BaseException` → `except Exception` in the
fan-out loop).

## Tests

63 tests across the Slack client (mocked with `responses`), the command
(failure paths, `--no-watch`/`--no-approve`, the fanout auto-approve blocker
regression, MCP cancel-event forwarding), the approve/reject gate, template
marker persistence/parsing, and the MCP job cancellation path.

Full suite green (1811 passed); ruff format/check + ty clean; patch coverage
100% on the new modules.

## Resume & repost (added after design-panel + adversarial review fan-out)

Re-running `request_review` on a report that already records a `Slack Review:`
marker **resumes** the existing request instead of posting a duplicate: the
thread's replies so far are forwarded to the user first (the watch streams
from the top), and a 👍 that arrived in the meantime approves immediately — so
an interrupted watch (Ctrl-C, a cancelled MCP job) just picks up where it left
off. `--repost` forces a fresh post, replacing the marker and leaving a
best-effort "superseded" reply under the old thread; it is refused for an
unscoped multi-template fan-out (both in the foreground `run()` and in the MCP
`start_jobs` per-template fan-out, which would otherwise bypass the check via
its `-T`-scoped jobs).

State correctness (per the design panel + review):

- The **on-disk marker is the source of truth**: a new
  `TestReport.get_slack_review()` re-parses the file (with the same regex the
  load-time parser uses) after the pre-commit's `svn up`, so a colleague's
  newer / reposted / removed marker is respected. The `approve`/`reject` gate
  now reads the marker file-fresh too.
- **Pre-approve guards**: after an ack, the auto-approve is skipped if the
  cancel event was flagged during the watch (a cancelled job's final in-flight
  poll must never approve) or if the marker — refreshed via a best-effort
  `svn up` — no longer matches the watched `(channel, ts)` (superseded
  mid-watch, including from another checkout).
- A marker-write failure after a successful post now **voids the orphaned
  message in-thread** (best-effort) and the remediation names the exact
  scoped re-run (`--repost -T <RRID>`).
- A template deleted underneath the session (`svn up` on a withdrawn report)
  reads as "no marker" instead of raising.

Deliberately out of scope (pre-existing, noted for reviewers): Approve
idempotence when the update is already approved, and fail-fast when an
in-flight watch job already holds the template's lock (documented in mcp.rst).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,26 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   unresolvable fails closed. A new `[obs]` config section (`api_url`,
   `request_timeout`) is introduced.
 
+- New `request_review` command — posts the loaded update's test report to a
+  configured Slack channel (`[slack]` config section), then watches the request
+  thread and auto-approves the update once a reviewer reacts with 👍 (skin-toned
+  👍 variants count), relaying any thread replies as they arrive. With several
+  templates loaded, all review requests are posted first and a single combined
+  watch then polls every thread, approving each update as its ack arrives;
+  Ctrl-C stops the watch with a per-template status summary. When the report
+  already records a Slack review marker, a re-run resumes the existing request
+  instead of posting a duplicate — forwarding the thread's replies so far — so
+  an interrupted watch (Ctrl-C, a cancelled MCP job) just picks up where it
+  left off; `--repost` forces a fresh post, replacing the marker and leaving a
+  "superseded" reply under the old thread. A cancelled or superseded watch
+  reports an observed 👍 but never approves from it, and a cancelled MCP job
+  stops before its next side effect — `job_cancel` reports "cancelled" only
+  once the worker has actually stopped (a `cancelling` state covers the stop
+  window), and a foreground MCP `request_review` that would watch several
+  templates is refused (background it instead). In turn, `approve` and `reject` now
+  require a recorded Slack review that still shows a 👍 — the gate verifies the
+  referenced Slack message really is this update's review request (it must
+  name the RRID) before trusting any reaction, and refuses otherwise.
 - `mtui-mcp` is now much more token-efficient. Tool schemas are automatically
   slimmed of redundant pydantic boilerplate (the per-field `title` keys and the
   `anyOf: [T, null]` optional-field unions), shrinking the tool-list payload
--- a/Documentation/cfg.rst
+++ b/Documentation/cfg.rst
@@ -520,6 +520,78 @@ port must be numeric); an invalid value is rejected when the
 configuration is read and the default is used instead.
 
 
+``slack.base_url``
+~~~~~~~~~~~~~~~~~~~~
+  | **type**
+  |     URL
+  | **default**
+  |     https://slack.com/api
+
+Base URL of the Slack Web API used by the ``request_review`` command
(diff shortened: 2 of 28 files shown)


plusky:
Closing: #337 replaced the codebase with the Rust rewrite, so this Python branch can no longer merge. The feature is being re-planned as a native Rust port (slack datasource client + request_review command on the mtui-core registry, reusing the existing MCP background-job plumbing). The semantics hardened here through review — the marker-bound approve/reject gate, the two-phase fan-out watch with shared deadline and cancel re-checks, truthful MCP job cancellation — carry over verbatim as the port's specification; this branch stays as the reference implementation.

plusky:
Superseded by #339, which ports this feature to the Rust tree (the branch was rebuilt on the Rust workspace, so GitHub would not let this PR reopen). The reviewed semantics from here carry over: marker-bound gate, canonical `(channel, ts)` persisted, skin-tone-normalised acks.

PR #260, opened 2026-07-02
fix(config): validate ssl_verify at parse time, default to system CA bundle

plusky:
## What

Field report: `[mtui] ssl_verify = false1` (a typo for `false`) in `~/.mtuirc`
crashed `mtui -a SUSE:SLFO:1.2:5797` with a raw traceback ending in
`OSError: Could not find a suitable TLS CA certificate bundle, invalid path:
false1` — the string flowed verbatim into `requests` as a supposed CA-bundle
path and detonated at the first HTTPS call (Gitea `get_hash` during template
load). The `OSError` is not a `RequestException`, so it bypasses every
consumer's error handling: raw traceback in gitea/openQA paths, a
never-collected future (fully silent) in the export log downloader, a silent
degrade to the possibly-stale local `refhosts.yml` in the refhosts resolver.

A second, related field failure: with `ssl_verify` unset and
`ca-certificates-suse` correctly installed, internal hosts (src.suse.de)
still failed verification when running mtui from a git checkout — `requests`
validates against PyPI `certifi`'s bundled CAs, never the system trust store
(distribution `python-certifi` packages patch this away, which is why RPM
installs were unaffected).

## Change

- `_parse_ssl_verify` keeps the boolean spellings byte-for-byte, expands `~`,
  and requires any other value to be the path of an **existing** file or
  directory; otherwise it raises with the accepted forms. Riding the config
  machinery's documented fallback contract, that surfaces as **one**
  actionable ERROR at parse time and the option falls back to the verifying
  default — the codebase's established policy for bad config values.
- The generic config-failure log is now a single clean line (`logger.error`
  with the exception text inline) instead of a full traceback
  (`logger.exception`); the traceback moved to DEBUG.
- New `system_ca_bundle()`: when `ssl_verify` is unset, verification prefers
  the distribution CA bundle (`/etc/ssl/ca-bundle.pem` on SUSE, plus the
  Debian/RHEL/Alpine equivalents), falling back to `True` (certifi) when none
  exists. System-installed CAs now work from a checkout exactly as from an
  RPM install.
- `ssl_verification_hint` names the concrete `ssl_verify = <system bundle>`
  remedy and no longer implies that installing the CA system-wide alone is
  sufficient (it wasn't, pre-fix, for checkout runs).
- Docs (`cfg.rst`), option comment, and CHANGELOG updated.

Verified end-to-end on a machine with the SUSE CA installed: the `false1`
config now yields one ERROR + fallback to `/etc/ssl/ca-bundle.pem`, and an
unset config fetches `https://src.suse.de` successfully through mtui's HTTP
path.

## Follow-up candidates (found during the investigation, out of scope here)

- `openqa_instance`/`openqa_instance_baremetal`: malformed URL escapes the
  connector's handlers as `requests.InvalidURL` at first query.
- `install_logs`/`template_dir`: bad paths crash `mkdir` after a successful
  checkout / outside svn error handling.
- `connection_timeout` accepts negative values, misattributed later as an SSH
  banner error.
- REPL `config set ssl_verify ...` bypasses fixups entirely (generic to all
  options).
- The export log downloader never collects its futures, so *any* exception
  there is silently dropped.

## Rework (response to the adversarial review — 10 findings, all addressed)

The first iteration had real defects; this rework addresses every one:

1. **`config set ssl_verify false` stored the literal string** (the runtime
   set path infers coercion from the current attribute's type, which became
   `str`): fixed properly by the second commit — `config set` now routes
   every declared option through its ConfigOption getter semantics + fixup,
   which also fixes the pre-existing `config set use_keyring true` →
   **False** bug and makes bad runtime values rejected with the option
   unchanged.
2. **Relative CA paths broke after `chdir_to_template_dir`**: paths are now
   absolutised at parse time.
3. **Blank `ssl_verify =` silently flipped from verify-off to verify-on**:
   blank keeps its historical meaning (off) and logs a warning.
4. **Explicit `true` diverged from unset** (kept certifi, losing the
   system-bundle fix): `true` is now deliberately identical to unset.
5. **Un-rehashed cert directories validated but could never verify**: a
   directory is only accepted when it contains OpenSSL hash-named entries;
   otherwise rejected with a message naming `c_rehash`.
6. **The failure hint recommended the exact bundle that just failed**: the
   custom-bundle example is now generic; installing the CA into the system
   store (which regenerates the default bundle) is the primary remedy.
7. **Tracebacks lost for unexpected parser errors**: only `ValueError`
   (validation) gets the clean one-line treatment; any other exception keeps
   `logger.exception` with the full stack.
8. **Hand-rolled bundle probe**: `system_ca_bundle()` now prefers
   `ssl.get_default_verify_paths().cafile` (the interpreter's real OpenSSL
   bundle, honouring `SSL_CERT_FILE`), with the well-known distro paths as
   fallback only.
9. **Late-mounted CA path rejected at startup** (NFS/automount): accepted
   trade-off — the rejection logs exactly why and falls back to the
   verifying default; create the file before launch or use `SSL_CERT_FILE`.
10. **Stale system bundle preferred over certifi**: accepted trade-off — on
    the target audience's systems the distribution bundle includes the
    Mozilla roots and is what every other system tool trusts.

### DIFF
--- a/mtui/support/config.py
+++ b/mtui/support/config.py
@@ -6,6 +6,7 @@ overriding configuration options with command-line arguments.
 
 import configparser
 import getpass
+import re
 from argparse import Namespace
 from collections.abc import Callable
 from dataclasses import dataclass, field
@@ -14,6 +15,7 @@ from os import getenv
 from pathlib import Path
 from typing import Any
 
+from .http import system_ca_bundle
 from .paths import terms_path
 
 logger = getLogger("mtui.config")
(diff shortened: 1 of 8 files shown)


mtui/support/config.py
```diff
@@ -6,6 +6,8 @@
 
 import configparser
 import getpass
+import os.path
```
mimi1vx: use pathlib and Path

mimi1vx (changes requested):

mtui/support/config.py
plusky: Done — the validation now uses pathlib.Path throughout (import dropped).

mimi1vx (approved):
Rework addresses the review: 'import os.path' is gone, validation uses pathlib.Path throughout. All CI green. LGTM.

PR #261, opened 2026-07-02
fix(config): validate endpoint URLs, positive ints and install_logs at parse time

plusky:
## What

Follow-up to #260, hardening the rest of the config options whose invalid
values surfaced as late tracebacks (or silent misbehavior) instead of a clear
parse-time error. All fixes ride the same tested fallback contract as #260:
an invalid value logs **one** actionable ERROR naming the option, the bad
value and the accepted forms, then falls back to the declared default.

- **Endpoint URLs** (`openqa.openqa`, `openqa.baremetal`, `url.qem_dashboard_api`,
  `url.teregen_api`, `refhosts.https_uri`): new `_parse_base_url` fixup requires
  an http(s) scheme, a non-empty host, and a parseable port. Previously e.g.
  `openqa = https://openqa.suse.de:44e3` escaped as `requests.InvalidURL` at the
  first query, outside every handler.
- **Positive integers** (`connection_timeout`, `lock.wait_poll`,
  `mcp.session_cap`, `mcp.session_idle_timeout`, `refhosts.https_expiration`):
  zero/negative rejected. A negative `connection_timeout` previously reached
  paramiko and failed every host with a misleading "Error reading SSH protocol
  banner". Documented zero-semantics (`lock.wait`, `lock.stale_age`) untouched.
- **`install_logs`**: must be a single relative path component (it is joined as
  `template_dir/<rrid>/install_logs`; a nested value crashed `mkdir` after a
  successful checkout, and an absolute value silently replaced the whole base
  in the pathlib join).
- **`template_dir`**: an unusable path (unwritable parent, path through a file)
  now raises a single actionable `TemplateDirNotUsableError` naming the option
  instead of a raw `OSError` outside the svn error handling.
- **Defensive**: `openqa` `_get_jobs` also catches `requests.exceptions.RequestException`
  so no URL/transport shape escapes as a traceback.

## Notes

Stacked on #260 (`fix(config): validate ssl_verify at parse time...`) — the
first commit here is that PR's commit; review only the second. Will be rebased
onto main once #260 merges.

Full gates green: ruff format/check, ty, 1795 tests, sphinx docs build.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -417,6 +417,23 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   config-file parse time, falling back to the verifying default). An invalid
   value is rejected with a single actionable error and the option is left
   unchanged.
+- More config options are validated when the configuration is read, each
+  invalid value producing one actionable error and falling back to its default
+  instead of a delayed crash: the endpoint URLs (`[openqa] openqa`/`baremetal`,
+  `[qem_dashboard] api`, `[teregen] api`, `[refhosts] https_uri`) must be
+  http(s) URLs with a host and a numeric port (a typo like
+  `https://openqa.suse.de:44e3` used to surface as a raw `InvalidURL`
+  traceback at the first query), the duration/count options
+  (`connection_timeout`, `[lock] wait_poll`, `[mcp] session_cap` /
+  `session_idle_timeout`, `[refhosts] https_expiration`) must be positive
+  integers (a negative `connection_timeout` reached paramiko and failed every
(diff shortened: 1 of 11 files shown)


mimi1vx (approved):
LGTM. Config options now fail fast at parse time with a clear log line and safe fallback to defaults; validators consistently raise ValueError, caught and handled in _parse_config. New TemplateDirNotUsableError follows the existing ErrorMessage pattern and is caught in both checkout paths. The added requests.exceptions.RequestException handler in _get_jobs is additive and correctly ordered. CHANGELOG under [Unreleased], docs updated, coverage matches the contract. All CI checks green.

PR #263, opened 2026-07-02
fix(cli): render mid-spin log lines flush-left, coordinating spinner and logging

plusky:
## What

Field report: during `update`, the first worker-thread log line rendered with
~19 spaces of leading padding:

    info: Updating
                       info: Removing repo .../SLES-16.0-aarch64 on lorikeet...
    info: Removing repo .../SLES-16.0-ppc64le on bianca...

## Root cause (proven with a pty byte capture)

`ColorFormatter` prepends `\x1b[2K` (erase entire line) to the levelname to
clear a live spinner frame — but EL2 does not move the cursor. The spinner had
just painted `\r[|] set_repo remove` (exactly 19 columns), so the erased line's
text started at column 19. Follow-up lines land flush-left only because all
four workers log within one ~1 ms burst while the spinner repaints every
100 ms — the artifact recurs on the first log line of every spinner-wrapped
phase. With `NO_COLOR` on a TTY the erase was skipped entirely, leaving the
raw frame text in front of the line.

## Change

Screen management moved out of the formatter into a coordinated pair:

- `mtui/support/spinner.py`: a module-level paint lock + active-spinner
  registry; frames paint and erase under the lock, and a new
  `spinner_suspended()` context manager erases the visible frame with
  `\r\x1b[K` (CR first — the missing piece) and pauses repainting while held.
  Off a TTY spinners never register, so it writes nothing.
- `mtui/cli/colors/formatter.py`: `create_logger` installs a
  `SpinnerAwareStreamHandler` whose `emit()` runs under `spinner_suspended()`;
  the `\x1b[2K` injection is deleted from both format branches (this also
  fixes the `NO_COLOR`-on-TTY variant).
- `mtui/cli/prompter.py`: the interactive prompt suspends the spinner the same
  way instead of hand-rolling the erase.

Regression tests capture raw bytes through a pty (the investigation's repro
technique): a log line emitted mid-spin must be preceded by a clean
line-erase and render flush-left, in `always` and `never` color modes; the
no-TTY paths stay byte-identical no-ops.

Full gates green: ruff format/check, ty, 1748 tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -434,6 +434,15 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   (e.g. a plain file in the way) is now reported as one clear error naming the
   option instead of an `OSError` traceback, and openQA queries catch any
   remaining URL/transport error shape instead of crashing the command.
+- Log lines emitted while the TTY spinner is painting (e.g. the per-host
+  `Removing repo … on <host>` lines under the `set_repo remove` spinner during
+  `update`/`prepare`) no longer render with phantom leading padding — or, with
+  colours off, glued to the leftover frame text. The spinner and the logging
+  handler now coordinate through a shared paint lock: each record first erases
+  the live frame and homes the cursor to column 0, then prints, and the spinner
+  repaints on its next tick. Interactive prompts raised under a live spinner
+  (e.g. the SSH command-timeout question) also pause the spinner for the whole
+  read instead of repainting over the prompt. Off a TTY nothing changes.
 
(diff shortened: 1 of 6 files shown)


mimi1vx (dismissed):

mimi1vx (approved):
Rebased onto current main (conflict in CHANGELOG.md [Unreleased] resolved additively — kept both the config-validation and spinner/log entries). Full gate set green locally on the whole repo: ruff format --check, ruff check, ty check, and pytest (1841 passed). The spinner/logging coordination via the shared paint lock correctly closes the erase/repaint race.

PR #264, opened 2026-07-02
ci: enforce docs build, dedupe triggers, add concurrency and permissions

plusky:
Same-repo PR branches ran every commit twice because `push` and `pull_request` both fired. This PR trims the CI workflow and adds a docs gate:

- **Dedupe triggers**: restrict `push` to `main` (plus tags) and keep plain `pull_request`, so PR branches on the main repo no longer run every commit twice.
- **Least-privilege token**: grant the `GITHUB_TOKEN` only `contents: read`.
- **Concurrency**: cancel stale PR runs on force-push via a per-ref concurrency group; main/tag runs are never cancelled so their Codecov upload lands.
- **Docs job**: build the Sphinx documentation with `-W` (warnings as errors), so documentation regressions fail CI instead of surfacing at release time.
- Minor cleanup: trailing whitespace on the codecov `if:` lines, long `run:` lines folded with `>-`.

Separately, `.mergify.yml` drops `check-success=codecov/patch` from the automerge base checks: config-only PRs produce no patch coverage status, which wedges the automerge rule. `codecov/project` remains as the required final check.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,17 @@
 ---
 name: ci
-# yamllint disable-line rule:truthy
-on: [push, pull_request]
+on:
+  push:
+    branches: [main]
+    tags: ['**']
+  pull_request:
+permissions:
+  contents: read
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  # Only PR runs are cancelled by a newer push; a main/tag run must finish
+  # so its Codecov upload lands (PR comparisons need the main baseline).
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 jobs:
   lint:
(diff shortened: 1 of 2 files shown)


mimi1vx (changes requested):
timeouts .. not needed
yamllint ? Why ?
and mergify ? remove check-success/patch

plusky:
Reworked as requested: dropped all the job timeouts, removed yamllint entirely (.yamllint, the pre-commit hook, and the codecov.yml/mergify reformatting that only existed to satisfy it), and reverted .mergify.yml to its original formatting with a single change — `check-success=codecov/patch` removed from the automerge base checks. The remaining scope is the trigger dedupe, `contents: read` permissions, the concurrency group, and the new docs job.

mimi1vx (approved):

PR #273, opened 2026-07-08
fix(dashboard): drop obsoleted openQA jobs and stop the false "All jobs passed"

plusky:
## What

Two co-located correctness bugs in the dashboard-backed auto-workflow openQA provider (`DashboardAutoOpenQA`), both in how it interprets openQA job results. They're bundled because they're the same file/subsystem and complementary — #8 removes one source of #7, and #7 still stands on its own for genuinely running/`parallel_failed` jobs.

### Finding #8 — obsoleted (superseded) jobs were never filtered

`_normalize_job` captured the dashboard's `obsolete` flag but nothing used it. When an openQA job is retriggered the dashboard keeps the older run, so its **stale result stayed in the working set** — folded into the install verdict (`_has_passed_install_jobs`) *and* listed as a phantom failure. A `qam-incidentinstall` that failed, was retriggered, and now passes was therefore reported as "no successful install", and its dead run showed up under *Failed jobs*.

**Fix:** new `_is_obsolete(job)` (`bool(job.get("obsolete")) or job.get("result") == "obsoleted"`); `_load_jobs` drops obsolete normalized jobs (with a debug log) instead of `jobs.extend(...)`. Mirrors `oqa_search.incident_jobs`, which already filters `result == "obsoleted"`. `_load_jobs` is the single choke point every consumer (`.jobs`, `.results`, `.pp`) reads from.

### Finding #7 — "All jobs passed." printed while the Summary lists a problem group

`_pretty_print_section` gated its trailer on `failed_by_group`, which is populated only for `FAILED_RESULTS` (`failed`/`incomplete`/`timeout_exceeded`). A group whose only problem is in the `other` bucket — a still-running job (`result="none"`), `parallel_failed`, `skipped`, … — is flagged as a problem by `_has_problems` (it counts `other`) and shown in the Summary, yet the section then printed **"All jobs passed."**: a self-contradiction a reviewer could act on.

**Fix:** gate the trailer on `problem_keys`. Print the per-job *Failed jobs* block when there are real failures; a `"some groups need review (see Summary above)"` note when problem groups exist but carry no `FAILED_RESULTS` job (so no empty *Failed jobs* header either); and keep `"All jobs passed."` only when there are genuinely no problem groups.

## Tests

- `test_is_obsolete_flags_superseded_runs` — the flag and the `obsoleted` result both mark a stale run; a live `failed` run does not.
- `test_load_jobs_drops_obsoleted_runs` — HTTP-mocked `.run()`: a superseded failed install run (by flag) and an `obsoleted`-result run are dropped, so only the current passing run survives, the verdict stays passed, and neither appears as a phantom failure. **Revert-verified**: restoring `jobs.extend(...)` fails this test.
- `test_pretty_print_parallel_failed_not_reported_as_all_passed` / `..._running_job_...` — a `parallel_failed` (distinct group) and a running `none` job each flag a problem group; the section no longer claims all-passed and emits the review note instead. **Revert-verified**: restoring the bare `else` fails both.

Patch coverage of the added lines is complete (branch coverage on).

## Gates

`ruff format --check .`, `ruff check .`, `ty check`, `pytest` (1845 passed) — all green on the whole repo.

Resolves findings #7 and #8 from the recent code review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -349,6 +349,18 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   logged and abandoned so `close()` — and the http registry's idle-sweep behind
   it — always returns. Previously the bound was defeated by the thread pool's
   context-manager exit, which re-joined every worker regardless of the timeout.
+- The dashboard-backed auto-workflow openQA report no longer counts superseded
+  (obsoleted) jobs. When a job is retriggered the dashboard keeps the older run;
+  its stale result was still folded into the install verdict and listed as a
+  phantom failure. Obsoleted runs (marked by the dashboard's `obsolete` flag or
+  an `obsoleted` result) are now dropped, matching the openQA-search connector,
+  so a retriggered-and-now-passing install is reported as passed and only the
+  current run per scenario is shown.
+- The dashboard-backed auto-workflow openQA section no longer prints
+  "All jobs passed." while its own Summary lists a problem group. A group whose
+  only problem is a still-running, `parallel_failed`, or otherwise non-failed
+  job (i.e. it has no failed/incomplete/timeout job) is now reported with a
+  "some groups need review" note instead of a contradictory all-passed verdict.
 - An unscoped multi-template fan-out of a host-phase command (e.g. `lock`,
   `run`) no longer fails the whole fan-out — or, for `lock`, pretends to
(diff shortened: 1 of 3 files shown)


mimi1vx (approved):
Reviewed: drops obsoleted openQA jobs + fixes false 'All jobs passed', with regression tests, CI green. Rebased CHANGELOG onto main. Approving.

PR #275, opened 2026-07-09
fix(connection): decode command and shell output tolerantly, not strict UTF-8

plusky:
## What

A test command whose output contains a single non-UTF-8 byte was recorded as **failed with its output lost** — even when it exited 0. `Connection.run()` accumulates raw stdout/stderr bytes and decoded them strictly at the end; a Latin-1 locale, `rpm` metadata with odd bytes, or `cat` of a binary raised `UnicodeDecodeError` out of `run()`. `Target.run()`'s broad `except` swallowed it, logged "failed to run command", set exitcode `-1`, and left the captured output empty (it is reset on entry and was never assigned).

The per-line debug-logging path a few lines above already decoded tolerantly (`errors="ignore"`) — only the final decode was strict.

## Fix

Decode the accumulated stdout/stderr with `errors="replace"`: undecodable bytes become U+FFFD and the surrounding output survives. `replace` rather than the logging path's `ignore` — a visible replacement character is better evidence than silently dropped bytes.

The interactive `shell()` mirror one function below had the same defect class (flagged in adversarial review), with a sharper edge: it strictly decoded each 1024-byte PTY chunk separately, so even a *valid* multibyte character straddling a chunk boundary raised and killed the interactive session mid-use. It now feeds chunks through an incremental UTF-8 decoder (`errors="replace"`), which buffers split sequences across chunks.

## Tests

- `test_run_tolerates_non_utf8_output` — a command emitting `b"caf\xe9 ok\n"` / `b"warn\xff\n"` completes with exit code 0 and both streams captured (with U+FFFD in place of the bad bytes). **Revert-verified**: on the old strict decode the test dies with `UnicodeDecodeError`.
- `test_shell_tolerates_split_multibyte_and_invalid_bytes` — `é` split across two recv chunks survives intact and a lone `0xFF` is replaced, not fatal. **Revert-verified.**

## Gates

Resolves finding #9 from the recent code review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -223,6 +223,17 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   filter leaves nothing, the message now distinguishes an in-progress build
   ("No failed openQA jobs ...; 12 of 12 still pending") from a build with no
   jobs at all.
+- A command whose output contains non-UTF-8 bytes (a Latin-1 locale, `rpm`
+  metadata with odd bytes, `cat` of a binary) is no longer recorded as
+  failed. The final decode of the captured stdout/stderr was strict, so a
+  single invalid byte raised `UnicodeDecodeError` out of the SSH run; the
+  command — which may have exited 0 — was then logged as "failed to run"
+  with exit code -1 and its whole output lost. Undecodable bytes are now
+  replaced (U+FFFD) and the rest of the output is preserved, matching the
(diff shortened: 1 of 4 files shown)


mimi1vx (dismissed):
Reviewed: focused bug fix with a regression test, CI green. Approving.

mimi1vx (approved):
Reviewed: tolerant UTF-8 decode with regression tests, CI green. Rebased CHANGELOG onto main. Approving.

PR #277, opened 2026-07-09
fix(export): stop kernel re-export deleting the openqa_overview block

plusky:
## What

Re-exporting a kernel-workflow testreport silently deleted the injected `openqa_overview` block. `KernelExport.run()` ordered `inject_overview()` **before** `kernel_results()`:

- First export: `kernel_results` finds and removes only the `(put your details here)` placeholder — the overview block survives.
- Every export after that: the placeholder is gone, so `kernel_results`' except-branch bulk-deletes the whole regression-section body up to `build log review:` — exactly the range the marker-bounded overview block had just been spliced into. The overview data vanished on the second export and never came back.

## Fix

Reorder `run()`: `kernel_results()` first, `inject_overview()` after. The bulk delete now clears only the previous run's results (plus any stale overview block from an older export), then the fresh overview is inserted into the cleaned section. Layout is unchanged (kernel results, then overview, then `build log review:`).

## Tests

- `test_reexport_keeps_openqa_overview_block` — drives the **real `run()` pipeline twice** (network/filesystem stages patched out) like a tester exporting, refreshing, and exporting again: exactly one overview block and one results run must survive the re-export. **Revert-verified**: with the old ordering the block disappears on the second export.

## Gates

Resolves finding #13 from the recent code review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -242,6 +242,13 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   fall back to an empty list, like the sibling `jira`/`bugs`/`packages` keys
   already did. An explicitly null `repositories` key — which crashed the
   parse itself (`frozenset(None)`) — is guarded the same way.
+- Re-exporting a kernel-workflow testreport no longer silently deletes the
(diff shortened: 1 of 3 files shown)


mimi1vx (approved):

PR #278, opened 2026-07-09
fix(mcp): re-validate staleness before each idle-sweep eviction

plusky:
## What

The `mtui-mcp` http idle sweeper could tear down a session that a client had just re-activated. `_sweep_loop` snapshots the stale keys, then evicts them one by one; `evict()` awaits `McpSession.close()`, which can take seconds (host disconnects). While one eviction was blocked in `close()`, `get_or_create()` could hand a *later* stale-listed session back to a client — refreshing its last-touch — yet the sweep still reached that key and evicted it, closing SSH host connections out from under the client's in-flight command.

## Fix

Re-validate staleness immediately before each eviction: skip the key if its last-touch was refreshed since the snapshot (or it is already gone). The re-check and `evict()`'s pop run with **no await in between**, so on the single-threaded event loop a refresh cannot slip into that gap — the previous exposure window was the full duration of every earlier `close()` in the round.

## Tests

- `test_sweep_spares_session_reactivated_during_the_sweep` — drives the **real sweeper**: two aged sessions, the first wedged in a slow `close()` (event-gated, no timing sleeps for the interleaving); mid-sweep a client re-acquires the second session. The first must be reaped, the second must survive un-closed. **Revert-verified**: on the old code the second session is evicted and its `close()` called.

## Gates

Resolves finding #24 from the recent code review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -249,6 +249,14 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   exactly in that range: it survived the first export and vanished on every
   one after. The overview is now injected after the results stage, so both
   the refreshed kernel results and the overview block survive re-exports.
+- The `mtui-mcp` http idle sweeper no longer tears down a session that a
+  client re-activated while the same sweep round was running. The sweep
+  snapshotted the stale keys and then evicted them one by one; while one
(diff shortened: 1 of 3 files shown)


mimi1vx (approved):

PR #279, opened 2026-07-09
fix(mcp): refcount the mtui logger level across concurrent captures

plusky:
## What

Concurrent `mtui-mcp` commands raced on the process-global `mtui` logger level, silently dropping captured `INFO` lines from a reply. `_run_sync` lowers the logger to INFO around each command's log capture and restored a plain saved value in `finally`. Commands overlap (different per-RRID locks; different client sessions), so the dance raced:

1. Thread A lowers `WARNING → INFO`.
2. Thread B observes an effective INFO and records nothing to restore.
3. A finishes and puts WARNING back **while B is still running**.
4. Every INFO record B logs after that is filtered at the logger before B's capture handler sees it — missing from B's MCP reply.

## Fix

Reference-count the lowering (module-level `_push_capture_level`/`_pop_capture_level` under a `threading.Lock`): the first concurrent capture lowers and remembers the prior level, the last one restores it. A user's `set_log_level` choice is still honoured — restored exactly once, to the exact prior value.

## Tests

- `test_concurrent_captures_keep_info_after_first_call_finishes` — interleaves two **real `run_command` calls** on two sessions using events (deterministic, no timing sleeps): the first command finishes mid-way through the second, which then logs INFO; the record must still reach the second reply, and the original level must be restored at the end. **Revert-verified**: on the old save/restore the INFO line vanishes.

## Gates

Resolves finding #25 from the recent code review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -257,6 +257,14 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   still closed it, cutting SSH host connections out from under the active
   client's command. Staleness is now re-checked immediately before each
   eviction, so a just-re-activated session is spared.
+- Concurrent `mtui-mcp` commands no longer race on the shared `mtui` logger
+  level, which could silently drop captured `INFO` log lines from a reply.
+  Two overlapping calls (different templates or different client sessions)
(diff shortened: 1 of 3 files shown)


mimi1vx (approved):

PR #280, opened 2026-07-09
fix(config): treat blank ssl_verify as unset instead of disabling TLS

plusky:
## What

A blank `[mtui] ssl_verify` value (`ssl_verify =` — almost always an unfinished edit, not intent) disabled TLS certificate verification for **every** outbound HTTPS call (openQA, QEM dashboard, TeReGen, Gitea). `_parse_ssl_verify` returned the stripped empty string, which reached `requests` as a falsy `verify` — CERT_NONE. Since #260 this at least warned, but the warning still left TLS off; the review flagged the remaining downgrade as the one security finding (#26).

## Fix

Blank is now treated exactly like an **unset** option: it falls back to the verifying default (`_default_verify()` — system CA bundle, else certifi `True`), keeping a warning that points at `ssl_verify = false` as the explicit way to disable. Only the documented false spellings turn verification off.

This intentionally changes the historical blank-disables behaviour: an accidental empty value must fail **secure**, and anyone who genuinely wants verification off has an explicit spelling for it.

## Tests

- `test_ssl_verify_blank_treated_as_unset_stays_secure` and `test_parse_ssl_verify_blank_stays_secure_with_warning` — rewritten from the two tests that pinned the insecure behaviour; blank now yields the verifying default plus the warning. **Revert-verified.**
- `test_ssl_verify_blank_prefers_system_bundle` — the blank fallback is identical to unset (prefers the system CA bundle).

## Gates

Resolves finding #26 (security) from the recent code review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -265,6 +265,13 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   still running, filtering its remaining INFO records before the capture
   handler saw them. The lowering is now reference-counted, so the level is
   restored only when the last concurrent capture ends.
+- A blank `[mtui] ssl_verify` value (`ssl_verify =`, usually an unfinished
+  edit) no longer turns TLS certificate verification off. The empty string
+  used to reach `requests` as a falsy `verify` — silently disabling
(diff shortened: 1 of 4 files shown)


mimi1vx (approved):

PR #281, opened 2026-07-10
fix(cli): skip malformed history lines instead of aborting list_history

plusky:
## What

One malformed line in `/var/log/mtui.log` aborted the entire `list_history` command. The per-line parse guards `IndexError` (too few fields) with `continue`, but the timestamp conversion sat outside that guard: a line whose first colon-delimited field is not a parseable float raised `ValueError` out of `list_history`, dropping every remaining host's history.

This happens in practice: the log is appended to over sftp by independent mtui sessions against shared reference hosts, so concurrent writes can interleave/tear a line — and the file may be hand-edited or rotated.

## Fix

Move the conversion (`float` + `datetime.fromtimestamp`) inside the guard and broaden it to the conversion's own failure modes: `ValueError` (non-numeric field) plus `OverflowError`/`OSError` (an absurd numeric value out of `fromtimestamp`'s range). A bad line is now skipped exactly like the too-few-fields case; good lines before and after it still print.

## Tests

- `test_list_history_skips_malformed_lines` — a torn line (`update:foo: 1720166400:...`), a no-colon fragment, and an absurd timestamp are each skipped while the surrounding good entries render. **Revert-verified**: the old code dies on the first unguarded conversion.

## Gates

Resolves finding #14 from the recent code review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -272,6 +272,12 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   Gitea). A blank value is now treated exactly like an unset option: the
   verifying default applies and a warning explains how to disable
   verification explicitly (`ssl_verify = false`) if that was really meant.
+- `list_history` no longer aborts the whole listing when `/var/log/mtui.log`
+  contains a malformed line. The log is appended to over sftp by independent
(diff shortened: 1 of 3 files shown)


mimi1vx (approved):

PR #282, opened 2026-07-10
fix(commands): stop list_packages printing literal 'None' for absent packages

plusky:
## What

With an update loaded, `list_packages -p <pkg>` for a package that is not part of the update printed the literal word **`None`** in the state column. The `KeyError` branch (package absent from `target.packages`) set `state = None`, which the row formatter rendered verbatim — while the parallel no-metadata branch already uses `""` for exactly the same no-state case.

## Fix

Mirror the no-metadata branch **exactly** in the `KeyError` branch: `state = "" if v else self.state_map[None]` — blank when the package is installed, blue `not installed` when the version querier returned `None` for it (adversarial review caught that a bare `""` was only half the parity). And since a not-installed package has no version, the version column no longer renders its `None` as a literal `None` either — in any branch.

## Tests

- `test_list_packages_blank_state_for_package_not_in_update` — a queried, installed package outside the update renders with a blank state; `None` appears nowhere. **Revert-verified.**
- `test_list_packages_not_installed_state_for_absent_package` — a queried package neither in the update nor installed reads `not installed`, with no literal `None` anywhere (state or version column). **Revert-verified.**

## Gates

Resolves finding #17 from the recent code review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -278,6 +278,11 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   hand-edited or rotated); a line whose first field is not a parseable
   timestamp used to raise out of the command, dropping every remaining
   host's history. Such lines are now skipped like the too-few-fields case.
+- `list_packages -p <pkg>` no longer prints the literal word `None` for a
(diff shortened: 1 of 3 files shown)


mimi1vx (approved):

PR #283, opened 2026-07-10
fix(hosts): record install/uninstall host history again

plusky:
## What

`install` and `uninstall` events were **silently never recorded** in the per-host history (`/var/log/mtui.log`, shown by `list_history`). `Target.add_history` builds its log line with `":".join(comment)`, which needs flat strings — but `perform_install`/`perform_uninstall` passed a *nested* list (`["install", packages]`, where `packages` is argparse's `nargs="+"` list). The join raised `TypeError` inside a `try` whose bare `except Exception: pass` swallowed it: nothing was written, nothing surfaced. `update`/`downgrade` (all-string lists) recorded fine, which is why the gap went unnoticed.

## Fix

- Callers pass flat strings: `add_history(["install", " ".join(packages)])` (and uninstall likewise).
- `add_history`'s wrong `comment: str` annotation corrected to `list[str]`.
- The silent swallow around the write is now a **warning log** (`exc_info` included). Best-effort semantics are kept — a read-only or full remote filesystem still never breaks the operation — but the next such programming error will be visible instead of hidden for years.

## Tests

- `test_perform_install_records_history_and_delegates` / `..._uninstall_...` — the history entry is flat strings whose join is exactly `install:bash curl`. **Revert-verified** against the nested-list callers.
- `test_add_history_writes_install_entry` — the exact caller shape produces a written line ending `:install:nginx curl\n`.
- `test_add_history_write_failure_warns_but_does_not_raise` — replaces the old "silently swallowed" test; pins the new warning. **Revert-verified** against the bare `pass`.

## Gates

Resolves finding #21 from the recent code review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -283,6 +283,13 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   now behaves exactly like the no-update case: blank when the package is
   installed, `not installed` when it is not — and a not-installed package's
   empty version no longer renders as a literal `None` either.
+- `install` and `uninstall` events are recorded in the per-host history
+  again (`/var/log/mtui.log`, shown by `list_history`). The history writer
+  was handed a nested package list, the resulting `TypeError` was silently
+  swallowed, and no install/uninstall entry was ever written — unlike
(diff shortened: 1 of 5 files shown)


mimi1vx (approved):

PR #284, opened 2026-07-10
fix(support): write atomic files as UTF-8, not the locale codec

plusky:
## What

`atomic_write_file` decodes incoming bytes as UTF-8 but wrote them via `os.fdopen(fd, "w")` with **no explicit encoding** — the process locale codec. Every read path is explicitly UTF-8 (`FileList.load` opens with `encoding="utf-8"`; downloaded payloads are UTF-8 documents), so the write was asymmetric. Under a non-UTF-8 locale (a genuinely installed C/latin locale, or `PYTHONUTF8=0` / `PYTHONCOERCECLOCALE=0` automation), writing a testreport template containing non-ASCII (bug summaries, maintainer names) raised `UnicodeEncodeError` out of the `FileList` context manager — losing the user's edits — or silently wrote mojibake that the next UTF-8 load misread.

## Fix

`os.fdopen(fd, "w", encoding="utf-8")` — the write now matches the read path byte for byte, regardless of locale.

Adversarial review caught the one consumer whose *read* side also used the locale codec: the refhosts.yml store (the HTTPS resolver persists the downloaded cache through `atomic_write_file`), which would have mis-decoded the now-UTF-8 cache under a non-UTF-8 locale. It now opens with explicit UTF-8 as well — which also covers user-supplied `refhosts.yml` files.

## Tests

- `test_atomic_write_non_ascii_survives_non_utf8_locale` — reproduces the original failure in a **subprocess** with `LC_ALL=C`, `PYTHONUTF8=0`, `PYTHONCOERCECLOCALE=0` (locale and UTF-8 mode are fixed at interpreter startup, so it can't be simulated in-process); asserts the write succeeds and the on-disk bytes are UTF-8. **Revert-verified**: the old code dies with `UnicodeEncodeError` in that environment.
- `test_atomic_write_non_ascii_is_utf8_on_disk` — byte-for-byte on-disk assertion in the normal environment.
- `test_parse_refhosts_reads_utf8_regardless_of_locale` — a UTF-8 refhosts.yml with a non-ASCII hostname parses correctly in the same C-locale subprocess environment. **Revert-verified** against the encoding-less `open()`.

## Gates

Resolves finding #27 from the recent code review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -290,6 +290,14 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   `update`/`downgrade`, which recorded fine. The callers now pass flat
   strings, and a failed history write logs a warning instead of hiding
   the error (the write stays best-effort and never breaks the operation).
+- Writing a testreport (and every other atomic file write, e.g. the
+  refhosts.yml cache) now always encodes UTF-8 on disk, matching the UTF-8
+  read path. The write previously used the process locale codec, so under a
+  non-UTF-8 locale a template containing non-ASCII content (bug summaries,
+  maintainer names) either failed with `UnicodeEncodeError` — losing the
+  edits — or silently wrote mojibake that the next load misread. The
+  refhosts.yml store now reads with explicit UTF-8 to match (it read with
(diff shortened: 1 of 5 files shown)


mimi1vx (approved):

PR #285, opened 2026-07-10
fix(commands): stop commit -m storing literal quotes in the SVN log

plusky:
## What

`commit -m fix the thing` recorded the SVN log message as `"fix the thing"` — with the surrounding double quotes stored verbatim in the commit. The `-m` branch manually wrapped the joined message in `"` for a shell that is never involved: `svn_commit_testreport` passes the argv list straight to `subprocess.check_call(["svn", "ci", *msg])`. The no-`-m` default path and `approve.py`'s SVN commit both pass their messages unquoted; the wrapping was a leftover from a former shell-based invocation.

## Fix

Drop the manual quoting: `msg = ["-m", " ".join(self.args.msg[0])]`.

## Tests

- `test_commit_explicit_message_passed_through` — updated from pinning the quoted behaviour to pinning the verbatim message (`my message`, no quotes). **Revert-verified.**

## Gates

Resolves finding #16 from the recent code review.


mimi1vx (approved):

PR #286, opened 2026-07-10
fix(connection): close the channel when a command times out

plusky:
## What

A command that hit the inactivity timeout **leaked its SSH channel and orphaned the remote command**. `run()` raises `CommandTimeoutError` on two paths — the non-interactive abort (mtui-mcp) and an interactive "don't wait" answer — and neither called `close_session(session)`. The Channel stays registered on the paramiko Transport (kept in its internal channel map, so never reclaimed by GC) and the remote command keeps running. `Target.run()` catches the error and keeps using the same Connection, so over a session with several timeouts (e.g. hung zypper transactions under mtui-mcp) orphaned channels and still-running remote processes accumulated.

## Fix

Wrap the read loop in a close-and-reraise guard — one close point covering both raise sites. The loop body itself is unchanged (re-indented only); the normal EOF path still closes as before.

The guard catches `BaseException`, not just `CommandTimeoutError` (adversarial-review catch): the wrapped region includes the interactive timeout prompt, so **Ctrl-D** (`EOFError`) or **Ctrl-C** (`KeyboardInterrupt`) at the "wait? (Y/n)" prompt abandons the command exactly the same way and must release the channel too.

## Tests

- `test_run_command_timeout_noninteractive_aborts` — extended: the abandoned channel must be closed. **Revert-verified.**
- `test_run_command_timeout_declined_wait_closes_channel` — same assertion for the interactive decline path. **Revert-verified.**
- `test_run_command_prompt_eof_closes_channel` — Ctrl-D at the wait prompt propagates `EOFError` with the channel closed. **Revert-verified** against a `CommandTimeoutError`-only guard.

## Gates

Resolves finding #19 from the recent code review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -302,6 +302,13 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   message in the SVN log. The message was manually wrapped in `"` for a
   shell that is never involved — svn receives the argv list directly — so
   `commit -m fix the thing` recorded `"fix the thing"` verbatim.
+- A command that hits the inactivity timeout no longer leaks its SSH channel.
+  Both timeout paths (the non-interactive abort used under `mtui-mcp` and an
+  interactive "don't wait" answer) raised without closing the channel, so it
+  stayed registered on the paramiko transport and the remote command kept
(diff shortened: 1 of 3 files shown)


mimi1vx (approved):

PR #287, opened 2026-07-10
fix(hosts): tolerate unreadable addon product files on connect

plusky:
## What

A single stale addon product file made a reference host **impossible to add**. The base-product read and the os-release fallback both deliberately catch `OSError` so a broken product file cannot kill the connect — but the addon loop had no such guard: `sftp.open()` on a dangling-symlink (or otherwise unreadable) `/etc/products.d/*.prod` entry raised `OSError` out of `parse_system` into `Target.connect()`, aborting the whole connect. `sftp.listdir` still returns the entry (it matches the `.prod` filter), so this is exactly the dangling-symlink case the base-product path already handles.

## Fix

Wrap the per-addon open/parse in `try/except OSError` that logs a warning naming the offending file and skips just that addon, mirroring the base-product handling.

## Tests

- `test_unreadable_addon_prod_is_skipped_not_fatal` — a host with a good base product, one dangling addon, and one readable addon connects fine: base parsed, readable addon kept, broken addon skipped with a warning. **Revert-verified**: the old loop propagates the `OSError`.

## Gates

Resolves finding #20 from the recent code review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -309,6 +309,11 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   running — repeated timeouts accumulated orphaned channels and remote
   processes. The channel is now closed before the timeout error propagates,
   and likewise when the wait-prompt is abandoned with Ctrl-D or Ctrl-C.
+- Connecting a reference host no longer fails outright when one of its addon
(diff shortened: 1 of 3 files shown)


mimi1vx (approved):

PR #288, opened 2026-07-10
fix(support): match single-quoted os-release values in detect_system

plusky:
## What

On hosts whose `/etc/os-release` single-quotes its values (`NAME='openSUSE'` — permitted by the os-release spec and used by some derivatives/dev machines), the testreport export/commit footer rendered garbled — `## export MTUI:… on -  (kernel: …)` — with no distro or version. `detect_system()` compiled `NAME=["|](.*)["|]`: inside a character class `|` is a literal pipe, not alternation, so the classes matched a double quote **or a pipe**. SUSE's double-quoted format happened to work; single-quoted values never matched, leaving `distro`/`verid` empty.

## Fix

Real quote classes: `NAME=["'](.*)["']` and likewise `VERSION_ID`. No other behaviour change.

## Tests

- `test_detect_system_single_quoted_os_release` — a single-quoted os-release parses to the same fields the double-quoted one does. **Revert-verified**: on the old classes both fields come back empty.

## Gates

Resolves finding #28 from the recent code review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -314,6 +314,10 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   unreadable. The base-product path already tolerated exactly this; the addon
   loop did not, so a single stale `.prod` entry made the whole host
   impossible to add. The offending addon is now skipped with a warning.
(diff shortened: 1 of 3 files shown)


mimi1vx (approved):

PR #291, opened 2026-07-10
fix(cli): return (width, height) from the termsize ACCTEST fallback

plusky:
## What

Under the acceptance-test harness (`ACCTEST_ROWS`/`ACCTEST_COLS`, no controlling TTY) `termsize()` returned the **transposed geometry**: the ioctl path unpacks TIOCGWINSZ as (rows, cols) and returns `(width, height)`, but the fallback returned `int(env["ACCTEST_ROWS"]), int(env["ACCTEST_COLS"])` — (rows, cols). Every caller unpacks `width, height = termsize()`, so `page()` wrapped lines at the row count and printed a column-count of lines per page, and the interactive SSH PTY got swapped dimensions.

## Fix

The fallback returns `(cols, rows)`, matching the normal path's ordering.

## Tests

- `test_termsize_fallback_matches_normal_path_order` — ioctl forced to raise, `ACCTEST_ROWS=24 ACCTEST_COLS=80` ⇒ `(80, 24)`. **Revert-verified.**

## Gates

Resolves finding #15 from the recent code review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -213,6 +213,11 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
 
 ### Fixed
 
+- Under the acceptance-test harness (`ACCTEST_ROWS`/`ACCTEST_COLS`, no
+  controlling TTY) the terminal geometry is no longer transposed: the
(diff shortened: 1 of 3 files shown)


mimi1vx (approved):

PR #18, opened 2023-10-18
Add fixes and noqa from ruff

mimi1vx:

PR #292, opened 2026-07-10
fix(commands): preserve the directory tree in put

plusky:
## What

`put mydir/` **flattened the tree and silently overwrote same-named files**. The walk collected every nested file but uploaded each to `target_wd(filename.name)` — basename only — so `mydir/sub1/test.sh` and `mydir/sub2/test.sh` both resolved to the same remote path; only the last survived while the per-file success log claimed both were transferred.

## Fix

Each walked file keeps its path relative to the uploaded directory's parent (so the directory name itself is included) and the remote path is built from those parts; `Connection.sftp_put` already creates the intermediate remote directories. Plain-file arguments still upload under their basename.

The walked directory is **resolved first** (adversarial-review catch): a `..`-shaped argument (`put ..`) would otherwise survive `relative_to()` as a literal `..` part and build a remote path escaping the run's working directory into the shared `target_tempdir`.

## Tests

- `test_sftp_put_directory_preserves_tree` — `mydir/sub1/test.sh` + `mydir/sub2/test.sh` land on two distinct structured remote paths. **Revert-verified**: the old code produces one clobbered path.
- `test_sftp_put_dotdot_stays_inside_working_directory` — `put ..` stays confined (no `..` parts in any remote path). **Revert-verified** against the unresolved walk.

## Gates

Resolves finding #18 from the recent code review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -218,6 +218,11 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   fallback returned (rows, cols) while the normal path returns
   (width, height), so the pager wrapped at the row count and printed a
   column-count of lines per page.
+- `put` of a directory now preserves the directory tree on the remote
+  hosts. Every nested file used to be uploaded flat under its basename, so
+  two files with the same name in different subdirectories silently
+  overwrote each other (only the last survived, while the per-file success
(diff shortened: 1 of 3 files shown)


mimi1vx (dismissed):

mimi1vx (approved):

PR #293, opened 2026-07-10
fix(connection): re-raise sftp_remove errors so the rmdir fallback works

plusky:
## What

Removing a remote **directory** silently no-oped, and a **failed file removal was reported as success**. `Target.sftp_remove` distinguishes ENOENT from other failures and falls back to `sftp_rmdir` for directories — but only if `Connection.sftp_remove` raises. The connection layer caught `OSError` internally, logged it, and returned normally, so the whole target-layer except-branch (ENOENT debug, directory fallback, "unable to remove" warning) was **dead code**. `TargetLock`'s lockfile removal carries the same never-exercised ENOENT tolerance.

## Fix

`Connection.sftp_remove` logs at debug (the callers own the reporting) and **re-raises**. Both existing callers already carry the `except OSError` handling that was written for exactly this.

## Tests

- `test_sftp_remove_reraises_oserror` — updated from pinning the swallow to pinning the propagation. **Revert-verified.**
- The existing target-layer tests (`ENOENT`, rmdir fallback, rmdir-failure warning) now cover live code paths.

## Gates

Resolves finding #22 from the recent code review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -223,6 +223,11 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   two files with the same name in different subdirectories silently
   overwrote each other (only the last survived, while the per-file success
   log suggested both were transferred).
+- Removing a remote directory (e.g. via `FileDelete`/host cleanup) works
+  again, and a failed file removal is no longer reported as success. The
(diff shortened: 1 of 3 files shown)


mimi1vx (dismissed):

mimi1vx (approved):

PR #294, opened 2026-07-10
fix(mcp): stop schema slimming eating properties named 'title'

plusky:
## What

`slim_tool_schema` (the `mtui-mcp` token-slimming pass) **deleted any tool parameter literally named `title`**. It recursed the whole JSON schema skipping every dict key equal to `title` (and rewriting `description` values) — but on a `properties` map the keys are parameter *names*, not schema keywords. A parameter named `title` vanished from `properties` while staying in `required`: the tool advertised a required property with no schema, impossible to call correctly. Latent today (no current tool has such a parameter), but `title` is an ordinary flag name.

## Fix

The recursion tracks whether the current dict's keys are names (children of `properties`/`patternProperties`/`$defs`/`definitions`) and suspends the keyword transforms — title-drop, description-rewrite, nullable-flatten — for exactly that level; each property's own schema is slimmed normally. The transform is now actually meaning-preserving, as documented.

## Tests

- `test_slim_keeps_property_named_title` — properties named `title`/`description` survive (their *keyword* `title` still stripped), `required` intact. **Revert-verified.**
- `test_slim_keeps_nested_defs_property_names` — nested object properties and `$defs` entries named `title` survive. **Revert-verified.**

## Gates

Resolves finding #23 from the recent code review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -228,6 +228,12 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   connection layer swallowed the SFTP error internally, so the target
   layer's ENOENT tolerance and its directory `rmdir` fallback were dead
   code; the error now propagates to the layer that handles it.
+- The `mtui-mcp` tool-schema slimming no longer deletes a tool parameter
+  (or nested object property) literally named `title` or `description`.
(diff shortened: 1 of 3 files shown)


mimi1vx (dismissed):

mimi1vx (approved):

PR #295, opened 2026-07-10
fix(reports): merge connected systems across rounds; close leaked targets

plusky:
## What

Two defects in the refhost connect path, bundled (same function pair in `testreport.py`):

**#29** — `connect_targets` rebuilt `self.systems` from only the hosts connected in the *current* invocation (already-connected hosts are excluded from the work set), while `self.targets` accumulates. `autoconnect` calls `connect_targets` twice, so the second call wiped the first batch from `self.systems`: both hosts stayed connected, but `list_metadata`'s "Hosts" line showed only the latest batch.

**#30** — `connect_target` **leaked the open SSH session** when `try_claim` or the autolock `target.lock()` raised *after* `connect()` succeeded: the broad except returned `(False, False)` without closing, and the target was never stored in `self.targets` (paramiko transports are not reliably closed by GC).

## Fix

- Merge `self.systems` with `update()` instead of rebuilding, and prune entries together with the inactive-target sweep. On Ctrl-C the partially collected `new_systems` batch is discarded too (adversarial-review catch: phantom hosts whose Target objects were just dropped could otherwise merge into `self.systems`).
- `connect_target` initialises `target = None` and both abort paths (`except Exception` and `KeyboardInterrupt`) close a half-set-up target best-effort (`_close_failed_target`). The sibling `add_target` path (`add_host -t`) had the identical leak (adversarial-review catch) and closes now as well.

## Tests

- `test_connect_targets_second_batch_keeps_first_batch_systems` — two rounds; the first round's host must not vanish from `systems`. **Revert-verified.**
- `test_connect_target_closes_target_when_lock_raises` — lock failure after connect ⇒ `(False, False)` *and* the session closed. **Revert-verified.**
- `test_connect_target_closes_target_on_keyboard_interrupt`, `test_add_target_closes_target_when_setup_fails`, `test_connect_targets_prunes_systems_of_inactive_targets` — review-amendment coverage, each revert-verified against a partial revert.

## Gates

Resolves findings #29 and #30 from the recent code review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -234,6 +234,16 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   keyword — including the keys of a `properties` map, which are parameter
   *names* — so such a parameter vanished from the schema while staying
   listed in `required`, making the tool impossible to call correctly.
+- Connecting reference hosts in several rounds (e.g. template autoconnect,
+  which resolves recorded hosts and testplatform hosts separately) no longer
+  drops the earlier rounds from the "Hosts" metadata line: the connected-
+  systems map is now merged across rounds instead of rebuilt from the latest
+  one, and pruned together with inactive hosts.
+- A host-setup failure right after a successful SSH connect (e.g. the
+  connection dropping while the remote lock file is written) no longer leaks
+  the open SSH session; the half-set-up connection is closed before the host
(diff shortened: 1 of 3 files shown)


mimi1vx (dismissed):

mimi1vx (approved):

PR #298, opened 2026-07-11
test: make the suite mutation-ready (hermetic under in-process reruns)

plusky:
## What
Groundwork for mutation testing (mutmut 3.6), modeled on openSUSE/repose:

- **`[tool.mutmut]`** in pyproject.toml: `source_paths` covers the whole package (the `mutants/` copy must `import mtui`); `do_not_mutate_patterns` drops logger-call mutants so survivors reflect real logic gaps. Inert for the normal toolchain; `mutants/` gitignored.
- **Hermeticity under in-process reruns** — mutmut re-enters `pytest.main()` repeatedly in one interpreter, which exposed tests leaking process-global state:
  - `test_products_sle`: parametrize tables are built once per import and the normalize functions mutate their argument in place → deepcopy per call;
  - `test_main`/`test_mcp_main`: `main()` wires real handlers + levels onto the process-global `mtui`/`mtui-mcp` loggers → restored after each test;
  - `test_mcp_session`: test-body `Command` subclasses register globally via `__init_subclass__` and collide with themselves on rerun → unbound afterwards;
  - `test_fileops`: class-level callback state cleared;
  - conftest: pins mutmut's relative `source_paths` to absolute during stats collection (it re-resolves against the live CWD on every trampoline hit, so `chdir()`ing tests crashed the run). Env-guarded; inert in CI.

With this plus the ColorFormatter PR, the suite passes three consecutive in-process pytest runs — the property mutation testing needs. The first full run: 15,293 mutants, 63.2% kill rate; the surfaced bugs are fixed in the sibling PRs of this series.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -175,3 +175,6 @@ tags
 
 # Claude Code session artefacts
 .claude/
+
+# mutmut mutation-testing working tree
+mutants/
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -65,6 +65,18 @@ markers = [
   "network: tests that perform real network I/O",
 ]
 
+[tool.mutmut]
+# Mutation-testing config. Run with:
+#   uv run --with 'mutmut>=3.6' mutmut run
+# ``source_paths`` must be the whole package: mutmut copies it into a
+# ``mutants/`` tree that has to ``import mtui``, and a single-file scope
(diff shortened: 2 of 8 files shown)


mimi1vx (approved):
Approve: additive hermetic-test hardening; CI green (lint, typecheck, docs, tests 3.13/3.14, codecov).

PR #299, opened 2026-07-11
fix(export): match indented state headers within the host's own block

plusky:
## What
Two defects in `ManualExport._fillup_hosts_to_template`, found by mutation testing:

1. **Dormant lookup** — the primary state-header lookup read `self.template.index("      {state}:\n", index)`: the f-prefix was lost, so the literal `{state}` could never match. On templates with six-space-indented `      before:` / `      after:` headers the version lines were silently omitted and the verdict never filled.
2. **Silent swallow** — a bare `except Exception: pass` around the version-line write hid malformed-template errors, producing reports without the very data reviewers verify.

## Fix
The f-prefix is restored and the per-host state-header search is **bounded to the host's own block** (nearest match of both header forms before the next `(reference host:` line) — the adversarial review proved an unbounded search would write host A's versions into host B's block whenever a newly created unindented section precedes an indented one. The bare except is narrowed to `IndexError` with a warning naming state, package and host.

## Tests
Indented-header placement, mixed indented/unindented blocks in **both orders**, and malformed-template warning — each revert-verified (behavioral failure on reverted code).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -231,6 +231,17 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   prepare and downgrade checks now label the captured command output
   correctly as "stdout:" (it was mislabeled "stdin:"), and an "errocode"
   typo in the downgrade check's exit-106 warning is fixed.
+- MANUAL `export` now writes the per-host package-version lines into templates
+  that use the indented `      before:` / `      after:` state headers, and
+  bounds that lookup to the current host's own block so it can no longer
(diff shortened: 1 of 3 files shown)


mimi1vx (dismissed):
Approve: per-host state-header search bounded to host's own block (prevents cross-host version bleed); except narrowed to IndexError. CI green.

mimi1vx (approved):
Approve: bounds indented state-header lookup to host's own block (prevents cross-host bleed); malformed-template warning. Rebased on main, export tests pass.

PR #300, opened 2026-07-11
fix(checks): correct zypper exit-104 reason and stdout log labels

plusky:
## What
Three defects in `update_workflow/checks/`, found by mutation testing:

1. `checks/update.py` mapped zypper exit **104** to `UpdateError("update stack locked")` — 104 is `ZYPPER_EXIT_INF_CAP_NOT_FOUND`; testers were shown a wrong failure reason (the lock code is 7). Now raises `"package not found"`, matching `checks/install.py`.
2. All four checks modules labeled the command's **stdout as `stdin:`** in the shared `logger.critical` failure log — misleading during failure diagnosis. Label fixed at all 13 sites.
3. `checks/downgrade.py` warned about `"errocode 106"` — typo fixed (test updated).

## Tests
Exit-104 classification asserted via `ei.value.reason`/`ei.value.host`; the stdout-label fix is pinned **per failure branch across all four modules** (the review showed a single-site pin let partial reverts pass). All revert-verified.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -222,6 +222,15 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   own globals rather than `inspect.getmodule`, which returned "unknown"
   under import hooks that report a different file path than the code object
   records.
+- A zypper exit code 104 during `update` is now reported as "package not
+  found" instead of the misleading "update stack locked". Zypper's 104 is
+  `ZYPPER_EXIT_INF_CAP_NOT_FOUND` (the requested package/capability does not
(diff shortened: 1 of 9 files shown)


mimi1vx (dismissed):
Approve: exit-104 now 'package not found' (matches install.py; 104 is CAP_NOT_FOUND, not lock), stdin:->stdout: log label fix, errocode typo. CI green.

mimi1vx (approved):
Approve: exit-104 -> package-not-found (matches install.py), stdin:->stdout: log labels, errocode typo. CI green.

PR #301, opened 2026-07-11
fix(hosts): render lock timestamps in actual UTC

plusky:
## What
`TargetLock.time()` converted the lock's epoch timestamp with `datetime.fromtimestamp()` — the **local** timezone — but formatted it with a literal `UTC` suffix. Any tester outside UTC saw lock times wrong by their UTC offset in `list_locks` output and the `update` lock warning.

## Fix
Timezone-aware conversion (`datetime.fromtimestamp(ts, tz=timezone.utc)`); output format otherwise unchanged.

## Tests
Exact expected string for a fixed epoch, plus a TZ-forced case (`TZ=America/New_York` + `time.tzset`) that fails under local-time conversion regardless of the host timezone. Revert-verified.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -242,6 +242,11 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   written because the template is malformed (a state header as the very last
   line), the export now logs a warning naming the host and package instead of
   silently skipping the line.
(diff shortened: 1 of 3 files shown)


mimi1vx (approved):
Approve: lock timestamps now tz-aware UTC (fromtimestamp tz=timezone.utc); format unchanged. CI green.

PR #302, opened 2026-07-11
fix(cli): complete partial tilde paths and stop hiding siblings

plusky:
## What
Tilde completion was broken for partial paths: `complete_choices_filelist` expanded `~` and then unconditionally appended `/`, so `~/Doc<TAB>` listed the non-existent directory `/home/user/Doc/` and offered nothing. A dead filter (`filename` always `''`) masked the intended per-basename filtering.

## Fix
Tilde expansion no longer forces a slash; the directory to list and the basename prefix are derived properly. From the adversarial review: a tilde path naming an existing directory (or bare `~`) now **descends into it when unambiguous**, directory candidates carry a trailing `/`, and `complete_choices()` no longer short-circuits on an exact text match (which dropped same-prefix siblings depending on set order — this also benefits the other completers using it).

## Tests
Partial prefix, bare `~`, exact-directory-without-slash, dir/dir and file/dir sibling cases — each revert-verified against its specific production hunk; no `chdir` anywhere.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -247,6 +247,17 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   lock timestamp was converted to the tester's local timezone before being
   formatted with the "UTC" suffix, so anyone outside UTC saw a time wrong by
   their UTC offset.
+- Tab completion of file paths now works for partial paths under `~`: typing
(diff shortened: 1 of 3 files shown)


mimi1vx (approved):
Approve: fixes tilde partial-path completion (no forced slash, proper basename prefix, descends unambiguous dirs, keeps same-prefix siblings). CI green.

PR #303, opened 2026-07-11
fix(commands): surface host-teardown failures on quit

plusky:
## What
`quit` submitted per-host disconnects to a thread pool and only called `concurrent.futures.wait(futures, timeout=45)` — exceptions and timed-out closes were never inspected, so mtui exited 0 with every teardown failure invisible (mutation testing proved a close raising `KeyError` on every quit passes the whole suite).

## Fix
After the wait, every failed close is logged with its hostname; stragglers get a *"still disconnecting after N seconds"* notice and — since the executor's context exit joins them anyway (pre-existing behavior, unchanged) — their late failures are also reconciled and logged after the join, so no exception is ever silently dropped. Wording deliberately avoids promising a non-blocking exit.

## Tests
Failing close (warning names the right host), multi-host attribution, straggler-exception reconciliation, and a clean quit emitting no warnings — timeout constant patched for speed; all revert-verified.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -258,6 +258,12 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   already typed, e.g. `$HOME//Documents`. Directory candidates (including
   bare `~`'s home-directory entries) are now suffixed with `/`, matching
   shell convention, so a following `TAB` descends into them.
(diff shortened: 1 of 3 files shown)


mimi1vx (approved):
Approve: quit now surfaces per-host teardown failures with attribution; bare except narrowed. CI green.

PR #304, opened 2026-07-11
fix(products): normalize compound SLE15 LTSS-ERICSSON versions

plusky:
## What
`normalize_sle15` checked its generic `LTSS` branch before `ERICSSON`, so a compound `...-LTSS-ERICSSON` version normalized to `SLES-LTSS` with `-ERICSSON` dangling in the version string — an unmatchable product.

## Fix
A dedicated `LTSS-ERICSSON` compound branch (mirroring `normalize_sle12` and the existing `LTSS-TERADATA` handling) produces `SLES-LTSS-ERICSSON` and strips the whole suffix, so `15-SP4-LTSS-ERICSSON` → (`SLES-LTSS-ERICSSON`, `15-SP4`); the specific-before-generic reorder stays. The adversarial review caught that the original reorder-only version of this fix still left `-LTSS` dangling — the compound branch and exact name+version assertions came out of that review.

## Tests
Compound, plain-LTSS and plain-ERICSSON cases pinning **both** the normalized name and the cleaned version string. Revert-verified.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -264,6 +264,16 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   the affected host. The exit status is still unaffected, but quitting
   already blocked until every disconnect finished before this fix and
   still does — only the visibility of the outcome changed, not the wait.
+- SLE 15 LTSS-ERICSSON products are now normalized to a dedicated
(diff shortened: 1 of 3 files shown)


mimi1vx (approved):
Approve: dedicated LTSS-ERICSSON compound branch (mirrors TERADATA) -> SLES-LTSS-ERICSSON with full suffix strip; specific-before-generic order kept. CI green.

PR #305, opened 2026-07-11
fix(support): parse unquoted os-release values

plusky:
## What
`detect_system` parsed `/etc/os-release` with regexes that require quoted values. Spec-legal unquoted values (`NAME=Fedora`, `VERSION_ID=15.6`) silently never matched, so distro/version reported empty in the export footer and commit line.

## Fix
The patterns accept quoted or unquoted values. The quoted alternative keeps the old greedy semantics **byte-identical** (including values with embedded/escaped quotes — the review caught that a naive `[^"]*` alternation regressed those); the unquoted alternative captures up to whitespace.

## Tests
Unquoted, quoted (unchanged behavior), mixed, and escaped-quote fixtures. Revert-verified.


mimi1vx (approved):
Approve: detect_system now parses spec-legal unquoted os-release values (NAME=Fedora, VERSION_ID=15.6); quoted behavior unchanged. CI green.

PR #307, opened 2026-07-12
test(export): pin export idempotency and template-mutation behavior

plusky:
## What
50 pinning tests (4 new files) for the export pipeline's biggest survivor clusters (~445 mutants): `inject_openqa` section replacement and re-export bounds, `installlogs_lines` idempotency/convergence, `_writer` overwrite/prompt/sidecar contract, `_fillup_hosts_to_template` section creation and version overwrites, AUTO/KERNEL `install_results`/`kernel_results` exact layouts, and downloader URL/path construction.

Notable: the re-export golden tests pin exact full-template equality, killing the `overview_inject` mutant class that could silently delete the `build log review:` header. Three suspected production quirks are documented in-test as pinned current behavior (AutoExport.run's `install_logs_current` guard never matches a real results line; `add_sysinfo`'s dead guard; `inject_openqa`'s one-blank-line drift per re-export) — follow-up material, deliberately not changed here.

### DIFF
--- a/tests/test_export_base_pins.py
+++ b/tests/test_export_base_pins.py
@@ -1,0 +1,449 @@
+"""Mutation-killing pins for ``mtui.update_workflow.export.base``.
+
+A full mutmut run left survivors in code the export suite executes but
+never asserts on. These tests pin the exact template layout produced by
+``BaseExport.inject_openqa``, ``installlogs_lines``, ``_writer``,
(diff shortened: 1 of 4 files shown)


mimi1vx (approved):
Approve: pins export idempotency and template-mutation behavior (mutation-hardening, tests-only). CI green.

PR #308, opened 2026-07-12
test(checks): pin zypper error reasons, hosts, and print slices

plusky:
## What
38 pinning tests for all four `update_workflow/checks/*` modules (~149 survivors): every raising branch now asserts `UpdateError.reason` **and** `.host` (previously most tests only checked `pytest.raises(UpdateError)`, so reason/host mutants all survived), success-exitcode precedence over scary output, the 106 warning gating, and the exact RPM-output/vendor-block print slices with and without colors.

**Stacked on #300** (which corrects the exit-104 reason this branch pins) — merge #300 first; I'll rebase this to a clean diff afterwards.


mimi1vx (approved):
Approve: pins zypper error reasons, hosts, and print slices (tests-only). CI green.

PR #309, opened 2026-07-12
test(products): pin normalize branch behavior for sle11/12/15

plusky:
## What
Branch-by-branch pinning tests for the products normalize trio (~121 survivors): every rewrite branch's exact output tuple (name AND version) is asserted, including the compound LTSS flavors, so branch-order and replace-string mutants die instead of surviving behind first-element-only assertions.


mimi1vx (approved):
Approve: pins products normalize branch behavior for sle11/12/15 (tests-only). CI green.

PR #310, opened 2026-07-12
test(openqa): pin dashboard/search/updateid data-path behavior

plusky:
## What
51 pinning tests (3 new files) for the openQA data path (~409 survivors): `DashboardAutoOpenQA._normalize_job` field mapping and `_get_logs_url` selection, `_pretty_print_section` layout, `oqa_search` incident/aggregate version scans and result classification, and `UpdateID._checkout`'s prompt flow — including the real `prompt_user` accept lists (both `"yes"` and `"y"` literals, via the low-level reader rather than mocking `prompt_user` away).


mimi1vx (approved):
Approve: pins dashboard/search/updateid data-path behavior (tests-only). CI green.

PR #311, opened 2026-07-12
test(mcp): pin the background-job lifecycle envelopes

plusky:
## What
22 pinning tests for `McpSession`'s job machinery (~131 survivors): running-state visibility while a job is in flight, failed/cancelled/unknown-id `McpCommandError` envelopes (exit_code, stdout, stderr — not just the message substring), `sys.exit` translation boundaries in `_run_sync`, `_job_view` elapsed rounding, cancellation pinned to the `job_cancel` call itself (not the loop teardown), and `make_testreport` call-contract on load.


mimi1vx (approved):
Approve: pins the background-job lifecycle envelopes (tests-only). CI green.

PR #312, opened 2026-07-12
test(mcp): pin testreport tool contracts

plusky:
## What
26 pinning tests for `testreport_*` MCP tools (~132 survivors): the documented error-envelope contract (`stdout == ""`, `exit_code == 1`) on every refusal path, read window/limit edges (including `limit=0`), template scoping with multiple loaded reports, output-cap wiring, heartbeat progress arguments, logs size/path fields, and `_atomic_write_text`'s same-directory temp placement (the point of the atomic pattern).


mimi1vx (approved):
Approve: pins testreport tool contracts (tests-only). CI green.

PR #313, opened 2026-07-12
test(mcp): pin argv synthesis and schema generation

plusky:
## What
36 pinning tests (3 new files) for the MCP request plumbing (~89 survivors): `kwargs_to_argv` round-trips (store_const mutex groups incl. the pre-scan skip semantics, nargs shapes, negative numbers, remainder), `_schema.action_to_parameter` typing/required rules, and tool wiring in `_make_wrapper`/`_fan_out_subparser`.

The adversarial review caught a golden-value coincidence in the const pre-scan test (the mutant emitted the same flag the assertion expected); the case now uses a const that differs from the group's first member, so the `continue`→`break` mutant fails loudly.


mimi1vx (approved):
Approve: pins argv synthesis and schema generation (tests-only). CI green.

PR #322, opened 2026-07-13
feat(obs)!: replace the osc qam plugin with a native direct-OBS-API backend

plusky:
## What & why

The QAM review actions (**approve / assign / unassign / reject / comment**) previously shelled out to the external **`osc qam`** CLI plugin via `mtui/data_sources/oscqam.py`. This replaces that with a **native, in-process backend that calls the OBS/IBS API directly** over `requests` (mirroring the Gitea client), with **native SSH-signature auth reproduced with paramiko**. Credentials come from the user's existing **`~/.oscrc`** — mtui parses it and authenticates itself. No `osc` library, no `osc qam` subprocess; the `osc-plugin-qam` dependency is dropped.

The drop-in `OSC(config, rrid)` seam and its 5 bool / never-raising methods are unchanged, so `commands/apicall.py` and `commands/approve.py` are untouched.

## How it works

- **Auth** (`obs/sshsig.py`, `obs/auth.py`): on a `401`, read the `WWW-Authenticate: Signature` realm from urllib3's raw multi-value header list, build an SSHSIG over `(created): <epoch>` with paramiko (Ed25519), and resend once with `Authorization: Signature …`. The SSHSIG output is **verified byte-for-byte against `ssh-keygen -Y sign`** (golden-vector test). The Authorization header/signature are never logged.
- **Credentials** (`obs/oscrc.py`): stdlib `configparser` over `~/.oscrc`; resolves the `https://api.suse.de` section (honouring `[general]` `sshkey` inheritance); **never reads `pass`/`passx`** for this Signature-only target; fail-closed `ObsConfigError` (never prompts).
- **Transport** (`obs/client.py`): one `requests.Session` on `support/http` — per-call `HTTP_TIMEOUT` + a coarse between-call `[obs] request_timeout` budget, TLS-hint reuse, error-envelope parsing.
- **Models + inference** (`obs/models.py`, `obs/inference.py`): OBS XML parsers and the assignment state machine **ported verbatim from `openSUSE/osc-plugin-qam`** (GPL-2.0) — add-on-accepted / remove-on-reopened / ignore-assigned, then drop finished reviewers. Backs both the unassign guard and the approve role-check.
- **Operations** (`obs/qam.py`): the 5 ops with faithful semantics — single-group auto-inference, open-state, the qam.suse.de template/SUMMARY preconditions (skipped for PI/SLFO), previous-reject with the plugin's early-returns, refused group-approve, `by_user` reject with the `MAINT:RejectReason` read-modify-write, `[oscqam] ` prefix on approve/reject only.
- **Facade** (`oscqam.py`): `OSC(config, rrid)` builds everything lazily inside one try/except so the methods keep their bool / never-raising contract.

## Behaviour changes (documented in CHANGELOG)

- **No `osc qam` fallback** — the backend is native-only (the transitional `[obs] backend` selector was dropped per review).
- OBS **TLS** is governed by `[mtui] ssl_verify`, not oscrc's own knobs (`sslcertck`/`cafile`/`capath`/`no_verify`).
- `[obs] request_timeout` (default 180s) is a **coarse between-call budget**, not a hard mid-call kill.
- Day-one key scope: an **Ed25519 private key on disk**. ssh-agent, encrypted, and RSA/ECDSA keys are not yet supported (deferred, fail-closed).

## Design & review

Full design + adversarial-review writeup: the plan artifact (direct-API redesign; 3 research → design → 5 adversarial lenses → revise; 21 findings, 1 blocker on the inference state machine — all folded in). This PR supersedes the earlier osc-library approach.

## Testing

Entirely offline via `responses` + golden vectors — no real OBS. New `obs/` package coverage ~99%; the SSHSIG golden vector is cross-checked against `ssh-keygen`. Gates green: `ruff format --check .` · `ruff check .` · `ty check` · `pytest` (2367 passed).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


okurz:
With this change in place: should we archive osc-plugin-qam in https://github.com/openSUSE/osc-plugin-qam ?

PR #331, opened 2026-07-17
perf(export): download AUTO-workflow install logs concurrently

plusky:
`AutoExport.get_logs` pulled each openQA install log through a lazy `map` inside
`zip_longest`, so the N downloads ran strictly one after another (each also
building its own `requests` session). This fans them out with `ContextExecutor`:
`executor.map` preserves input order, so results still line up with
`auto.results`, and the file-writing loop stays serial (it touches the
filesystem and may prompt on overwrite).

- N is typically 2–6 install jobs; ~0.5–2s serialized becomes ~the single
  slowest log. One-time per report export.
- **No session is shared across workers** — each `_openqa_installog_to_template`
  builds its own, as before; failures are still caught per-log and yield an
  empty result (that log is skipped).

### Testing
- Order preservation (each log's content lands in its own, correctly-named
  file).
- A **Barrier-based** concurrency proof (verified to fail with
  `BrokenBarrierError` on a serial revert).
- A failed-download test (one log returns empty → skipped, pairing of the
  remaining two survives the concurrent fan-out).
- `ruff` / `ty` / full `pytest` (2421) green locally; 100% patch coverage on the
  changed lines.

Adversarial pre-PR review (3 dimensions × 3 lenses): **no must-fixes / no
should-fixes** — folded in the one useful nit (the failed-download regression
guard).

Completes the performance series: #327 (mtui-mcp startup), #328
(openqa_overview), #329 (refhosts memoization), #330 (reboot reconnect).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,9 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   The wait itself is kept (it guards against reconnecting to a
   still-shutting-down sshd); only the serialization is removed. Most visible on
   transactional (SL Micro) updates, which reboot every host.
+- Exporting an AUTO-workflow testreport is faster. The openQA install logs are
+  now downloaded concurrently instead of one after another; results are written
+  in the original order, so the exported report is unchanged.
 
 ## 19.0.1 - 2026-07-17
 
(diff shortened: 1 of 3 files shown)


plusky:
For the record: superseded by #337 — the Rust tree carries this change (concurrent, order-preserving export log fan-out in crates/mtui-testreport/src/export/auto.rs, which cites this PR). Nothing to forward-port.

PR #332, opened 2026-07-18
perf(refhosts): cache the refhosts.yml parse process-wide

plusky:
Second layer under #329. `refhosts_from_tp` / `list_refhosts` / `add_host` each
build a `Refhosts` store via the resolver, and each build is a **~1s GIL-held
ruamel parse** of the ~68KB `refhosts.yml`. #329 collapsed that to once per
report; under the `mtui-mcp` **http transport**, hundreds of concurrent sessions
still each re-parse the same file, serialising the interpreter on one core (the
scale audit's finding **A2**).

This adds a **process-wide single-flight memo** at the `Refhosts(path)`
construction boundary (`store.load_refhosts`), injected as the default resolver
factory. The parsed store is read-only, so one instance is safely shared across
all callers. The key is `(resolved path, st_mtime_ns, st_size)`, so an edited
file — or the periodic HTTPS cache refresh — is picked up on its next load. A
single lock held **across** the parse (not a bare `lru_cache`) makes concurrent
cold-cache callers wait for one parse instead of stampeding into N simultaneous
~1s parses; **cache hits are lock-free**. An un-statable path bypasses the cache
so the real error surfaces unchanged.

Composes with #329: the per-report memo sits above this process-wide one.

### Testing
- Memoization (same instance), mtime/size invalidation **and** an equal-size
  edit invalidated via `st_mtime_ns`, **single-flight under 6 concurrent
  threads** (verified to fail — 6 parses — on a no-lock revert), FIFO eviction
  at the size bound, and uncacheable-path passthrough.
- A conftest autouse fixture isolates the process-wide cache between tests.
- `ruff` / `ty` / full `pytest` (2431) green locally.

Adversarial pre-PR review (4 dims × 3 lenses): **no must-fixes**; folded in the
should-fix (FIFO eviction coverage) and nits (single-flight revert-sensitivity,
a same-size invalidation test, a docstring note that the memo relies on
filesystem `st_mtime_ns` resolution).

First of the scale-audit top-three fixes; A1 (command executor) and A3 (Slack
watch) follow on their own branches.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,12 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   updates. `refhosts.yml` is now parsed once per report and shared across every
   testplatform (and with the product-drift check) instead of being re-parsed
   (~1s each) for every testplatform during autoconnect / `add_host`.
+- `refhosts.yml` is now parsed at most once per file version for the whole
+  process, shared across all reports and sessions, instead of once per report.
+  This matters most under the `mtui-mcp` server, where many concurrent sessions
+  previously each spent ~1s of CPU re-parsing the same file; a single-flight
+  cache (keyed on the file's path, mtime and size) collapses that to one parse
+  and picks up an updated file on its next load.
 - Rebooting a host group is faster. After the reboot is dispatched, the hosts
   are now reconnected concurrently instead of one at a time, so the fixed
   per-host reconnect wait (~10s) overlaps across hosts rather than stacking.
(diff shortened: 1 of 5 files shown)


plusky:
For the record: superseded by #337 — serde_yaml parses refhosts.yml in ~ms, so the parse memo has no Rust counterpart. One residual worth a follow-up issue: the HTTPS refhosts refresh (crates/mtui-datasources/src/refhost/resolvers.rs) has no single-flight, so many concurrent sessions can re-download in a herd at cache expiry.

PR #333, opened 2026-07-18
perf(mcp): explicit, configurable command thread pool

plusky:
Every blocking command body under `mtui-mcp` runs via `asyncio.to_thread`, which
uses the event loop's **implicit** default `ThreadPoolExecutor`. Under the http
transport that pool is completely decoupled from `[mcp] session_cap`: an
operator can raise the cap to admit hundreds of sessions, but command
concurrency stays pinned near 32 and calls queue (scale-audit finding **A1**).

This installs a process-wide `ContextExecutor` as the loop's default executor
from a FastMCP lifespan hook, sized via a new **`[mcp] command_pool_size`**
(default `min(32, process_cpu_count()+4)` — matching asyncio's own
affinity/cgroup-aware default, so stdio and modest http are byte-identical). It
is a `ContextExecutor`, not a bare pool, so per-call log-capture contextvars
still propagate into the worker threads.

**The multi-session subtlety (caught in review):** under http, FastMCP enters
the lifespan **once per session** on one shared loop. So the pool is installed
exactly **once per loop** and **never torn down per session** — a per-session
`shutdown()` would poison the shared default executor for every live peer (their
next `to_thread` would raise *"cannot schedule new futures after shutdown"*).
`concurrent.futures`' atexit join reclaims it at interpreter exit.

Scoped out (deliberately, as a separately-reviewable follow-up): splitting into
separate read-only and slow-op pools so a burst of slow commands can't
head-of-line-block quick reads. This PR is the configurable-pool foundation.

### Testing
- Config read / default / reject, with **host-independent shape pins**
  (`process_cpu_count` monkeypatched to 4 → 8 and 64 → 32).
- A lifespan runtime test (installs a `ContextExecutor` sized from the knob).
- A **multi-session regression test** — a second session's lifespan exit must
  not break a live peer's `to_thread` — verified RED against a per-session
  install/shutdown implementation.
- `ruff` / `ty` / `sphinx -W` / full `pytest` (2428) green locally.

Adversarial pre-PR review (4 dims × 3 lenses): it caught a **must-fix** (the
per-session teardown bug above) and two should-fixes (affinity-aware default;
the regression test) — all folded in; the concept was confirmed sound and
honestly scoped.

Third of the scale-audit top-three fixes (with #332 refhosts memo; A3 Slack
watch follows on the Slack branch).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,16 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   still-shutting-down sshd); only the serialization is removed. Most visible on
   transactional (SL Micro) updates, which reboot every host.
 
+### Added
+
+- New `[mcp] command_pool_size` option sizes the thread pool that runs blocking
+  command bodies under `mtui-mcp` (installed as the asyncio loop's default
+  executor). It defaults to `min(32, cpu+4)` — asyncio's own default — so stdio
+  and modest http deployments are unchanged. An http deployment that raises
+  `[mcp] session_cap` for a large agent fleet should raise this too; otherwise
+  command concurrency stays pinned near 32 regardless of how many sessions
+  connect.
+
 ## 19.0.1 - 2026-07-17
 
 ### Added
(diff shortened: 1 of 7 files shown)


plusky:
For the record: superseded by #337 — SSH is fully async (russh on tokio), so executor sizing no longer applies; the analogous bound is the existing [connection] max_parallel option.

PR #334, opened 2026-07-20
fix(obs): don't reject a usable sshkey over credentials_mgr_class

plusky:
Fixes a reported failure: `approve` (and `reject`) died outright for a common
oscrc shape.

```
error: OBS operation on SUSE:SLFO:1.1:416806 failed: oscrc [https://api.suse.de]
uses credentials_mgr_class='osc.credentials.TransientCredentialsManager'; the
native OBS backend supports only SSH-signature auth (an 'sshkey' entry) —
keyring/transient-password managers are not supported
```

...for an oscrc that **already had a perfectly usable `sshkey`**:

```ini
[https://api.suse.de]
user=simonlm
sshkey=id_rsa
credentials_mgr_class=osc.credentials.TransientCredentialsManager
pass=xxxxxxx
```

`read_credentials` checked `credentials_mgr_class` *before* resolving the key,
so the presence of a password manager vetoed an auth method that would have
worked.

### Why ignoring the manager is right

Verified against the installed `osc` (1.27.2): `osc` disables its password path
**outright** for `TransientCredentialsManager` — `connection.py:648`,
`if creds_mgr == "osc.credentials.TransientCredentialsManager":
self.basic_auth_password = False` — so that exact oscrc authenticates by SSH
signature under `osc` too, and never reads `pass`.

To be precise about the general case: `osc` does *not* universally prefer
signature auth. For other managers it prefers Basic once a password resolves
(`connection.py:793`). mtui has **no** Basic path at all, so a configured key is
its only usable credential either way — signature-or-fail is the only option we
have.

The manager is only consulted when **no** key is configured, where a
keyring/transient manager still **fails closed** — its secret isn't in the file
and mtui never prompts (which would hang headless `mtui-mcp`).

Also: `credentials_mgr_class` is no longer inherited from `[general]`. `osc`
gives it no parent default (`conf.py:278`) unlike `sshkey`'s `FromParent`
(`conf.py:299`), so a global manager must not veto a usable per-host key.

### Secret hygiene

`configparser` folds an indented continuation into the *previous* option's
value, so:

```ini
sshkey = id_rsa
 pass = hunter2
```

yields `sshkey == "id_rsa\npass = hunter2"`. Two error paths interpolated such a
value raw, echoing the password into the log **and** the `mtui-mcp` transcript.
Values now go through a display helper that prints only values matching a
conservative charset and replaces anything else wholesale with `<malformed>` (a
truncation length is a guess, and an empty first line defeats it).

The keyless error also no longer claims "plaintext-password auth is not
supported" — misleading for a keyring or obfuscated manager — and instead names
every unsupported mechanism.

### Scope

`pass`/`passx` are still never read. Password (HTTP Basic) auth for users with
**no** key at all is a separate, larger change (obfuscated-`passx` decoding, a
redaction-safe credential type, TLS-and-same-host guards on the Basic header, a
no-preemptive-send handler); a full plan exists and can follow. This PR is the
fix for the reported breakage.

### Testing
- Regression test using the **exact reported oscrc**: succeeds via the key, and
  the password never reaches the credentials' `repr`.
- A folded-continuation test asserting a password can't reach an error message.
- A discriminating `[general]`-inheritance test (the first attempt wasn't — it
  had a host `sshkey`, so the manager lookup was unreachable and it passed
  either way).
- A usable key beats a keyring manager; with no key, keyring/transient still
  fails closed.
- `ruff` / `ty` (ty-strict tree) / `sphinx -W` / full `pytest` (2429) green.

Went through an adversarial review (4 dimensions × 3 lenses): **no must-fixes**;
it independently verified the `osc` claims against the installed source and
**refuted** my original "osc orders Signature ahead of Basic" wording, which is
corrected above and in the code comment. The secret-echo path and the
non-discriminating test were its findings too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,9 +13,10 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   `credentials_mgr_class` when a usable `sshkey` is also configured. `approve`
   and `reject` failed outright with "supports only SSH-signature auth" for the
   common oscrc that sets both an `sshkey` and a password manager (e.g.
-  `TransientCredentialsManager`), even though the key was perfectly usable —
-  and that is exactly what `osc` itself does, ordering SSH-signature auth ahead
-  of the password path. The key is now resolved first and the credentials
+  `TransientCredentialsManager`), even though the key was perfectly usable.
+  `osc` disables its own password path outright for that manager, so the same
+  oscrc authenticates by SSH signature under `osc` too. The key is now
+  resolved first and the credentials
   manager is only consulted when no usable key exists, where a
   keyring/transient manager still fails closed (its secret is not in the file
   and mtui never prompts). `credentials_mgr_class` is also no longer inherited
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,20 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
 
 ## [Unreleased]
 
+### Fixed
+
+- The native OBS backend no longer refuses an oscrc that names a
+  `credentials_mgr_class` when a usable `sshkey` is also configured. `approve`
+  and `reject` failed outright with "supports only SSH-signature auth" for the
+  common oscrc that sets both an `sshkey` and a password manager (e.g.
+  `TransientCredentialsManager`), even though the key was perfectly usable —
+  and that is exactly what `osc` itself does, ordering SSH-signature auth ahead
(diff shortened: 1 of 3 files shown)


plusky:
For the record: superseded by #337 — the native oscrc reader resolves sshkey before credentials_mgr_class (host-section-only) and carries this PR's reproducer as a regression test (crates/mtui-datasources/tests/obs_oscrc.rs); parse errors are fixed-message and rust-ini does not fold continuation lines, so the secret-echo vector is structurally gone. Keyless/Basic auth remains unimplemented in Rust too — that plan is now a Rust task.

PR #335, opened 2026-07-20
feat(teregen): sync the client with the current report API

plusky:
Syncs mtui's TeReGen client with the current `/api/v1` (teregen master moved 29
commits: live per-report assignment, granular parsed-report endpoints, live
priority/deadline for classic Maintenance, a `GET /updates` queue).

**Nothing was actually broken** — the one server-side shape break (dropping the
flat `assignee` scalar on `/updates`) was reverted in teregen itself
(`5f9b864`) — so this is additive:

### `assign` shows assignment state
After assigning, mtui prints who holds (or has decided) each review group, from
the report's live `assignees` map (`{group: [{user, state}]}`). Deliberate
contract, encoded in tests:
- **context only, never a gate** — an empty map stays silent because it's
  indistinguishable from an upstream lookup failure;
- the server caches the endpoint ~300s, so the output says *"may lag ~5 min"*
  (right after assigning it typically shows the pre-assign snapshot);
- a group may carry both a decision entry and a live assignment (decider ≠
  tester);
- malformed payloads are filtered, never raised — this prints *after* the
  assignment succeeded, and a crash here would dress success up as an error.

### New client calls (shapes verified against the teregen source)
- `parsed(rrid, section=None)` — full parse, or an `{id, section, data}`
  envelope per section;
- `bugs(rrid, bug_id=None)` — flat index rows, single bug as an
  `{id, bug_id, bug}` envelope; `bsc#…` ids percent-encoded (an unencoded `#`
  truncates the URL as a fragment);
- `completeness(rrid)` — against the **canonical** `GET /reports/{id}/completeness`,
  not `/parsed/completeness` (unspecified section behaviour the server may
  close).

### Cleanup
- `priority_deadline()` removed (its only caller now reads `info()` directly);
- `updates` listing survives a non-string deadline;
- stale docstring + `--review-group` help corrected; `iui.rst` documents the
  new assign output.

### Process
Adversarial pre-MR review (4 dims × 3 lenses, contract claims verified against
the teregen repo, not the plan): **it caught my `completeness()` hitting the
non-canonical path with a wrongly-documented shape** (must-fix, fixed), the
envelope shapes for `parsed`/`bugs` sections that my first tests mocked flat, a
duplicated CHANGELOG heading, and the misleading "already has" wording — all
folded in. `ruff` / `ty` (strict tree) / `sphinx -W` / full `pytest` (2433)
green locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,15 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
 
 ### Changed
 
+- The TeReGen client is synced with the current report API. `assign` now also
+  surfaces who already holds (or has decided) each review group, sourced from
+  the report's live per-group assignment map — context only, never a gate,
+  since an empty map can also mean the upstream lookup failed. New client
+  calls expose the parsed-report endpoints (`parsed`, `bugs` — with `bsc#`
+  ids percent-encoded — and `completeness`) for consumers without a local
+  checkout. The `updates` listing no longer crashes on a malformed deadline
+  value, and the `--review-group` help clarifies the bare group-name form.
+
 - `mtui-mcp` starts faster and ships a leaner tool manifest. The headless
   server no longer imports `prompt_toolkit` (~115 modules, ~170ms) at startup —
   it is loaded lazily only when the interactive REPL actually reads a line — and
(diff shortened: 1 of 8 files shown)


plusky:
For the record: superseded by #337 — crates/mtui-datasources/src/teregen.rs implements the full current /api/v1 surface (parsed/bugs envelopes, canonical /reports/{id}/completeness) and the assign/updates commands carry identical output and warn-only semantics.

PR #336, opened 2026-07-20
fix(downgrade): abort loudly instead of silently half-rolling back

plusky:
## The trap (field report)

On a slow ppc64le host, `downgrade` aborted twice with only a WARNING `downgrade not completed` — leaving **all 28 packages still at the TEST version**. An agent only caught it by re-verifying with `rpm -q` instead of trusting the tool.

Root cause chain:
1. The version probe ran `zypper -n se -s` **once per package** in a remote `for` loop, piped through `grep|grep|sed|awk`. Without a PTY, awk block-buffers — **zero bytes** reach the SSH channel until the whole loop ends. 28 sequential repo-metadata loads on slow hardware exceed the 300s no-output window (`connection_timeout`).
2. The headless session raised `CommandTimeoutError`; `Target.run` swallowed it as exit `-1` with empty stdout.
3. `perform_downgrade` never checked the probe's exit status: the empty output parsed to an empty version map, **zero downgrade commands ran**, and the flow "completed".
4. The final check (`before == after`) logged a bare WARNING and the command looked successful.

## The fix

- **One probe invocation.** `zypper -n se -s --match-exact -t package <all packages>` — one metadata load instead of N, finishing well inside the timeout. Empty package list returns early (zero names would list the entire catalog).
- **Dead probes abort, per host.** A host whose probe records a non-zero int exit is skipped entirely (no partial parse, no stale `-1` tripping later checks) and reported via `UpdateError` **after** the healthy hosts roll back and transactional ones reboot. All-probed-hosts-dead aborts immediately.
- **Exit `-1` is a hard failure in the downgrade checks** — and transactional hosts now have their own `("slmicro", True)` check: previously they fell back to the registry's silent no-op, so a dead `transactional-update` sailed on to a reboot with no snapshot staged.
- **Honest verdict.** The final verification compares each package's current version against the update's `required` version and names, per host at ERROR level with versions, everything still at/above the update's version. The old `before == after` warning fired spuriously on a double downgrade and stayed quiet in logs. New/multiversion packages are documented as expected residents of the list.
- **Headless callers get a real failure.** When non-interactive (mtui-mcp), `UpdateError` re-raises so the tool call fails instead of returning a success-looking log.

## Verification

- 14 new/extended tests across templates, checks, hostgroup, and command layers; the behavioral ones **proven to fail on revert** (mutation-checked: the `isinstance` gate and the no-short-circuit bookkeeping both have killing tests).
- Full gate set green: `ruff format --check`, `ruff check`, `ty check`, 2439 tests, strict docs build.
- Pre-open adversarial review (4 dimensions × 3 verification lenses): 2 must-fixes it found (the slmicro no-op check fallback, a CHANGELOG merge artifact) and the per-host abort semantics, MCP failure signaling, empty-list guard, and both mutation-verified test holes are all folded into this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,23 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   still-shutting-down sshd); only the serialization is removed. Most visible on
   transactional (SL Micro) updates, which reboot every host.
 
+### Fixed
+
+- `downgrade` can no longer end as a silent half-rollback. The version probe
+  now runs a single `zypper se -s` invocation for the whole package list
+  instead of one per package (the per-package loop, block-buffered behind the
+  pipeline, emitted no output until the end and blew the SSH no-output timeout
+  on slow hosts, e.g. ppc64le with many packages). When the probe or a
+  downgrade command dies (times out or fails to run) or hits a recognized
+  zypper failure, the flow now aborts with an explicit error instead of
+  "completing" with nothing rolled back — per host: a dead probe on one host
+  no longer cancels the rollback of the healthy ones. After a completed run,
+  the final check names, per host at ERROR level, every package still at or
+  above the update's shipped version — previously the only hint was a bare
+  "downgrade not completed" warning. Headless (MCP) callers now get a failed
+  command instead of a success-looking log. An empty package list no longer
+  probes the entire repository catalog.
+
 ## 19.0.1 - 2026-07-17
 
 ### Added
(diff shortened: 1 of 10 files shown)


plusky:
For the record: superseded by #337, which ports this fix explicitly (single-invocation probe, per-host dead-probe abort, exit -1 hard-fail incl. the transactional check, current >= required verdict — update_flow.rs and checks/downgrade.rs cite this PR) and adds an absolute command deadline on top.

PR #339, opened 2026-07-20
feat(commands): add request_review Slack review command and gate approve on it

plusky:
## Summary

Adds `request_review`, which posts a review request for the loaded update to a Slack channel, records the resulting message in the testreport, and optionally watches it for reviewer reactions. Enabling the integration also gates `approve` on an acknowledged request, so a requested review becomes something the tool can verify rather than a message someone may or may not have read.

Supersedes #257 (the Python implementation, closed at the Rust cutover). GitHub refused to reopen that PR because the branch was rebuilt on the Rust tree, so this is a fresh PR for the same feature; the reviewed semantics from #257 carry over.

**The integration is off by default.** Posting into a chat workspace is an outward-facing side effect, so it never happens implicitly: `[slack] enabled` must be `true` *and* both `token` and `channel` must be set. An mtui with no `[slack]` section never contacts Slack, `request_review` refuses with the reason and the config key that fixes it, and `approve` is completely unchanged.

## Changes

- **`[slack]` config section** (`enabled`, `token`, `channel`, `api_url`, `poll_interval`, `watch_timeout`). The token is a secret: masked as `<set>` by both `config show` and `config set`, and never logged. The API base is refused unless it is `https` (or `http` to loopback, which is what makes the wiremock suite possible), so a misconfiguration cannot put the bearer token on the wire in clear text.
- **`mtui-datasources::slack`** — a Web API client. Two Slack-specific behaviours shape it: application errors arrive as **HTTP 200** with `{"ok": false}`, so the envelope is checked rather than the status; and `429` is a typed `SlackError::RateLimited` treated as "keep watching" rather than a failure, so throttling cannot end a watch early and report "no reaction" when the truth is "we were not allowed to look".
- **`request_review` command** — posts, records, and with `--watch` polls for reactions (👍 approves, 👎 rejects, skin tone normalised, the bot's own reaction ignored). A rejection outranks a simultaneous approval. Ctrl-C prints a summary instead of killing mtui.
- **`Slack Review:` testreport marker** — written into the template and committed, so the marker is visible from another reviewer's checkout. Inserts when absent (the line never pre-exists in a generated template), replaces when present, and collapses duplicates so read and write agree.
- **`approve` gate** — with Slack enabled, refuses unless a marker exists, the marked message still names that RRID (so a marker copied between templates cannot launder an approval), and someone other than the bot acknowledged it. Unreadable Slack fails closed. No per-command bypass — a gate with one is not a gate; it is turned off by configuration, which is explicit and auditable. `reject` is never gated, because blocking it would strand an update a reviewer wants stopped.
- **MCP** — `request_review` joins `SLOW_COMMANDS` so it gains `background`, which is how a `--watch` run should be started there; a foreground watch would outlive any client timeout. Tool-surface snapshots regenerated.
- Docs (`docs/src/configuration.md`, generated `cli.md`) and CHANGELOG.

## Notable design decisions

- **The watch is opt-in (`--watch`), not implicit as in the Python original.** Posting is fast and total; watching is the long-running extra the caller asks for. This also makes the command behave sanely over MCP without a special case.
- **The marker is committed immediately** rather than riding a later `commit`: the person who approves is often not the person who asked, and the gate has to see it.
- The SVN runner is injected into the record step, because otherwise the success-path test passed via the failure branch (a tempdir is never a working copy).

## Checklist

The repo's PR template still lists the Python gates (`uv run ruff`/`pytest`/`Documentation/`), which the Rust rewrite replaced — worth updating separately. The actual gates, all observed locally:

- [x] `cargo fmt --all --check` is clean.
- [x] `cargo clippy --workspace --all-targets -- -D warnings` is clean.
- [x] `cargo test --workspace` passes.
- [x] `cargo test -p mtui-mcp --features mcp` passes.
- [x] Feature matrix builds (`--no-default-features`, `--all-features`).
- [x] User-visible changes recorded in `CHANGELOG.md`.
- [x] Docs under `docs/src/` updated; generated pages re-run via `cargo xtask gen-docs`.

52 tests added across the config, client, marker, command, and gate surfaces.

## Related issues

Supersedes #257.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,19 @@ _The entries below document the final Python releases, retained for history._
   still-shutting-down sshd); only the serialization is removed. Most visible on
   transactional (SL Micro) updates, which reboot every host.
 
+### Added
+
+- New `request_review` command: posts a review request for the loaded update to
+  a Slack channel and records the resulting message in the testreport template,
+  so a request is traceable back to the update it belongs to. With `--watch` it
+  then polls the message for reviewer reactions (👍 approves, 👎 rejects, skin
+  tone ignored) until a verdict or timeout, and Ctrl-C stops watching without
+  killing mtui. The integration is **off by default** and configured under
+  `[slack]`: set `enabled = true` plus a `token` and `channel` to opt in. The
+  token is masked as `<set>` in `config` output and never logged. Over
+  `mtui-mcp` the command is exposed with `background=true`, which is how a
+  `--watch` run should be started there.
+
 ## 19.0.1 - 2026-07-17
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,7 +60,7 @@ _The entries below document the final Python releases, retained for history._
 
 - New `request_review` command: posts a review request for the loaded update to
   a Slack channel and records the resulting message in the testreport template,
-  so a request is traceable back to the update it belongs to. With `--watch` it
+  committing it so the request is visible from any checkout. With `--watch` it
   then polls the message for reviewer reactions (👍 approves, 👎 rejects, skin
   tone ignored) until a verdict or timeout, and Ctrl-C stops watching without
   killing mtui. The integration is **off by default** and configured under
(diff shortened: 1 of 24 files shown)


mimi1vx:
@plusky  please fix test suite :D

plusky:
Test suite is fixed — CI is green on `040cfef` (7/7: tests, coverage, lint, format, feature matrix, both codecov checks). Sorry for the noise.

Two separate things were failing, only one of them mine:

1. **My flake.** `watch_times_out_with_an_elapsed_deadline_after_one_poll` asserted "exactly one poll" using a 1ms watch timeout, so whether the loop polled once or twice depended on whether the mock server answered inside that millisecond. It passed locally and on the first CI run, then failed on a re-run. Now uses a zero timeout, which puts the deadline at the loop's own start instant so it is unconditionally reached by the time the first poll returns — that is the property the test is actually about (the loop polls before it sleeps, so an already-expired watch still looks once). Confirmed with 25 consecutive runs.

2. **Not mine.** `ssh_integration::sftp_write_exclusive_then_overwrite_and_read_back` failed once with `Connect { reason: "Unknown server key" }` and passed on re-run — and also passed on #340, which changed only markdown. Looks like a host-key race in the sshd fixture rather than anything in this branch, but flagging it since it will bite again.

Chasing (1) turned up a real bug in the code, not just the test: a `Retry-After: 0` produced a zero-length back-off, so the watch would spin against the endpoint that had just asked it to slow down — `MAX_BACKOFF` capped the top but nothing floored the bottom. Now clamped to a 1s minimum, with a test that fails loudly if the spin returns.

Also pushed `4f7c319`: codecov had `metadata_parsers.rs` at 25% patch coverage, which was the marker's read-back arm — the half `approve` gates on. Covered now.

PR #342, opened 2026-07-20
chore(ci): add CodeQL, dependency-review and Scorecard; harden release build

plusky:
## Summary

Brings mtui's security tooling up to the same level as
[`openSUSE/repose`](https://github.com/openSUSE/repose), and hardens the release
build. The repository-level settings have already been enabled to match repose;
this PR is the in-tree half.

No Rust code is touched — this is `.github/`, `CHANGELOG.md` and `SECURITY.md`
only.

## Changes

**New security workflows** (from repose, retargeted `master` → `main`)

- `codeql.yml` — CodeQL for `rust` and `actions`, plus a weekly schedule so a
  newly published query still reaches code that no PR happens to touch.
- `dependency-review.yml` — blocks a PR that introduces a high-severity
  advisory before merge, where Dependabot only alerts once the code is already
  on `main`.
- `scorecard.yml` — OpenSSF Scorecard, publishing results.

All three are SHA-pinned, and every pin was verified to resolve to the tag it
claims. (repose still floats `codeql-action@v4`; pinning is what Scorecard's
Pinned-Dependencies check actually wants.)

**Release hardening (`release.yml`)**

- `cargo build --locked` for both binaries, so a published release is compiled
  from exactly the versions in `Cargo.lock` rather than whatever resolves at tag
  time.
- `persist-credentials: false` on checkout — the release is published via
  `gh release create` with `github.token`, so no git credential needs to be left
  behind in `.git/config`.

**`SECURITY.md`** was stale from the Python era: it asked reporters for their
`paramiko` / `openqa-client` versions and scoped out `ruamel.yaml`, `pyxdg` and
`osc`. Rewritten for the Rust tool — SSH pubkey-only posture, credential masking
via `is_secret_attr`, `sanitize_url` for logged URLs, and the loopback-only MCP
HTTP transport. It now also offers GitHub private advisories alongside the
existing qa-maintenance@suse.de address.

## Repository settings (already applied, listed for the record)

| Setting | Before | Now |
| --- | --- | --- |
| Secret scanning | disabled | enabled |
| Secret scanning push protection | disabled | enabled |
| Dependabot vulnerability alerts | disabled | enabled |
| Dependabot security updates | disabled | enabled |
| Private vulnerability reporting | disabled | enabled |

Secret scanning found 0 alerts across history. Private vulnerability reporting
had to be enabled for the "Report a vulnerability" link in `SECURITY.md` to work
at all — worth noting that repose links to the same form while having it
disabled.

## Notes for the reviewer

- **`deny-licenses` is intentionally unset in `dependency-review`.** mtui is
  `GPL-2.0-only`, but `russh`, `rmcp` and `ring` are Apache-2.0-**only**, which
  is generally read as GPLv2-incompatible. That predates this PR and wants a
  real decision (relicense to `-or-later`, or replace those stacks); gating
  every PR on it would just paint CI red without moving it forward. Flagging it
  here rather than silently encoding a policy either way.
- **No required status checks were added to the ruleset**, and the existing
  admin/maintainer bypass was left alone — both change merge behaviour and are
  yours to choose. Note mtui's ruleset is already stricter than repose's here
  (1 required approval + thread resolution, vs 0), so nothing was copied down.

## Checklist

- [x] `cargo fmt --all --check` is clean.
- [x] `cargo clippy --workspace --all-targets -- -D warnings` is clean.
- [x] `cargo test --workspace` passes.
- [x] `cargo test -p mtui-mcp -F mcp` passes.
- [x] Feature matrix — unaffected; no Rust or manifest changes in this PR.
- [x] Coverage — no Rust code paths added.
- [x] User-visible changes are recorded in `CHANGELOG.md`.
- [x] Documentation under `docs/src/` — not applicable.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,49 @@
+---
+name: codeql
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+  schedule:
+    # Weekly, so a newly published query still finds old code that no PR touches.
+    - cron: '31 5 * * 3'
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+jobs:
+  analyze:
+    name: Analyze (${{ matrix.language }})
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    permissions:
+      security-events: write
+      contents: read
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          # Rust is source-extracted, so no build orchestration is needed;
+          # the workflows themselves are scanned as `actions`.
+          - language: rust
+            build-mode: none
+          - language: actions
+            build-mode: none
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
+        with:
+          persist-credentials: false
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@7188fc363630916deb702c7fdcf4e481b751f97a  # v4
+        with:
+          languages: ${{ matrix.language }}
+          build-mode: ${{ matrix.build-mode }}
+      - name: Analyze
+        uses: github/codeql-action/analyze@7188fc363630916deb702c7fdcf4e481b751f97a  # v4
+        with:
+          category: "/language:${{ matrix.language }}"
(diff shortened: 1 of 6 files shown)


mimi1vx (approved):
LGTM. Actions pinned by SHA with version comments, minimal per-job permissions, concurrency groups with cancel-in-progress, `--locked` release builds — solid hardening. Appreciate the honest comment on the GPL-2.0-only vs Apache-2.0-only dependency tension instead of silently gating on `deny-licenses`; worth a follow-up issue to actually resolve, but not a blocker for this PR.

PR #344, opened 2026-07-20
feat(ci): configure 7-day cooldown for Dependabot GitHub Actions updates

okurz:
Motivation:
Reduce CI pipeline load and review fatigue by pacing automated updates.

Design Choices:
Configure cooldown with default-days set to 7 in the github-actions section
of .github/dependabot.yml.

Benefits:
Paces automated pull requests to match the established openQA and qem-dashboard
update patterns.

Related issue: https://progress.opensuse.org/issues/203049

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -21,6 +21,8 @@ updates:
     target-branch: "main"
     schedule:
       interval: "weekly"
+    cooldown:
+      default-days: 7
     open-pull-requests-limit: 5
     groups:
       actions-minor-and-patch:


plusky:
We have discussed it with @mimi1vx and we dont think this would benefit four our tools.

okurz:
> We have discussed it with @mimi1vx and we dont think this would benefit four our tools.

Can you please elaborate why? We applied the same change in many and multiple projects and I think mtui does not differ in that regard and the change would guard against some potential supply-chain attacks.

PR #346, opened 2026-07-21
refactor(commands)!: remove the lrun local-execution command

plusky:
## What

Removes the `lrun` command (run a command in the local shell) from mtui entirely — from the command registry (so it is gone from the REPL) and from the MCP deny-list machinery that existed only to suppress it.

## Why

Design choice: the command earns its keep on neither surface.

- **Over MCP it never made sense.** The client already has its own local execution, which is exactly why `lrun` was permanently deny-listed there from the start. A command that exists only to be denied is dead surface.
- **In the REPL the tester is already at a shell.** Prefixing a local command with `lrun` bought nothing over running it directly — `run` remains for commands on the reference hosts.

## How

Removal, not denial:

- `crates/mtui-core/src/commands/localrun.rs` deleted; unregistered from `register_all()`; `"lrun"` dropped from `MCP_DENYLIST`, which is now purely the interactive/REPL-only set.
- A new registry test `lrun_is_fully_removed` pins the removal: absent from the registry **and** from the deny-list, so no stale entry can linger (the boot drift-warning stays quiet).
- MCP tests that used `lrun` as their deny-list specimen now use `shell` — same assertions (`tools_allow` cannot restore a denied command; the HTTP factory denies it; a denied *alias* blocks synthesis; a denied call is `method_not_found`), still meaningful.
- **The MCP tool surface is unchanged** — `lrun` was never synthesised, so downstream LLM configs are unaffected (all MCP schema/surface snapshots are byte-identical).
- Docs: README security boundary, `docs/src/mcp.md`, `docs/src/architecture.md`, `AGENTS.md` (now: removed by design, do not reintroduce); `docs/src/cli.md` regenerated via `cargo xtask gen-docs`; help-listing snapshot refreshed; CHANGELOG `### Removed` bullet appended.
- A user typing `lrun` now gets the ordinary unknown-command error (existing `unknown_command_renders_error_and_continues` covers the path); completion and highlighting are registry-driven, so nothing else changes.

**BREAKING CHANGE:** the `lrun` REPL command no longer exists.

## Testing

- Full gate: `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace` (all 37 test binaries green), feature matrix (`--no-default-features` + `--all-features`) — all pass.
- Adversarial review (3 independent read-only reviewers: completeness, MCP-contract, runtime-behavior): one stale doc sentence found (`architecture.md`) and fixed; MCP contract and behavior lenses came back clean.


mimi1vx (approved):
LGTM. Complete, well-documented breaking removal (impl, registration, MCP deny-list, docs, CHANGELOG). Verified zero orphaned lrun/localrun references outside intentional historical/explanatory prose. Rebased onto current main (which had since gained request_review, bumping the canonical command count to 62): recomputed count is 61 and the help-listing snapshot now correctly includes request_review without lrun. Ran cargo fmt --check, clippy --workspace --all-targets -D warnings, and cargo test --workspace on the rebased branch — all clean.

PR #347, opened 2026-07-21
fix(refhost): single-flight concurrent HTTPS cache refreshes

plusky:
## What

Single-flights concurrent `refhosts.yml` HTTPS cache refreshes: when several callers hit a stale (or missing) cache at once, exactly one downloads while the rest wait and then read the freshly written cache.

## Why

Every call site (Session autoconnect, `list_refhosts`, `add_host`) builds its own `RefhostsFactory` on demand, so N concurrent resolves each ran their own `stat → fetch → write` sequence — N duplicate downloads of the same file stampeding the refhosts server, plus N racing (atomic, but redundant) cache writes.

## How

- A process-wide async lock registry keyed by cache path (`REFRESH_LOCKS`), because the per-call construction means concurrent resolves never share a resolver instance. Production configures exactly one cache path, so the map holds one entry.
- `refresh_if_needed`: unlocked fast path when fresh → lock → **re-check under the lock** → fetch only if still stale. The leader's atomic write bumps the mtime, so waiters build from the fresh cache.
- **Failures don't multiply either**: a waiter that waited through a *failed* download adopts that outcome (new `RefhostError::RefreshJustFailed`) instead of serially repeating a doomed attempt — a down server costs the group one transport timeout, not one per waiter, and each caller falls back to the next configured resolver exactly as before. A later caller retries normally, so nothing latches.
- No `std` mutex guard is ever held across an await; the tokio guard is `Send`, keeping the resolver futures `Send` for the Session callers.

Upstream Python had no locking (single-threaded); this is an intentional improvement, and refresh behaviour is not one of the documented interop contracts.

## Testing

- `concurrent_stale_resolves_share_one_fetch` — 4 resolver instances, one missing cache, real clock/stat/atomic-writer seams: exactly 1 download. **Mutation-checked**: with lock-sharing disabled the test fails (4 downloads observed).
- `waiters_fail_fast_after_concurrent_failure` — down-server shape: 1 download total; leader surfaces the transport error, waiter gets `RefreshJustFailed`.
- `failed_refresh_is_retried_not_latched` — a later caller after a failure retries and succeeds.
- Full gate: fmt, clippy `-D warnings`, `cargo test --workspace` (37/37 binaries), feature matrix — green.
- Adversarial review (2 independent read-only reviewers: concurrency-correctness, behavioral-parity): the concurrency lens confirmed lock hygiene/Send/mtime-recheck soundness and flagged the failure-path serialisation latency — addressed here with the fail-fast waiter behaviour; the parity lens found two doc nits, folded.

## Out of scope (pre-existing, surfaced during review)

`config set refhosts_https_expiration 0` bypasses the positivity validation the config-file loader applies (`set_attr` uses bare `parse_u64`), contradicting the adjacent comment that runtime `set` cannot store a value the file would reject. With `0`, caching is effectively disabled (though refreshes stay serialised under this PR's lock, so still no herd). Will address separately.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,6 +77,16 @@ _The entries below document the final Python releases, retained for history._
   explicit and auditable. `reject` is never gated. With Slack disabled (the
   default) `approve` behaves exactly as before.
 
+### Fixed
+
+- Concurrent `refhosts.yml` HTTPS refreshes no longer stampede the server.
+  When several sessions or commands hit a stale cache at once, one of them
+  downloads while the rest wait and then read the freshly written cache —
+  a single download at a time instead of one per caller. If that download
+  fails, the callers that waited on it fail over to the next resolver
+  immediately instead of each repeating the doomed download, so a down
+  server costs one transport timeout, not one per caller.
+
 ## 19.0.1 - 2026-07-17
 
 ### Added
--- a/crates/mtui-datasources/src/error.rs
+++ b/crates/mtui-datasources/src/error.rs
@@ -92,6 +92,18 @@ pub enum RefhostError {
     /// happen; this variant is the terminal "all strategies exhausted" signal.
     #[error("no refhosts resolver could produce a usable database")]
     ResolveFailed,
+
+    /// A concurrent HTTPS cache refresh failed while this resolve waited on it.
+    ///
+    /// The waiter does not retry the download itself: the failure it just
+    /// waited through would almost certainly repeat, and each retry costs a
+    /// full transport timeout — serialised behind the refresh lock, that would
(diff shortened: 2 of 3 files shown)


mimi1vx (approved):
LGTM. Verified the single-flight refresh logic by hand and ran the two timing-sensitive concurrency tests (`concurrent_stale_resolves_share_one_fetch`, `waiters_fail_fast_after_concurrent_failure`) 8x locally with no flakes. The `wait_started`/`failed_at` Instant comparison correctly distinguishes a waiter that observed the in-flight failure from a later caller that should retry independently. Correct use of `tokio::sync::Mutex` held across the fetch await for the actual single-flight semantics.

PR #374, opened 2026-07-25
feat: cooperative cancellation at the boundaries that can express it

plusky:
> **Stacked on #373** — the `Session` cancellation seam it builds on lands there. Review/merge that first.

## Problem

#373 gave `Session` a cancellation token and had the dispatch driver check it, but nothing *long-running* observed it. Cancelling a single-template `run`/`update` still burned the full cooperative grace and force-aborted, because all the wall clock is spent inside one `perform_*` call. This makes a cancel actually stop work.

## Where cancellation is now checked

Only at explicit, safe boundaries:

| Boundary | Why there |
|---|---|
| `Command::run` (from #373) | pre-dispatch and between fan-out templates |
| Serial per-package loops — `prepare --installed-only`, `downgrade` | the longest interruptible stretches: one SSH fan-out **per package** |
| `perform_update` entry, and again immediately before the patch is dispatched | the last point where stopping is free |

**A fan-out is never interrupted part-way**, and that is deliberate — see below.

## The design decision that matters

The first version of this branch gated cancellation *inside* the shared fan-out primitive, skipping hosts that had not started yet. Three independent adversarial reviews showed that is unsafe, so it was removed:

- `run_parallel` returns `()`. A **skipped host is indistinguishable from a successful one** — its `Target` still holds the *previous* command's `last*` snapshot, so `run_checks` reads stale clean output and passes it. A cancelled `update` returned `Ok(())` with hosts unpatched, and the MCP job was recorded **successful**.
- Every teardown is itself a fan-out, so `unlock`/`close`/`pool_unlock` were skipped exactly when a cancel made them matter — leaving refhosts locked against the whole team. Teardown paths (`quit`, `unload`, MCP `close`) bypass `Session::activate`, so they inherited a stale cancelled token.
- `perform_update`'s own cancel-cleanup was gated on the token that triggered it, making it a guaranteed no-op.

Whole-batch granularity keeps "ran" and "did not run" distinguishable. The wall-clock win survives where it actually mattered: the serial package loops.

## Two adversarial review rounds shaped this

Round 1 (three reviewers) killed the original design — gating inside the fan-out. Round 2 (on the rework) found three defects the rework itself introduced, all fixed here:

- The per-package loops early-`return`ed past their own fall-through, so `prepare` skipped `reboot_transactional` — leaving a transactional host with an inert snapshot while reporting the packages as installed. They now `break`, and the message says "applied".
- Those early returns also discarded per-host failures collected before the cancel — the exact masking this design claims to prevent, one layer lower. A real failure now outranks a cancellation in both loops and in the fan-out driver.
- A cancel landing during the update run phase reached the **rollback's** own checkpoint and aborted it at package 0, stranding the half-applied update the rollback exists to undo. The rollback now runs under `suspend_cancellation`.

## Honest reporting

- A flow that stops at a checkpoint marks its own error `cancelled` (`UpdateError::cancelled`). The command layer keys on **that**, never on the session token — sniffing the token would report a genuine host failure that merely coincided with a cancel as a bare "cancelled", hiding e.g. a dead connection.
- `CommandError::Cancelled(String)` carries detail, so a cancelled package loop reports *which* packages were applied and which were not attempted, rather than an unactionable "cancelled" — and the MCP job layer records that verdict on a cooperatively-stopped job, so `job_result` hands it back instead of a fixed "was cancelled" string.
- A genuine host failure **always outranks** a cancellation, so a broken host is never buried behind "cancelled".
- `UpdateFailure::Cancelled` skips the rollback (the update never dispatched, so there is nothing to undo, and a rollback downgrade is itself minutes of work the caller just asked to stop). Its docs state plainly that a `prepare` which already ran is **left in place** — this is not a "nothing happened" verdict.

## Bug fix (found while wiring this)

`HostsGroup::select` / `select_split` rebuilt the split halves copying only `plan_provider`, silently resetting `max_parallel` to the built-in default. `perform.rs` splits the group on **every** `install`/`uninstall`/`prepare`/`downgrade`/`update` — including with no `-t` — so every one of those commands has been fanning out at the default 50 instead of the configured `[connection] max_parallel`. Both halves now carry the pushed-down state, with regression tests.

Note for operators who tuned the value: a `-t` subset now correctly honours `max_parallel`, so someone running `max_parallel = 200` will see wider fan-out than before.

## Validation

Full workspace gate green: `fmt --check`, clippy `-D warnings` (all targets), `cargo test --workspace`, `cargo test -p mtui-mcp -F mcp`, compile-only feature matrix. New tests: fan-out primitive is cancellation-free, `select`/`select_split` state carry (both), `perform_update` entry gate runs no host command, downgrade cancel names its progress, and an inertness test proving an uncancelled flow is unaffected.

## Known, pre-existing, not addressed here

`select_split` returning `Err` (e.g. `update -t typo-host`) drops the moved-out `HostsGroup` without `close()`, leaking remote locks. Predates this branch; worth its own fix.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -138,12 +138,32 @@ next actionable task before working on a subsystem.
   `lrun` was removed by design; do not reintroduce it.
 - **Cancellation is cooperative-first.** `Session` carries a
   `CancellationToken` (the seam); the `Command::run` driver checks it before
-  dispatch and between fan-out templates, and a long-running body may poll
-  `session.cancel_requested()` (or `select!` on `session.cancel_token()`) at
-  its own host/step boundaries. MCP `job_cancel` cancels the job's token,
-  waits a short grace for a cooperative stop, then hard-aborts the worker —
-  its reply distinguishes the two and never claims to cancel a job that had
-  already finished. New long-running command bodies should observe the seam.
+  dispatch and between fan-out templates. `Session::activate` pushes the token
+  down onto the active report's `HostsGroup` so the flows can consult it.
+  **Never gate a parallel fan-out on it**: `run_parallel` returns `()`, so a
+  host skipped mid-batch is indistinguishable from one that ran — its stale
+  `last*` snapshot sails through the post-run checks and the update reports
+  success on a host it never touched (and every teardown is a fan-out too, so
+  gating strands remote locks exactly when a cancel needs them released).
+  Long serial flows poll `targets.cancel_requested()` at
+  their own boundaries (the per-package `prepare`/`downgrade` loops, the
+  `update` step sequence) — but **never past a point of no return**: `update`
+  makes its last check before dispatching the patch command, and the
+  post-failure rollback runs under `HostsGroup::suspend_cancellation` (its own
+  per-package checkpoint would otherwise abort the recovery at package 0 and
+  strand the half-applied update it exists to undo). `UpdateFailure::Cancelled`
+  skips the rollback because the update never ran. A checkpoint must `break`
+  into its function's normal fall-through, never early-`return` past cleanup —
+  and a real failure collected before the cancel always outranks it. MCP `job_cancel` cancels the job's token, waits a short grace for a
+  cooperative stop, then hard-aborts the worker — its reply distinguishes the
+  two and never claims to cancel a job that had already finished. A flow that
+  stops on a cancel must surface as `CommandError::Cancelled`, not a generic
+  failure — and that verdict must come from the flow's own `cancelled` flag,
+  never from sniffing the session token, which would mask a real host failure
+  that merely coincided with a cancel (see
+  `commands/perform.rs::map_flow_error`). New long-running command bodies
+  should observe the seam at their own step boundaries, and report what they
+  completed before stopping.
 
 ## Contracts (do not break without intent — these enable ecosystem interop)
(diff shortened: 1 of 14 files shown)


mimi1vx:
@plusky please rebase PR

plusky:
Rebased onto current `main` (`6b09dcd1`). #373 was rebased first and this branch on top of it, so the stacking is unchanged — **#373 still needs to merge first**.

The rebase was clean: no conflicts, and the CHANGELOG kept both entries. Full gate re-run green on the rebased tip — `fmt --check`, clippy `-D warnings` (all targets), `cargo test --workspace`, `cargo test -p mtui-mcp -F mcp`, and the compile-only feature matrix.

plusky:
Rebased onto current `main` (`e32359a5`).

The conflict was #373 — this branch carried its commit, and #373 rebase-merged upstream under a different SHA. Now that #373 is in `main`, that commit drops out and this PR is a single commit, fast-forward from `main`. The four dependabot bumps that landed in the same window (incl. `serial_test` 3.5.0 → 4.0.1) are included in the rebase.

No longer stacked on anything — mergeable on its own.

Full gate re-run green against the new base: `fmt --check`, clippy `-D warnings` (all targets), `cargo test --workspace`, `cargo test -p mtui-mcp -F mcp`, and the compile-only feature matrix.

mimi1vx (approved):

PR #386, opened 2026-07-27
fix: report a transactional host whose reboot did not take effect

plusky:
## The bug

On a read-only-root host (SL Micro) the packages an operation installs are staged into a snapshot that **only the reboot activates**. A host that fails to come back has therefore not had the operation applied, however cleanly its command exited — but the failed reconnect was logged at ERROR and dropped, so `install` printed `install completed on <hosts>` for a machine that was no longer reachable.

This is the same shape as #383 (install/uninstall ran nothing while reporting success), one layer further along.

Three distinct ways a transactional reboot could lie, all now reported:

| | before | now |
|---|---|---|
| host never came **back** | logged at ERROR, dropped | named, with its reason |
| host never went **down** (reboot silently no-ops) | invisible — reconnect succeeds against the still-live host | caught by the boot-id check |
| reboot never **dispatched** | invisible — the failed dispatch leaves the session open, so the reconnect succeeds trivially | recorded, and outranks any later verdict |

All five flows that reboot a transactional host were affected. `install`, `uninstall`, `prepare` and `update` reported outright success; `downgrade` reported only a vague "downgrade not completed" from the version re-probe the dead host then failed.

## What changed

`reboot_transactional` returns a per-host outcome map instead of `()`, collected through a side channel because the fan-out closure must keep returning `()` — a host skipped mid-batch has to stay indistinguishable from one that ran. `Operation::run` folds it into a new `HostError::RebootFailed`, reported **after** the unlock and never propagated with `?`, so the group lock is still released on every host.

That variant deliberately means the opposite of every other error the template returns: the command ran everywhere and only the reboot failed, so `perform_operation` routes the two apart and still computes the per-host verdict — the dead host is named *alongside* it, not instead of it.

`update_run_phase`'s hand-rolled aggregation was a copy of `aggregate_failures`, so it now calls it; the rendered message is byte-identical for the pre-existing single- and multi-host cases.

Also here, because they are the same seam:

- **`install`/`uninstall` observe cancellation** (they were the only long-running flows with no checkpoint at all — a `job_cancel` let the whole thing run to completion). Deliberately one gate, on entry: every later step is a parallel fan-out, and skipping the reboot would leave a snapshot inert while the packages reported installed. The message says a history line was already *dispatched* rather than implying a clean slate.
- **`From<HostError> for UpdateError`**, so host failures are lifted in one place instead of a `reason_only(e.to_string())` at each call site that silently hard-codes `cancelled: false`. No variant represents a cancellation today; a test pins that, so a future `#[non_exhaustive]` variant cannot quietly acquire the wrong flag.

A failed reboot on `update` stays an `UpdateFailure::Check`, so it still triggers the rollback. That is intended: the host is on the pre-update packages while the rest of the group is on the post-update ones, and a split-brain group is worse than a uniform pre-update one.

## Review

Adversarial review before committing, four reviewers on distinct lenses (remaining false-success paths / cancellation contract / test quality / contracts + both surfaces). It changed the work:

- the cancellation message claimed history "was already recorded on each host" — `add_op_history` is best-effort, so that was false for an unreachable host;
- the message used `as_operation_role()`, which renders **`installer`**, not the verb the operator typed — and the test's `contains("uninstall")` passed vacuously against `uninstaller`;
- `Target::reboot` swallowed its dispatch error — the third row of the table above, found by the reviewer, not by me;
- `docs/src/configuration.md` scoped `reboot_timeout` to `prepare`/`reboot`/`update`; this change is what made the omission user-visible;
- two CHANGELOG claims were wrong (`downgrade` never reported success; on `update` a reboot failure and a per-host verdict cannot co-occur).

Mutation-verified: 13 mutants, all killed. One survived the first pass — nothing tested that `prepare` propagated a reboot failure — which is why the prepare and downgrade tests exist. The tree was restored byte-identically (`git diff | sha256sum`) after every mutation.

Deferred to their own issues rather than smuggled in here: `update` has no verdict at all on slmicro/YUM hosts (no check table *and* no exit-code fallback), disabled transactional hosts are rebooted and reported installed, and a cancelled template is reported as a fan-out failure.

## Gate

`fmt` · `clippy -D warnings` · `cargo test --workspace` · `cargo test -p mtui-mcp -F mcp` · feature matrix (`--no-default-features` + `--all-features`) · `typos` — all green locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -169,6 +169,14 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   host has passed its check, so the two cannot arise together. A failed reboot
   on `update` is treated as any other per-host update failure, so it still
   triggers the rollback.
+- A transactional host that never *went down* was also reported as a success.
+  The post-operation reboot is dispatched fire-and-forget, so its exit status
+  is never seen: a `transactional-update reboot` that silently no-ops (a
+  systemd inhibitor, a masked rebootmgr, a polkit denial) leaves the host
+  reachable and its staged snapshot inert, and reconnecting proves only that
+  the host is up. That reboot now snapshots each host's boot id beforehand and
+  re-reads it after, as the `reboot` command already did. A boot id that cannot
+  be read is still not a failure, so an unreadable probe cannot manufacture one.
 - A reboot that could not be *dispatched* at all — an unconnected target, or a
   channel that would not open — was also dropped. Because a failed dispatch
   leaves the SSH session open, the reconnect that follows succeeded trivially
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -156,6 +156,31 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   error reports only the role and release, and a host whose product never parsed
   has no release, so the message read `Missing Installer for ` — while aborting
   the operation for every host in the group.
+- A transactional host that never came back from its post-operation reboot was
+  not reported. On a read-only-root host (SL Micro) the packages are staged into
+  a snapshot that only the reboot activates, so a host that fails to reconnect
+  has *not* had the operation applied however cleanly its command exited — yet
+  the failed reconnect was logged and dropped. `install`, `uninstall`, `prepare`
+  and `update` reported outright success; `downgrade` reported only a vague
+  "downgrade not completed" from the version re-probe that the dead host then
+  failed. All five now name the host and why it did not return. For `install`,
+  `uninstall`, `prepare` and `downgrade` that appears alongside the per-host
+  verdict rather than instead of it; on `update` the reboot only runs once every
+  host has passed its check, so the two cannot arise together. A failed reboot
+  on `update` is treated as any other per-host update failure, so it still
+  triggers the rollback.
+- A reboot that could not be *dispatched* at all — an unconnected target, or a
+  channel that would not open — was also dropped. Because a failed dispatch
+  leaves the SSH session open, the reconnect that follows succeeded trivially
+  and the host looked rebooted. Both the post-operation reboot and the `reboot`
+  command now report it.
+- `install`/`uninstall` now observe cooperative cancellation (MCP `job_cancel`)
+  at an entry gate, stopping before the group lock is taken instead of running
+  the whole operation to completion, and reporting the stop as a cancellation.
+  This is deliberately the only checkpoint in that flow: every later step is a
+  parallel host fan-out, and skipping one would leave a host's stale state to
+  pass the verdict as success — skipping the reboot in particular would leave a
+  transactional host's snapshot inert while the packages reported installed.
 
 ## [26.0.1] - 2026-07-22
 
--- a/crates/mtui-hosts/src/connection/mock.rs
+++ b/crates/mtui-hosts/src/connection/mock.rs
@@ -129,6 +129,10 @@ pub struct MockConnection {
     reconnects: Arc<Mutex<usize>>,
     /// When `true`, [`reconnect`](Connection::reconnect) fails.
     reconnect_fails: bool,
+    /// When `true`, [`fire_and_forget`](Connection::fire_and_forget) fails
+    /// *without* tearing down the local link, modelling a reboot command that
+    /// never reached the host and so left the session live.
+    fire_and_forget_fails: bool,
     /// The `(retry, backoff)` args of the most recent
     /// [`reconnect`](Connection::reconnect) call, so a test can assert the
     /// caller passed the expected budget (e.g. the reboot lifecycle's
@@ -247,6 +251,7 @@ impl MockConnection {
             close_fails: false,
             reconnects: Arc::new(Mutex::new(0)),
             reconnect_fails: false,
+            fire_and_forget_fails: false,
             last_reconnect_args: Arc::new(Mutex::new(None)),
             fired: Arc::new(Mutex::new(Vec::new())),
             sftp_ops: Arc::new(Mutex::new(Vec::new())),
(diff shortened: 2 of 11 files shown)


plusky:
Superseded by #385, which landed a few minutes ahead of this one and was developed independently. #385 covers the same reboot-failure surfacing and the same install/uninstall cancellation gate, and additionally fixes check-before-reboot and the history-write timing, which this PR did not touch. Its `OperationReport` is also the better contract: keeping `Err` to mean "never started" and returning per-host failures on the `Ok` path is cleaner than the `HostError::RebootFailed` variant used here, which had to invert the meaning of `Err` and warn callers about it in a doc comment.

Closing in favour of a focused follow-up. Recording what this PR found that #385 does not cover, so it is not lost:

**Two reboot failure modes are still undetected on main.** #385 detects a transactional host that never came *back* (the reconnect failed). It does not detect:

- **The host never went down.** `HostsGroup::reboot_transactional` (`crates/mtui-hosts/src/target/hostgroup.rs:1201`) takes no boot-id snapshot — its own doc comment says so: *"Unlike `reboot` this path takes no boot-id snapshot / verification"*. The machinery already exists and is used by the explicit `reboot` command a few hundred lines above; the post-operation path just does not call it. A reboot that silently no-ops (masked `rebootmgr`, a systemd inhibitor, a polkit denial) leaves the host reachable, so the reconnect succeeds and the operation reports success against a snapshot that was never activated.

- **The reboot was never dispatched.** `SshConnection::fire_and_forget` (`connection/ssh.rs:1173`) tears down the local link with `self.close().await` as its last statement, reached only after `channel_open_session` and `exec` both succeed. `reconnect` (`ssh.rs:1128`) opens with `if self.is_active() { return Ok(()) }`. So a failed dispatch leaves the session active and the following reconnect returns `Ok(())` immediately, in zero time. `Target::reboot` (`target/mod.rs:579`) returns `()` and only logs at ERROR, so nothing propagates. The same two no-ops compose on an unconnected target, where `reboot` logs *"reboot on unconnected target"* and returns while `reconnect` returns `Ok(())` — two "not connected" no-ops making a clean success.

**`OperationReport` is missing `#[must_use]`** (`target/operation.rs:216`). The doc directly above it notes that a caller discarding the report "cannot tell a transactional host that rebooted into an unreachable state from one that came back healthy" — but `op.run(&mut group).await?;` compiles clean and drops both failure lists, which is the same class of bug the report was added to fix. One-line fix.

**`update` still has no verdict at all on SL Micro / YUM hosts.** `run_checks` bails out via `else { continue }` when the registry has no entry for `(release, transactional)` (`reports/update_flow.rs:266-268`), and the update check table is `("11" | "12" | "15" | "16", false) => Some(...), _ => None` (`update_workflow/checks/update.rs:107-108`). Transactional hosts never match. The new exit-code fallback added in #385 lives in the `PlanProvider::check` adapter, which `run_checks` does not go through, so this survives both PRs — `transactional-update patch` can exit non-zero and `update` still reports success. Worth its own issue; I am happy to take it.

**On the rollback routing.** #385 skips the rollback for a reboot failure; this PR triggered it. Both are blanket rules, and on reflection neither is right in general — this PR's rationale was the weaker of the two. The rollback is group-wide (`perform_downgrade` takes the whole `HostsGroup`), so triggering it for a host that never came back downgrades the healthy remainder on behalf of a machine it cannot reach, and `downgrade_body` reboots the transactional hosts again while `perform_update_with_rollback` logs and discards that second pass's failures — so one lost host can quietly become two. #385's skip is right for that case.

But it is the wrong call for a host that is reachable and sitting on an un-activated snapshot, which is the genuinely split-brained state and exactly what a rollback repairs. That distinction is only available once the two modes above are detected, which is what the follow-up will add.

Follow-up PR to come, built on `OperationReport` rather than on this branch's design.

mimi1vx:
Superseded by #387 — flagging rather than closing, since the analysis here is what #387 is built on.

**Why a new branch.** This PR branches from `1ce6c41d`; `main` has since landed four commits (`a434fabe`, `3d3c2b8c`, `18d74194`, `1780a6e6`) solving the shared part with a different design — `OperationReport` as a success value, versus this PR's `HostError::RebootFailed`. Merging is a 6-file conflict covering the whole of `operation.rs`, `hostgroup.rs` and `update_flow.rs`: a reimplementation, not a rebase.

`main`'s shape is the better one. Returning "the operation ran, here is what failed" as `Ok(report)` keeps `Err` meaning "never started", where this PR has to document a variant that inverts its own enum's contract ("this one means the opposite of every other variant").

**What #387 keeps from here.** The two things `main` still lacks, re-expressed against `OperationReport`:

- the boot-id check on `reboot_transactional` — `main`'s doc comment still says this path "takes no boot-id snapshot / verification";
- `Target::reboot -> Result`, the undispatched-reboot case. That is the sharpest find in this PR and nobody else spotted it: a failed dispatch leaves the session live, so the reconnect succeeds trivially and every later probe reads a healthy host.

Commits 1–3 here are superseded; the cancellation entry gate and `From<HostError> for UpdateError` are already on `main` in equivalent form.

**What #387 changes on top.** Three defects found reviewing this branch:

1. All three causes rendered as `did not come back after the reboot`, which is false for two of them — a host caught by the boot-id check *did* come back, and one whose dispatch failed never went anywhere to come back from. The table in this PR's description promises the distinction; the code collapses it, and no test pinned the rendered string.
2. `HostError::Update(reason)` reused for the boot-id verdict after adding a dedicated variant for the other cause.
3. `plans()` has no state filter, so `Target::reboot` returning `Err(NotConnected)` turns a disabled/unconnected host into a hard operation failure. (My own first fix for this was also wrong — gating only `Target::reboot` half-gates the lifecycle, since `boot_id` and `reconnect` aren't gated; #387 excludes such hosts from all four phases instead.)

Happy to hand #387 over or fold in anything you'd rather keep from this branch, @plusky.

mimi1vx:
Correcting my comment above — it was wrong twice over, and I'd rather not leave it standing.

**It attributed the supersession to the wrong PR.** @plusky had already closed this citing #385, 85 minutes before I commented. I wrote "superseded by #387" without re-reading the thread. #385 is what superseded this; #387 did not exist yet when it was closed.

**And #387 no longer contains the work that comment described.** #388 landed the boot-id check and the undispatched-reboot detection — the two things I'd argued were worth salvaging from here — and did it better than either this branch or my #387: typed `RebootFailureCause` carried to the caller instead of strings, per-cause wording, and rollback routing on the cause (`Unreachable` skips the rollback; still-reachable takes it). That last part neither this PR nor mine attempted, and it is the part that actually matters.

I've rewritten #387 down to two things #388 doesn't touch, both of which its routing turned from cosmetic into destructive: `fire_and_forget` not reviving an idle link (failed dispatch + successful reconnect is the `NotDispatched` signature, so an idle TCP session rolls the whole group back), and disabled hosts still reaching the reboot map (rebooted despite staging nothing, and reporting `NotRebooted` → group-wide rollback for a host that was never in the run).

So: this PR's analysis of the undispatched-reboot case was right and is now in `main` via #388, which is the correct outcome. Sorry for the noise on a closed thread, @plusky.

PR #390, opened 2026-07-27
docs: add project logo and GitHub social preview header

plusky:
## What

A project logo and a GitHub social preview header.

- **`docs/assets/logo.svg`** — a flat two-colour vector mark: a package cube
  with a checkmark inside a refresh cycle ("tested maintenance updates" as a
  pictogram), SUSE green `#88B531` on dark charcoal `#081E26`. Native SVG,
  4 paths, embedded at the top of the README (right-aligned, 130 px).
- **`docs/assets/social-header.png`** — a 1280×640 social preview: the mark on
  a white tile with a `$ mtui` wordmark and tagline. GitHub does not read this
  from the repo — a maintainer uploads it once under
  **Settings → Social preview**.
- **`docs/assets/social-header.svg`** — the header's source, so the PNG can be
  regenerated if the tagline changes.

## Preview

| logo | social header |
|---|---|
|  |  |

## Notes

- The mark was AI-generated (Recraft V4.1 Vector, via Picsart); the header is
  composed from it locally, so the two stay pixel-identical. The generator's
  Ed25519 provenance signature is kept in the SVG `<metadata>`.
- No CHANGELOG entry: repo branding, not tool behaviour.
- If you'd prefer a transparent-background variant (the SVG's first path is a
  plain white canvas rect), or different placement/size in the README, happy to
  adjust.


mimi1vx (approved):

plusky:
Pushed a follow-up commit putting the logo in the mdBook as well:

- `docs/src/assets/logo.svg` — moved under `src/` because mdBook only copies files from there. It stays the **single** copy; the README now points at that path rather than a second one being kept in sync.
- Shown in `docs/src/introduction.md`.
- `docs/theme/favicon.svg` overrides mdBook's default favicon (a partial theme dir leaves the rest of the default theme alone).
- `docs/assets/` keeps only the social-header pair, which is uploaded to Settings and is not book content.

Verified with a local `mdbook build`: `assets/logo.svg` is copied into the output and referenced correctly from `introduction.html`, the emitted favicon is byte-identical to ours, and no asset reference in the built book is broken.

PR #393, opened 2026-07-27
fix: give update a verdict on SL Micro/YUM, keep history rows, stop rebooting into failed transactions

plusky:
Three silent failures in the update/prepare/downgrade flows. Five commits:
one per fix, bisectable and individually green; a fourth carrying the
adversarial review's findings against the first three — kept separate because
three of four reviewers independently caught the first commit's new `-1` gate
routing an unreachable host into the group-wide rollback, and that correction
is worth reading on its own; and a fifth, doc-only, answering the human
review's design question in the routing docs.

## 1. `update` had no verdict at all on SL Micro and YUM

`("slmicro", true)` and `("YUM", false)` have had an *updater* since the port
but no entry in the *check* table, so `run_checks` took its
`else { continue }` and those hosts contributed nothing. `update` reported
success however the patch went — and on SL Micro the host was then rebooted
into the failed transaction.

**SL Micro is judged on output markers, not the exit code.** Its template is a
multi-line script run as a single remote `exec` whose last line is the
`zypper -n lr | … | while read r; do zypper -n rr $r; done` cleanup loop. The
shell reports *that loop's* status — `0` whenever the body never runs — and
discards the `transactional-update` status entirely. An exit-code rule would
be wrong in both directions: a failed patch still exits 0, and a hiccup in the
cosmetic cleanup would read as a failed update — which, because an update
check failure routes the **group-wide rollback**, would roll the whole fleet
back over a no-op.

**YUM is judged only on whether the command ran.** That key is *every* `rhel`
version; on RHEL 8/9 `yum` is `dnf`, whose handling of a spec matching nothing
installed differs from yum 3's, and mtui hands the updater the whole package
list while a refhost carries only a subset. The `Error:` marker is no better —
the template runs `yum repolist` in the same `exec`, so an unreachable repo
lands in the same stderr. Either rule would fail a host that updated fine and
take the fleet down with it. Settling them needs observed output from a real
RHEL refhost, so the check claims only what it can prove.

Also adds the **"timed out or failed to run" gate** the downgrade check has and
this one lacked, on *every* key including plain zypper. `Target::run` records
`-1` for a timeout, a mid-command connection loss, or an unconnected target,
and such a run leaves stdout and stderr empty — so no marker matched and a
patch that never completed passed.

Crucially `-1` does **not** route the rollback (new `UpdateFailure::NotRun`).
It means the flow could not talk to the host, so a group-wide downgrade cannot
repair it and would only revert the hosts that patched cleanly — the same
reasoning `host_still_reachable` already applies on the reboot arm. `all`, not
`any`: a run mixing a lost host with a genuine check failure still rolls back,
on behalf of the host the rollback can repair.

## 2. A host lost to its reboot lost its history row

#385 moved `add_op_history` after the fan-out so a run that never started would
stop claiming it had — right, but it landed the write after the **reboot** too.
A transactional host that never comes back cannot be written to, so
`Target::add_history` hit a dead link and swallowed the failure at WARN. The
row went missing on precisely the host whose state most needs reconstructing,
for `install`, `uninstall`, `update` and `downgrade` alike.

It now lands between dispatch and reboot. #385's invariant holds — a run that
never started still writes nothing — and the row count and format are
unchanged, which matters because `/var/log/mtui.log` is parsed by other tools.
install/uninstall had no seam (their reboot is inside `Operation::run`), hence
`OperationGroup::add_history`; writing before `Operation::run` instead would
have skipped the `update_lock` gate and left a row for a run a foreign lock
stopped. The label comes from a new `Operation::history_label` returning the
command verb, deliberately **not** from `role()`, which yields
`"installer"`/`"uninstaller"`.

`MockConnection::sftp_append` now reconnects at entry like the real
`SshConnection::sftp()`. Without it the bug was not reproducible at all: the
mock accepted an append on a host that had gone away, so the regression tests
would have passed with the write on **either** side of the reboot.

## 3. `prepare`/`downgrade` rebooted into failed transactions

Both collected a host's failure and then handed `reboot_transactional` the full
map anyway, activating the snapshot the failure came from. The gate is
per-host, mirroring the install/uninstall template: a healthy host still
reboots so its own snapshot activates, and every skipped host is named at WARN.
Whole-group suppression is `update`'s shape because its outcome routes a
group-wide rollback; these two have no compensating group action.

The skip set never uses the stderr rule `host_command_failures` also applies —
`transactional-update` writes progress to stderr on a *successful* run, and
skipping such a host would leave a healthy snapshot inert. Exit codes are
usable here because, unlike update's script, these templates are single
commands.

A host with **nothing staged** is skipped too. It drops out of `combined` when
the probe resolved no versions — reachable with an exit-0 probe, so
`dead_probes` does not cover it. Rebooting it is not merely gratuitous: when
the downgrade *is* the update's rollback, the host would boot into the snapshot
the failed update left staged, undoing the update flow's own suppression.

`prepare --installed-only` now judges **every** package, not just the last. It
runs one fan-out per package but read a single post-loop snapshot, and its
template exits 0 when a package is absent — so a later no-op masked an earlier
failure in both the verdict and the reboot decision. Per-host failures are
deduped: a second entry for one host would push `aggregate_failures` out of its
verbatim branch into the summary, dropping `host` to `None`.

## Verification

Full gate green: `fmt`, `clippy --workspace --all-targets -D warnings`,
`test --workspace`, `test -p mtui-mcp -F mcp`, `--no-default-features` and
`--all-features` compile-only, `typos`. CHANGELOG diff is purely additive.

**14 mutants, each killed by exactly the intended test** — including "delete
the gate", "gate on stderr too", "route `-1` into the rollback again", "write
the history row after the reboot again", "label from `role()`", and "drop the
mock's reconnect guard".

Worth stating plainly: the review found that my own headline test for §1 was
**vacuous**. Its `with_default` fixture answered the repo-refresh command with
the same marker, so it passed with the patch dispatch deleted entirely. It is
rewritten to script the marker onto the patch alone plus assert the patch was
dispatched, and that mutant now dies.

## Deliberately out of scope

- The update template still masks its patch command's exit code. Fixing it
  means editing a remote command string that lands in `show_log` transcripts,
  and it would activate the currently-dead `104` branch so a zypper host could
  newly fail an update *and* fire the group-wide rollback. Own PR, wants
  sign-off.
- `prepare`/`install` still have no check for the `slmicro`/`YUM` keys — the
  gate here leans on the exit code, which is sound because those templates are
  single commands.
- `("slmicro", true)` downgrade check still gates only `-1` by design; the new
  reboot gate covers the plain non-zero case for the reboot decision.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,19 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
 
 ### Fixed
 
+- `update` now reaches a verdict on SL Micro and RHEL/YUM hosts. Both have had
+  an update command since the port but no post-update *check*, so the flow
+  skipped them silently and reported success however the patch went — a locked
+  update stack, an unresolved dependency or an RPM error on those hosts was
+  simply not looked at, and on SL Micro the host was then rebooted into the
+  failed transaction. SL Micro is judged on the command's output markers rather
+  than its exit code: its update template ends with a repo-cleanup loop, so the
+  status the shell reports is that loop's and not the patch's.
+- An `update` that timed out, or whose connection dropped part-way, is no
+  longer reported as successful on **any** host. The update check had no
+  equivalent of the downgrade check's "timed out or failed to run" gate, so a
+  patch that never completed left an empty transcript that matched none of the
+  failure markers and passed.
 - A post-operation reboot whose SSH link had gone idle during the (multi-
   minute) package transaction now reconnects before dispatching, as ordinary
   commands already did. Without it the reboot failed to dispatch against a
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,58 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
 
 ## [Unreleased]
 
+### Fixed
+
+- `update` now reaches a verdict on SL Micro and RHEL/YUM hosts. Both have had
+  an update command since the port but no post-update *check*, so the flow
+  skipped them silently and reported success however the patch went — a locked
+  update stack, an unresolved dependency or an RPM error on those hosts was
+  simply not looked at, and on SL Micro the host was then rebooted into the
+  failed transaction. As on every other host, a failed update check on these
+  also drives the group-wide rollback downgrade, which they could not reach
+  before.
+  SL Micro is judged on the command's output markers rather than its exit code:
+  its update template ends with a repo-cleanup loop, so the status the shell
+  reports is that loop's and not the patch's. RHEL/YUM is judged only on
+  whether the command ran: that key covers every RHEL version, `yum` is `dnf`
+  on RHEL 8/9, and mtui passes the whole package list to hosts that carry only
+  a subset — so neither a non-zero exit nor an `Error:` line (the template also
+  runs `yum repolist` in the same shell) reliably means the update failed.
+- An `update` that timed out, or whose connection dropped part-way, is no
+  longer reported as successful on **any** host. The update check had no
+  equivalent of the downgrade check's "timed out or failed to run" gate, so a
+  patch that never completed left an empty transcript that matched none of the
+  failure markers and passed. Such a host fails the command but does **not**
+  trigger the rollback downgrade: the flow could not talk to it, so a
+  group-wide downgrade could not repair it and would only revert the hosts that
+  patched cleanly. A run mixing one lost host with a genuine check failure
+  still rolls back, on behalf of the host the rollback can actually repair.
+- A transactional host is no longer rebooted by `prepare` or `downgrade` when
+  nothing was staged on it — no package resolved to a version to downgrade, or
+  no prepare command was dispatched. The reboot was gratuitous on its own, and
+  when the downgrade is the `update` rollback it activated whatever the failed
+  update had staged, undoing the update flow's own decision not to reboot.
+- `prepare --installed-only` reports a package that failed anywhere in its
+  per-package run, not only in the last one. It installs one package per
+  round-trip but read a single post-run snapshot, and its template exits 0 when
+  a package is absent — so a later no-op masked an earlier failure, and the
+  host was rebooted into it.
+- A transactional host lost to its post-operation reboot keeps its
+  `/var/log/mtui.log` row. The row was written after the reboot, so on the one
+  host an operator most needs to reconstruct — the one that never came back —
+  the append failed and was swallowed at WARN, leaving no record that
+  `install`, `uninstall`, `update` or `downgrade` had run there at all. It is
+  now written between the command dispatch and the reboot: still never for a
+  run that did not start, and still exactly one row per operation.
+- `prepare` and `downgrade` no longer reboot a transactional host whose
+  package transaction failed. Both flows collected the failure and then
+  rebooted every transactional host anyway, activating the very snapshot the
+  failure came from. The gate is per-host, matching `install`/`uninstall`: a
+  healthy host in the same group still reboots so its staged snapshot
+  activates, and every skipped host is named in the log. A failed combined
+  transactional downgrade is now also reported as a per-host failure rather
+  than relying solely on the post-rollback version verdict.
+
 ## [26.1.1] - 2026-07-27
 
 ### Added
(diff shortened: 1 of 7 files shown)


plusky:
Restructured into four commits instead of one — the three fixes are separable and should be bisectable:

| commit | |
|---|---|
| `cf2bfe5c` | `update` had no verdict at all on SL Micro/YUM |
| `d72bffa6` | `prepare`/`downgrade` rebooted into failed transactions |
| `c44765cd` | a host lost to its reboot lost its `/var/log/mtui.log` row |
| `37bd430b` | the adversarial review of the three above |

The fourth is deliberately its own commit rather than being folded back in: three of four reviewers independently found that the first commit's new `-1` gate routed an *unreachable* host into the group-wide rollback, and that story is worth reading separately from the fix it corrects.

Each of the four builds and passes `cargo test --workspace` on its own — verified individually, not assumed. The final tree is byte-identical to the one CI already went green on, so this is a history reshape with no content change.

mimi1vx (commented):
## Review

I checked this out locally, ran the gate, and independently re-ran the mutation argument rather than taking the PR description's word for it. The three fixes are real bugs, the fixes land at the right layer, and — unusually — the tests are actually load-bearing.

### Verified locally (`37bd430`)

`cargo fmt --all --check`, `cargo clippy --workspace --all-targets -D warnings`, `cargo test --workspace` — all clean. Upstream CI is green on every job including `codecov/patch` and the feature matrix.

Mutants I applied myself and watched die (this is the part I care about, given how many green suites this repo has been burned by):

| mutation | outcome |
|---|---|
| move `group.add_history(...)` back to after `group.reboot(reboot)` in `Operation::run` | `install_records_history_for_a_host_lost_to_its_reboot` **FAILED** |
| collapse the `never_ran` routing to always `UpdateFailure::Check` | `update_timeout_does_not_roll_back_healthy_hosts` **FAILED** |
| delete the whole prepare reboot-gate closure body (`return true`) | 3 tests **FAILED** (`..._skips_the_reboot_of_a_host_whose_prepare_failed`, `..._judges_every_package_not_just_the_last`, `..._does_not_reboot_a_host_with_nothing_staged`) |
| delete the `!combined.contains_key` + `failed_downgrade` arms of the downgrade gate | 2 tests **FAILED** |
| unregister `("slmicro", true)` from `update_check` | 5 tests **FAILED**, including the end-to-end `perform_update_fails_a_transactional_host_whose_stack_is_locked` |

So the claim holds. Notably `perform_prepare_still_reboots_a_host_that_only_wrote_to_stderr` and the "clean combined downgrade still reboots" assertion survive all of the above — the gates are pinned in both directions, which is the failure mode that would otherwise leave a healthy snapshot inert.

Also confirmed independently:

- `Target::add_history` does **not** push to `self.out`, so moving the write between the fan-out and `run_checks` in `update_run_phase` cannot clobber the `last*` snapshot the checks read. This was my first worry and it's clean.
- `MockConnection::sftp_append` now genuinely mirrors `SshConnection::sftp()` (`ssh.rs:479-482`). Making the mock *less* permissive to disarm a vacuous test is the right instinct, and putting a test for the guard in `mtui-hosts` so its removal fails locally rather than silently in `mtui-testreport` is better than the usual.
- The `/var/log/mtui.log` row format is byte-identical: `Operation::run` builds `[history_label, packages.join(" ")]`, which is exactly what `add_op_history(.., None, ..)` produced. `install_writes_no_history_when_no_doer_resolves` still pins #385's invariant.
- `UpdateFailure` has no consumers outside `update_flow.rs`, so the new variant is contained.

### One design question worth answering before merge

**`NotRun` swallows the rollback for a host that is still perfectly reachable.** `Target::run` records `-1` for a `HostError::Timeout` on a live host just as it does for a lost connection (`target/mod.rs:1013-1028`). The reboot arm's analogue is genuinely reachability-based (`host_still_reachable`); this one is not — it infers "we lost the host" from a sentinel that also means "zypper took longer than the timeout on a host that is right there". That host is the *most* likely to be half-patched, i.e. exactly the split-brain the group-wide rollback exists to undo, and it's now routed away from it.

There is a decent counter-argument the PR doesn't make: after a client-side timeout the remote transaction is probably *still running*, so firing a downgrade into it would collide with the zypper lock and make things worse. If that's the reasoning, say so in the doc comment on `never_ran` — it's a stronger justification than the reachability analogy, and it's the thing a future reader will need. If it isn't, consider narrowing `NotRun` to hosts that fail a liveness probe, the way the reboot arm does.

Either way this is strictly better than the status quo (where `-1` produced no verdict at all), so I don't think it blocks.

### Minor

1. `update_run_phase`'s own doc comment (`update_flow.rs:1391-1396`) still enumerates only `Check` / `Reboot` / `RebootNotTaken`. It's the one place a reader looks for the routing table and it's now incomplete.
2. `transactional_update` inherits `markers`' `stderr.contains("Error:")` rule, and `SLM_UPDATE` is a four-command script whose stderr also carries `zypper -n lr -puU` and the `zypper -n rr` cleanup loop — i.e. precisely the "the marker can come from a sibling command in the same `exec`" objection used to *refuse* that rule for YUM. It's consistent with what the zypper key has always done, so I'd leave the behaviour alone, but the asymmetry in the two doc comments reads as an oversight rather than a decision. A sentence in `transactional_update`'s doc naming the residual risk would square them.
3. `perform_operation` (`update_flow.rs:468-470`) now carries a three-line comment with no statement under it. The information is worth keeping — it's a "why not here" — but it currently reads like the body was deleted and the comment forgotten.
4. The PR description opens "Three silent failures ... One commit", but the branch carries four. Worth fixing the description or squashing, so the merge commit's rationale matches what actually landed.

Nothing here is a correctness objection. The gating logic is right, the tests can fail, and the interop contracts (`/var/log/mtui.log` row shape and count, lock format, MCP surface) are untouched.

plusky:
Thanks for re-running the mutants yourself rather than taking the table's word for it — given this repo's history with green suites, that's the only review of the claim that counts.

All four points addressed in `be31aa42` (doc-only, behaviour untouched):

**The design question — documented, not narrowed, and here's why.** Your counter-argument is the actual reasoning, and it's now stated where the routing lives (`UpdateFailure::NotRun`'s doc plus the `never_ran` comment): the timeout closes the SSH channel, which asks the remote side to reclaim the process — but rpm masks signals inside its transaction, so the patch may have died, finished after the flow stopped watching, or still be holding the package-manager lock. Firing the group-wide downgrade there dispatches a second transaction at the one host whose first was never observed to end, and reverts every healthy host to do it.

I considered narrowing to a liveness probe instead and think it's the wrong shape here: a probe would route the timed-out-but-live host back into `Check` — which is precisely the collision case above. Reachability answers "can the rollback reach this host"; it can't answer "is it safe to fire a transaction at it", and for a timeout those two questions disagree. The docs no longer borrow the reboot arm's reachability analogy — you're right that it was inexact, since that arm probes and this one infers from a sentinel.

**Minors:** `update_run_phase`'s routing table now includes `NotRun`; `transactional_update`'s doc names the sibling-command `Error:` exposure as a decision (the transcript is all zypper — the vocabulary the markers were written against — where on yum the same rule would judge a transcript it was never written for); the `Operation::run` history note now reads as the deliberate "why not here" it is; and the PR description matches the branch.

mimi1vx (approved):

Issue #415, opened 2026-07-30
updates lacks osc-qam's field selection (-F) and repeatable -G review groups


plusky:
Reported by a user comparing `mtui`'s `updates` with `osc qam list`:

> MTUI (which directly uses the OBS request API) and osc-plugin-qam are not equivalent in functionality. MTUI's `updates` command lacks field functionality. For example, the command
>
> ```
> osc qam list -G qam-sle -G qam-emergency -G qam-teradata \
>   -F ReviewRequestID -F SRCRPMs -F Rating -F Products -F "Incident Priority" \
>   -F Bugs -F "Unassigned Roles" -F "Assigned Roles" -F Package-Streams \
>   -F Creator -F Issues -F Comments
> ```
>
> Can we complete these fields in `mtui -> updates`?

The gap is real. Two separate gaps, in fact — field selection *and* repeatable `-G` — plus one correction to the premise that changes how expensive the fix is.

## Current behaviour

`crates/mtui-core/src/commands/updates.rs` renders a fixed-width row with **six** hardcoded fields (`render_row`, :203-219):

```rust
let mut row = format!(
    "  prio={:<5} {:<10} {:<12} {:<11} {}",
    field("priority"),
    field("status"),
    field("kind"),
    deadline,
    field("id"),
);
if want_assignment {
    ... row.push_str(&format!(" assignee={assignee}"));
}
```

There is no `-F`/`--field` equivalent, no way to add a column, and no machine-readable output.

`--review-group` is also **single-valued** — no `ArgAction::Append` — so the reporter's `-G qam-sle -G qam-emergency -G qam-teradata` has no equivalent either. That one is small and independent of the field question.

## Correction to the premise, because it drives the design

`updates` does **not** use the OBS request API. It queries **TeReGen**: `teregen.updates(&query)` (`crates/mtui-datasources/src/teregen.rs:235-259`), which returns the response's `updates` key as an opaque `serde_json::Value` array.

mtui does use OBS directly elsewhere, but `crates/mtui-datasources/src/obs/qam.rs` exposes only *mutating* QAM operations — `comment`, `assign`, `unassign`, `approve`, `reject`. **There is no per-request field reader on the OBS side.** So mtui has no existing path to fetch request-level attributes for a queue listing the way osc-plugin-qam does.

This matters: the two possible implementations differ by an order of magnitude in cost.

## The first thing to determine

[...]

PR #420, opened 2026-08-02
feat: give updates osc-qam field selection (-F), repeatable -G and --json

plusky:
## Why

#415 reports two gaps against `osc qam list`: no field selection and no way to pass several review groups. Investigation against live TeReGen (2026-08-02) showed the cheap half is real: five of the reporter's twelve fields are already in the `/updates` payload and silently discarded — `rating` among them — and `assignees`/`review_groups` are enough to derive both role fields. This PR lands that half; the seven genuinely missing fields need TeReGen enrichment, asked for in [qa-maintenance/teregen#1](https://gitlab.suse.de/qa-maintenance/teregen/-/issues/1).

## What changed

- **`-F/--field` (repeatable)** — osc-qam field names, case-insensitive with spaces/hyphens/underscores equivalent, rendered as one `Name: value` block per update in the requested order. Serves: `ReviewRequestID`, `Incident Priority`, `Rating`, `Category`, `Status`, `Kind`, `Deadline`, `Assignee`, `Assigned Roles`, `Unassigned Roles`, `Title`, `URL`. The unserved seven (`Products`, `SRCRPMs`, `Bugs`, `Package-Streams`, `Creator`, `Issues`, `Comments`) are named in an error citing the gap rather than rendering empty. `Assignee`/`Assigned Roles` are refused under `--status all` — TeReGen omits assignment data there on SLFO rows, and printing "unassigned" for an assigned row is false data (forcing `with_assignment` is no fix: it implies `status=testing` server-side).
- **`-G/--review-group` repeatable** — groups OR together as one server query per group: a repeated `review_group` param is not OR-ed by TeReGen (last wins), and the filter cannot be replicated client-side because the server matches SLFO rows against a group without ever serialising `review_groups` on them. Merged priority-descending (stable), duplicate ids collapsed, duplicate input groups collapse to one query.
- **`--json`** — the raw rows as a JSON array (each row whole, nothing re-rendered); `[]` on an empty queue; conflicts with `-F`. Follows `list_refhosts --json`.
- **`fix(mcp)` (first commit, pre-existing bug):** the MCP argv reconstruction emitted an Append flag once for a whole list (`--group a b`), which clap rejects for every default-`num_args` Append arg — `approve` with two groups failed to parse, and `run` with two hosts leaked the second host into the command. The flag is now repeated per element, the only form every Append arg parses.
- **MCP schema note:** `updates.review_group` changes string → array (breaking for schema-validating clients; a string still dispatches — scalars render as one-element lists), plus new `field`/`json` properties. Snapshot updated.
- Docs: FAQ example block + field list; `cli.md` regenerated; CHANGELOG entries in Added and Fixed.

## Verification

- `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`
- `cargo test --workspace` (29 `updates` tests, incl. 13 new) and `cargo test -p mtui-mcp -F mcp` (191+48, incl. 3 new argv round-trips through the real dispatch parser)
- Feature matrix compile-only: `--no-default-features` / `--all-features`
- Mutation-checked by hand-breaking and observing red: dedup disabled, merge sort removed, Assigned/Unassigned extractors swapped, render-all-ignoring-`-F`, assignment-refusal deleted; the argv fix's new test was observed red against the unfixed builder
- Adversarial review before commit by four independent read-only reviewers (correctness, test-vacuity, MCP-surface, docs-contract); all confirmed findings addressed — the `--status all` refusal and the empty-assignee-group handling came out of it
- Server behavior (last-wins repeated param, SLFO rows matching a group filter without serialising `review_groups`, priority-descending order, `with_assignment` presence effects) verified against live qam.suse.de, not assumed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,10 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
 
 ### Fixed
 
+- MCP calls passing a list to a repeatable single-value flag (e.g. `approve`
+  with two `group`s) no longer fail to parse: the reconstructed argv now
+  repeats the flag before every element instead of emitting it once for the
+  whole list.
 - `approve` no longer approves a Gitea update whose PR head differs from the
   checked-out testreport without saying so. Interactively it now logs the
   expected/actual hash pair and asks for confirmation (default no); a missing
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,24 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
 
 ## [Unreleased]
 
+### Added
+
+- `updates` gained osc-qam-style field selection and a scripting output
+  (#415): `-F/--field` (repeatable) selects output fields by the names
+  `osc qam list -F` uses (`Rating`, `Assigned Roles`, `Unassigned Roles`,
+  `Incident Priority`, … — case-insensitive, with spaces/hyphens/underscores
+  equivalent), and `--json` (not combinable with `-F`) prints the raw TeReGen
+  rows as a JSON array. Fields the TeReGen queue listing does not carry
+  (`Products`, `SRCRPMs`, `Bugs`, `Package-Streams`, `Creator`, `Issues`,
+  `Comments`) are named in an error instead of rendering empty, and
+  `Assignee`/`Assigned Roles` are refused under `--status all`, where TeReGen
+  omits the assignment data they would render.
+- `updates --review-group` gained the short form `-G` and is now repeatable;
+  multiple groups are OR-ed via one TeReGen query per group, merged
+  priority-descending with duplicates collapsed. **MCP schema note:** the
+  `review_group` property of the `updates` tool changed from a string to an
+  array of strings.
+
 ### Changed
 
 - `refhosts.yml` rows now resolve YAML merge keys (`<<: *anchor`) instead of
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,9 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   (`Products`, `SRCRPMs`, `Bugs`, `Package-Streams`, `Creator`, `Issues`,
   `Comments`) are named in an error instead of rendering empty, and
   `Assignee`/`Assigned Roles` are refused under `--status all`, where TeReGen
-  omits the assignment data they would render.
+  omits the assignment data they would render; `Unassigned Roles` degrades to
+  `n/a` on any row carrying no assignment data rather than listing every
+  group as open.
 - `updates --review-group` gained the short form `-G` and is now repeatable;
   multiple groups are OR-ed via one TeReGen query per group, merged
   priority-descending with duplicates collapsed. **MCP schema note:** the
--- a/crates/mtui-core/src/commands/updates.rs
+++ b/crates/mtui-core/src/commands/updates.rs
@@ -1,8 +1,12 @@
 //! The `updates` command — list the update queue via the TeReGen API.
 
+use std::collections::HashSet;
+
 use async_trait::async_trait;
 use clap::{Arg, ArgAction, ArgGroup, ArgMatches};
+use futures::future::join_all;
 use mtui_datasources::UpdatesQuery;
+use serde_json::Value;
 
 use crate::command::{Command, Scope};
 use crate::commands::apicall::teregen_client;
@@ -19,8 +23,9 @@ const STATUS_ALL: &str = "all";
 /// that are **in testing**.
 /// `--assignee`/`--mine`/`--all-assignees` pick another assignment view (and
 /// drop the unassigned default); `--status all` widens to every status. This is
-/// a session-global query, so it runs exactly once ([`Scope::Single`]) rather
-/// than fanning out per template.
+/// a session-global query, so it runs once per invocation ([`Scope::Single`])
+/// rather than fanning out per template — though it may issue one TeReGen
+/// query per `--review-group`.
 pub struct Updates;
 
 #[async_trait]
@@ -42,11 +47,38 @@ impl Command for Updates {
     fn configure(&self, cmd: clap::Command) -> clap::Command {
         cmd.arg(
             Arg::new("review_group")
+                .short('G')
                 .long("review-group")
                 .value_name("GROUP")
+                .action(ArgAction::Append)
                 .help(
                     "filter by review group as the bare group name, e.g. qam-sle \
-                     (not the '<group>-review' login form, which classic rows lack)",
+                     (not the '<group>-review' login form, which classic rows lack); \
+                     repeatable — groups are OR-ed (one server query per group)",
+                ),
+        )
+        .arg(
+            Arg::new("field")
+                .short('F')
+                .long("field")
+                .value_name("FIELD")
+                .action(ArgAction::Append)
+                .help(
+                    "select output fields by osc-qam name (e.g. -F Rating \
+                     -F 'Assigned Roles'); repeatable, rendered as one block per \
+                     update; names are case-insensitive and treat spaces, \
+                     hyphens and underscores as equivalent",
+                ),
+        )
+        .arg(
+            Arg::new("json")
+                .long("json")
+                .action(ArgAction::SetTrue)
+                .conflicts_with("field")
+                .help(
+                    "print the raw TeReGen rows as a JSON array (each row \
+                     emitted whole, unlike -F; honours --limit; not combinable \
+                     with -F); an empty queue prints []",
                 ),
         )
         .arg(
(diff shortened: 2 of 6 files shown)


mimi1vx (approved):

mimi1vx (changes requested):
Solid PR. The design reasoning is unusually well-evidenced — verifying the server's last-wins/`with_assignment` behaviour against live qam.suse.de rather than assuming it is the right call, and the `fix(mcp)` commit is a genuinely important catch. Requesting changes on one correctness question and three cheap test/UX gaps; everything else below is non-blocking.

I verified the points below by running them, not by reading.

## Gate — reproduced locally

`fmt` / `clippy -D warnings` / `cargo test --workspace` / `cargo test -p mtui-mcp -F mcp` (191+48) / feature matrix `--no-default-features` + `--all-features`: all green. `cargo run -p xtask -- gen-docs` produces **no diff**, so `docs/src/cli.md` is genuinely regenerated.

`RUSTDOCFLAGS="-D warnings" cargo doc …` fails on the head alone, but only because the head is 11 commits behind `main`, where the rustdoc fixes and `41dc64f0 ci: gate CI on rustdoc warnings` landed after the merge-base. Merging `main` in: clean merge, rustdoc clean, full suite still green. Non-issue — CI tests the merge and is green. Worth a rebase before merge anyway.

---

## 1. `Unassigned Roles` is exempt from the very gate this PR argues for

`crates/mtui-core/src/commands/updates.rs:433-438` marks it `needs_assignment: false`, but `extract_unassigned_roles` derives its answer from `assignees` — the key TeReGen omits without `with_assignment`. I ran it: a classic Maintenance row served under `--status all` with `review_groups` but no `assignees` prints

```
Update queue (1):
  Unassigned Roles: qam-sle, qam-openqa
```

with no way to tell it apart from the truth. That is exactly the false-data failure mode the PR refuses for `Assignee` / `Assigned Roles`.

It may still be correct — the `needs_assignment` doc says the omission is *on SLFO rows*, and SLFO rows lack `review_groups` so they take the `n/a` path. If so the two conditions are complementary and the field is safe. But that is an unstated, untested invariant carrying the field's whole correctness, and it is the one claim in the PR I can't check against the live server. Either state it on the `Unassigned Roles` `FieldSpec` and pin it with a `--status all` classic-row test, or flip it to `needs_assignment: true`.

## 2. The `run` case of the argv fix is untested — and it's the silent one

`approve --group a --group b` fails *loudly* pre-fix. `run` does not. Pre-fix, `{hosts: ["h1","h2"], command: ["uname","-a"]}` reconstructs as

```
["--target", "h1", "h2", "uname", "-a"]
```

which clap parses as `hosts=["h1"]`, `command=["h2","uname","-a"]` — mtui runs `h2 uname -a` on h1. Post-fix it's `--target h1 --target h2 uname -a` → `hosts=["h1","h2"]`, `command=["uname","-a"]`. Both observed by hand-reverting the loop.

That is the highest-consequence instance of the bug and the only one that cannot announce itself, and it has no test. Please add the round-trip (multi-value flag *plus* a trailing positional tail) next to `append_single_value_flag_repeats_per_element` — the existing tests all cover flag-only commands, so none of them pins the positional-tail interaction the `run` bug lived in.

## 3. The completion additions cannot fail

`complete()` gained `--field`/`--json` (`updates.rs:135-136`). I deleted both lines: all 29 `updates` tests stay green. `complete_offers_static_flags` is two entries short.

## 4. Completion doesn't know about `-G`/`-F`, and treats a now-repeatable flag as once-only

- The synonym groups are still `&["--review-group"]` / `&["--field"]`, so the new short forms are never offered and aren't registered as synonyms (cf. `hostslock`'s `["-c", "--comment"]`).
- `complete_choices` drops a synonym group once it appears on the line. Observed: `Updates.complete(.., "--r", "updates --review-group qam-sle --r")` → `[]`. The flag became repeatable in this PR and you can't tab-complete the second one — routing it through `extra` (never suppressed) is the cheapest fix.
- `-F` doesn't complete field names, though `FIELDS` is right there and `complete_choices` already takes an `extra` list. Twelve osc-qam names, several with embedded spaces, are exactly what users will fumble.

---

## Non-blocking

- **~80 lines duplicated** between the `review_groups.len() > 1` and single-group branches (`updates.rs:226-305`) — error mapping and malformed-array handling appear twice each. Collapsing to a `Vec<Option<&str>>` of groups (empty → `vec![None]`), one loop, dedup+sort gated on `len() > 1`, halves it without changing behaviour.
- **`is_multi` is now broader than the repeat-per-element rule is safe for** (`crates/mtui-mcp/src/argv.rs:136`): true for `Append` *or* `num_args().max_values() > 1`. Repeating is right for `Append`; for a future `ArgAction::Set` + `num_args(1..)` option the last occurrence would win and silently drop earlier values. Nothing hits it today — every multi-value *option* in the registry is `Append`, and `config attributes` / `run command` are positionals taking the earlier branch — but it's one arg away. Consider gating the repeat on `Append` and keeping flag-once for the `num_args` case.
- **Merge sort keys `priority` via `as_i64`**, non-numeric → `i64::MIN`. `render_row` stringifies priority defensively, implying it isn't guaranteed numeric. If it is, the defensive path elsewhere is dead; if it isn't, cross-group ordering collapses to group order.
- **Two sentinels for absent data**: `render_row` prints `?`, `-F` prints `-` (and `n/a`). Deliberate?
- **`-F Deadline`** emits the full ISO timestamp; `render_row` truncates to the date.
- `join_all` fans out unbounded, one concurrent request per `-G`. Fine for a handful, unbounded if a script passes thirty.

---

## What I confirmed is well-tested

Dedup-by-id, merge order across groups, `--limit` slicing *after* the merge-sort, partial-failure-is-an-error-not-partial-output, malformed `updates` on both paths, id-less rows kept, missing-`updates`-key skipped, `--json` preserving unknown keys and numeric types, empty-entry-list assignee group. The fixtures can express the mutations they claim to catch. `--json` matches `list_refhosts --json` (pretty array, emitted before the empty check), and the CHANGELOG covers both commits and flags the MCP schema break.

plusky:
All four blocking items addressed in `af0a9a4f`, on top of the rebase onto current main you suggested (rustdoc gate replicated locally with CI's exact invocation: clean). Point by point:

**1. `Unassigned Roles`** — you were right that an unstated invariant was carrying the field. I probed the invariant first: under the all-statuses view (2169 rows), **only `(Maintenance, testing)` rows carry *any* of `assignee`/`assignees`/`review_groups`** — the two keys travel together on every row today, so your hazard row doesn't occur live. But the coupling is incidental, so rather than document it, the extractor now enforces its own precondition: **`assignees` key absent → `n/a`**, same as the SLFO arm. Your exact scenario (classic row, `review_groups`, no `assignees`, `--status all`) is pinned by `unassigned_roles_without_assignment_data_is_na_not_all_open`, and the mutation (defaulting to an empty map, the pre-fix behaviour) was observed red. The field stays *allowed* under `--status all` precisely because it now degrades honestly per row — refusing it like `Assignee` would have thrown away the rows where it can still answer.

**2. `run` round-trip** — added `run_hosts_and_command_round_trip_without_leaking_across_boundary`: exact token vector plus re-parsed `hosts`/`command` value lists, through the same `command_parser` real dispatch uses. Observed red against the flag-once emission with precisely the corruption you described — `h2` bare, clap reading it as the start of the command.

**3./4. Completion** — `complete_offers_static_flags` now lists `-G`/`-F`/`--field`/`--json` (deleting the entries goes red). `-G`/`-F` ride in `extra` as you suggested, so your `--review-group … --r → []` case now completes (pinned by `complete_still_offers_repeatable_flags_after_first_use`); `--json` stays suppressed after use like any once-only flag. And `-F` completes the twelve field names from the registry (`complete_offers_field_names_after_field_flag`, observed red under a disabled match).

**Non-blocking, taken:** the ~80-line duplication collapsed along the lines you sketched (one loop over `Vec<Option<&str>>`, dedup+sort gated on multi — behaviour unchanged, all previous tests untouched and green); the fan-out is bounded (`buffered(4)`, order-preserving); and `is_multi`'s repeat-per-element is now gated on `Append` — a non-Append multi-value option keeps flag-once, where repeating would let the last occurrence win (pinned with a bespoke-parser test, since no registry arg has that shape).

**Non-blocking, deliberately not taken:** the `?` vs `-` sentinel split and `-F Deadline`'s full timestamp are both intentional — `render_row`'s `?`/date-truncation is the existing six-column contract left untouched, while the `-F` block view has no width pressure, so it shows full values and uses `-` for absent ones. `priority` via `as_i64` keeps the defensive fallback for the same reason `render_row` stringifies it: the queue payload is an external contract we read, not one we define.

One heads-up from the toolchain: clippy 1.97 (stable as of last month) fails `-F mcp` on two pre-existing `unwrap_or_else(CancellationToken::new)` in `mtui-mcp/src/session.rs:796/:835` (`clippy::unwrap_or_default`). CI's lint job doesn't compile the `mcp` feature so it can't fire there, and it's not this PR's code — flagging it so it doesn't surprise anyone running `clippy -F mcp` locally.

Full gate re-run on the head: fmt / clippy workspace / `cargo test --workspace` / `-p mtui-mcp -F mcp` (193+48) / feature matrix / rustdoc (CI invocation) / CHANGELOG additive vs main.

mimi1vx (approved):

PR #435, opened 2026-08-07
fix: decide old-vs-new per update, not from the RRID's shape (#433)

### LINKED ISSUE #433: SL-Micro 6.0/6.1 cutover: decide the old-vs-new workflow per update, not from the RRID's shape
## The premise

SL-Micro 6.0 and 6.1 are `SUSE:SLFO:1.1:*` — stated authoritatively in TeReGen's
`app/lib/TeReGen/Helpers.pm:15-18`, which maps each RRID shape to its product
*and its source*:

```
# - SUSE:Maintenance:123:1234 -> maintenance update for SLE-12/15 (from build service)
# - SUSE:PI:16.0:387765 -> product increment request for SLE-16 (from build service)
# - SUSE:SLFO:1.1:1234 -> update for SL-Micro 6.0/6.1 (from build service)
# - SUSE:SLFO:1.2:1234 -> update for SLFO, i.e. SLE-16 at this point (from gitea)
```

**For an interim period, `SLFO:1.1` updates will be served by BOTH the git and
the OBS workflow at the same time.** Two updates with the same RRID shape, the
same kind, and the same maintenance id will need opposite handling: different
update-repository URL layout, different review backend, different pre-approval
checks.

That single fact is what this issue is about. Every place mtui decides
old-vs-new today keys on the RRID's shape, and **no rule over the RRID can ever
be correct once both workflows share the `1.1` id space** — not
`maintenance_id != "1.1"`, not `kind`, not any refinement of either. The
discriminator has to become a per-update fact.

Note this is not a flag day we can wait out: during the interim, in-flight
classic `1.1` updates and new git `1.1` updates coexist, so flipping the
literals is as wrong as leaving them.

## `kind` cannot answer it either

The obvious candidate is TeReGen's `kind`, already shipped and already consumed
by `updates` as the `Kind` column (`crates/mtui-core/src/commands/updates.rs:420-423`,
`-F Kind`). It does not work, for two reasons:

- It is the RRID's own second component, which mtui already has offline — it
  adds no information.
- **kind already does not equal workflow, even today**: `SLFO:1.1` is
  `kind: SLFO` yet build-service-served. That is exactly why every predicate
  below carries a `1.1` epicycle, and why #396 exists against
  `SUSE:SLFO:1.1:418286`.

After the cutover starts, `1.1` and `1.2` are both `SLFO` and both partly
git-served, so `kind` separates nothing at all.

## What decides it today (all six sites)

| # | Site | Predicate | Controls |
|---|------|-----------|----------|
| 1 | `crates/mtui-core/src/commands/apicall.rs:34` (`is_gitea_workflow`) | `kind == Slfo && maintenance_id != "1.1"` | Gitea vs OSC backend for `assign`/`unassign`/`reject`/`comment`, and `approve` (`approve.rs:100`) |
| 2 | `crates/mtui-testreport/src/reports/sl.rs:94` (`update_repos_parser`) | `maintenance_id() == Some("1.1")` | `slrepoparse` (`<repo>/images/repo/<name>-<ver>-<arch>/`) vs `gitrepoparse` (`<repo>/standard`) |
| 3 | `crates/mtui-testreport/src/reports/sl.rs:169` (`check_hash`) | `maintenance_id() == Some("1.1")` | Returns `HashCheck::Ok` unconditionally, skipping the Gitea head comparison |
| 4 | `crates/mtui-datasources/src/qem_dashboard/incident.rs:78` | `kind == Slfo && maintenance_id == "1.2"` | Which id is used as the QEM incident number |
| 5 | `crates/mtui-datasources/src/openqa/base.rs:127` | `kind == Slfo` | openQA build prefix `git` vs `smelt` |
| 6 | `crates/mtui-datasources/src/obs/qam.rs:42` | `matches!(kind, Pi \| Slfo)` | Skips the maintenance testreport |

Sites 2 and 3 key on the id alone; kind is implied because only `SlReport`
reaches them. Sites 1-3 are the ones that become undecidable — they are exactly
the behaviours that differ between a git and an OBS `1.1`.

## What breaks, in order of blast radius

1. **Site 2 — wrong update repository.** The two parsers emit completely
   different URLs, so the wrong branch points every reference host at a repo
   that does not exist for that update; `set_repo` / `prepare` / `update` then
   operate on nothing. Worth checking whether #396 (prepare reports success
   without installing anything, on `SUSE:SLFO:1.1:418286`) is already a
   manifestation of this rather than a separate bug.
2. **Site 3 — the approve-time hash gate silently becomes a no-op.** For a
   git-served `1.1`, `check_hash` still short-circuits, so the check that the
   checked-out template matches the Gitea PR head is skipped for precisely the
   updates that need it. That is the check #398 restored, switched back off by
   an unrelated literal.
3. **Site 1 — review actions go to the wrong backend.** A git-workflow update
   assigned or approved through OBS (or the reverse) fails, or worse appears to
   succeed against the wrong system of record.
4. **Sites 4-6** need re-deciding explicitly rather than inheriting whatever
   falls out.

## A constraint on any fix: the class is chosen before the metadata is read

mtui picks the concrete report class from the RRID *before* it reads anything:
`crates/mtui-testreport/src/lifecycle.rs:119` `tr_factory(update, config)` →
`ObsReport`/`SlReport`/`PiReport`, and only then does `:134` `report.read()`
parse the template. The class is what parses the metadata, so **a metadata field
can never select the class**.

For this cutover that is survivable, and worth stating so the design does not
drift into assuming otherwise: **git and OBS `SLFO:1.1` updates resolve to the
same class** (`SlReport`), and all three sites that must differ (1, 2, 3) are
within-class branches evaluated after load. Anything that later needs a
*different* class must be decidable from the RRID or from an API call.

## The discriminator

**A. TeReGen `kind` as it stands** — ruled out above.

**B. The update's own `metadata.json`.** The envelope already carries
`gitea_pr`, `gitea_pr_api` and `gitea_commit_hash`
(`crates/mtui-testreport/src/metadata_parsers.rs:137-143`, applied `:204-206`),
and `check_hash`/`apicall` already consume `gitea_pr_api`. TeReGen writes those
fields at `95_ExportMetadata.pm:86`, today gated on `^SUSE:SLFO:1\.2` — so a
git-served update is self-describing and an OBS-served one is not. Per-update,
authoritative, available offline from the checked-out template, no round-trip.
The weakness is that it infers a workflow from the presence of a field rather
than reading a declared one.

**C. A declared field from TeReGen** — requested as
[qa-maintenance/teregen#2](https://gitlab.suse.de/qa-maintenance/teregen/-/issues/2).
After their feedback the ask is:

- on `/reports/{id}/metadata` — *not* on `/api/v1/updates`, since the field
  needs a complete template and queue rows may not have one (@jbaier_cz);
- **and in the exported `metadata.json`**, alongside `gitea_pr_api`, so mtui
  keeps an answer offline and on every post-load command without a round-trip;
- split into *class* (`obs`/`slfo`/`pi`) and **source** (build service vs
  gitea). A binary `git|smelt` is too thin (@osukup), and class alone does not
  separate a git `1.1` from an OBS `1.1` — source is the axis that does.

**D. Explicit `repositories`.** `sl.rs:91` already short-circuits to
`reporepoparse` when `repositories` is non-empty, bypassing the sl/git branch
entirely. TeReGen writes `repositories` and `binaries` for `1.2` and `PI` but
not `1.1` (`95_ExportMetadata.pm:86-99`). **If git-served `1.1` updates receive
`repositories`, the worst site resolves itself** — asked on teregen#2, and if
the answer is yes it is worth more than the declared field.

## Direction

C is the destination, B is what we build now so mtui is not blocked on their
release, and both converge on the same plumbing:

1. Resolve the workflow **once at template load**, store it as a single fact on
   the report, and make all six sites consume that fact instead of re-deriving
   it. This refactor is the bulk of the work and is identical either way.
2. Until the declared field ships, populate that fact from the `gitea_pr_api`
   inference; afterwards the declaration is consulted first and the inference
   becomes the fallback for templates generated before it.
3. Confirm D independently — it may retire site 2 on its own.

Two constraints:

- **Do not name it `Workflow`.** `mtui_types::Workflow` already means the
  *test-execution* workflow (`auto`/`manual`/`kernel`, `set_workflow`).
  Something like `UpdateSource { Obs, Git }` avoids the collision.
- **Unknown must refuse, not default.** Given #396, a wrong guess here silently
  operates on the wrong repository. If no signal answers, say so and stop.

An operator override (`--update-source git|obs`, or a config key) is cheap
insurance for the interim and removable once the cutover completes.

## Still open

1. Will git-served `SLFO:1.1` updates carry `gitea_pr_api`, and will they
   populate `repositories`? (teregen#2)
2. Is the source knowable at row-build time on the classic feed, so absence can
   keep meaning "predates the field" rather than "classic"? (teregen#2)
3. Can a single update ever be served both ways, or is it always one or the
   other? (teregen#2)
4. When the interim ends, is the OBS path for `1.1` retired, or kept?

## Related

- #396 — prepare/export on `SUSE:SLFO:1.1:418286`; possibly the same root cause
  as site 2, worth checking before fixing either.
- #397 — no fixture exercises a populated SLFO envelope; blocks validating B.
- #398 — made `approve` stop skipping the Gitea hash check; site 3 skips it
  again for these ids.
- The same hard-coded-literal pattern exists on the TeReGen side and moves with
  the cutover: `19_FetchGitea.pm:12` (`^SUSE:SLFO:1\.2:`), `20_FetchOBS.pm:55`
  (`^SUSE:SLFO:1\.1:`), `95_ExportMetadata.pm:86`, `Rendering.pm:48`, and
  `23_FetchOpenQA.pm:9` (`:smelt` vs `:git`, mirrored by `openqa/base.rs:127`).

mimi1vx:
Closes #433.

## Summary

Six sites decide git-vs-OBS handling from the RRID's shape (`kind`, and the
`1.1`/`1.2` maintenance-id literals). During the SL-Micro 6.0/6.1 cutover
both workflows share the `SLFO:1.1` id space, so no rule over the RRID can
be correct. `metadata.json` already carries `gitea_commit_hash`, and is
mandatory on every load, so a per-update fact is always resolvable at load
time with no round-trip.

This introduces `mtui_types::UpdateSource` (`Obs`/`Git`), resolved once at
load from the template's own `gitea_commit_hash` rather than inferred from
the RRID, and rewrites every site that used to key on the RRID's shape to
decide explicitly from it instead.

**This is a selection, not an observation.** During the OBS-1.1 → git-1.1
transition an update may be served *both* ways at once, and exactly one
workflow must be active. The rule is a precedence: Gitea metadata present
⇒ `Git`, otherwise `Obs`. On a dual-served update, `Git` wins by design and
mtui leaves that update's OBS review request untouched.

## Bugs found and fixed along the way

Researching the openQA/QEM half of the ecosystem (qem-bot/qem-dashboard,
not TeReGen) turned up three pre-existing bugs, all fixed here:

- **B1** — `QemIncident::incident_number` keyed the dashboard lookup on
  maintenance id for anything but `SLFO:1.2`, so an OBS-served `SLFO:1.1`
  made mtui GET `/api/incidents/1.1` — not an incident number. Fixed by
  keying on `kind == Slfo` alone (both git- and OBS-served SLFO share the
  review id).
- **B2** — `get_incident_name` took the plain shortest package name, where
  qem-bot's `sort_packages` demotes arch-suffixed and `-livepatch-` names
  first. Ported that ordering.
- **B3** — `OpenQABase::new`'s `build` query used the RRID's maintenance id
  as its middle component, where qem-bot writes the dashboard's own
  incident number — for SLFO those differ, so mtui's openQA lookup has
  never matched an SLFO job. `OpenQABase::new` now builds the whole `build`
  string from the fetched incident record.

A Product Increment is also connected to neither qem-dashboard nor openQA,
so the dashboard fetch is now skipped outright for PI instead of making a
request that could never succeed.

## Commits

1. `feat(types): add UpdateSource`
2. `fix(testreport): decide the SLFO repo layout and hash gate per update`
3. `fix(core): route review actions on the update's own source`
4. `fix(datasources): decide the QEM/openQA incident facts from the update's own source` (B1–B3 + PI skip; bundled because the signature changes and their call sites are too interleaved to split without breaking intermediate builds)
5. `test(testreport): add SLFO 1.1 dual-served end-to-end fixtures`
6. `docs, chore(changelog): record the openQA BUILD-string contract and the update-source decision`

## Testing

- `cargo fmt --all --check`
- `cargo test --workspace` (all crates, 0 failures)
- `cargo build --workspace --no-default-features` / `--all-features`
- Every new branch (B1/B2/B3, the SLFO-1.1 dispatch, the PI skip) has a test that was observed **red** against the unfixed code before landing.

## Not included

Filing B1–B3 as separate upstream issues for the record (originally step 13
of the plan) was skipped — the fixes and rationale are captured in code
comments and the CHANGELOG instead.


plusky (approved):
Approve.

Verified the qem-bot claims against master and all three match verbatim: BUILD is
`:{type}:{number}:{package}` (`types/submissions.py:236`), `default_submission_type
= "smelt"` (`config.py:117`), and the demoted suffixes plus `-livepatch-`
(`types/submission.py:30-35,59`).

The source x maintenance-id matrix in `update_repos_parser` and the zero-request
negative control in `check_hash_obs_at_1_1_issues_zero_requests` are the right
guards — both fail if anyone reverts this to an id-based rule.

I queried whether `gitea_commit_hash` alone was too weak a signal; withdrawn.
`query_gitea` dies on a non-success response (`Helpers/Generator.pm:149`), so a
failed Gitea fetch aborts generation outright — there is no template carrying the
PR fields but no hash, and keying on the hash is sound.

Issue #399, opened 2026-07-30
put does not expand globs, so 'put *.rpm' silently matches nothing — implement or reject?


plusky:
`put *.rpm` silently uploads nothing matching, because nothing in the Rust REPL expands shell globs — while the tool it replaced did expand them. This needs a decision, not necessarily a fix.

## State

`crates/mtui-core/src/commands/sftpput.rs:13-19`:

```rust
/// Uploads a local file (or directory tree) to every enabled reference host.
///
/// Files are placed under the
/// report's remote working directory (`metadata.target_wd(<name>)`); a directory
/// argument is walked and each contained file uploaded. Only enabled hosts are
/// contacted. Shell-glob expansion of the argument is a Phase-6 REPL concern;
/// here the argument is a concrete path (file or directory).
```

Phase 6 shipped. Nothing globs: the REPL is reedline + shlex, neither of which expands patterns, and there is no `glob` crate anywhere in the workspace (`grep glob` across every `Cargo.toml` returns nothing). So the comment defers to a phase that is over and to a component that never took the work.

The predecessor implementation did expand: `glob(self.args.filename[0])`. So `put *.rpm` used to work and now finds no file.

## The decision

1. **Implement globbing** in `put` (and `get`, if it has the same shape) — restores the prior behaviour; costs a dependency and needs a decision about whether a pattern matching nothing is an error or a no-op.
2. **Record it as a deliberate deviation** — a concrete path is arguably better for a tool that fans out to production reference hosts, since a glob expanded locally can silently mean different files than the operator pictured. If so, `put` should *reject* an argument containing an unexpanded `*`/`?` with a message saying globs are not supported, rather than reporting a not-found path.

Either way the comment must stop citing a finished phase, and option 2 is only defensible if the argument is actively rejected — the current behaviour, where the glob is treated as a literal filename that happens not to exist, is the one outcome nobody wants.

Filed as a question rather than a bug: which of the two is wanted is a project-direction call.

mimi1vx:
I'm for reject + document -> glob can put things you really don't want ..

PR #436, opened 2026-08-09
MCP get/put carry file content in-band instead of exchanging server-local paths

### LINKED ISSUE #434: get returns a server-side path the caller cannot read under --transport http (put has the mirror problem)
## The defect

`get` downloads a remote file to the **local** filesystem —
`{report_wd}/downloads/{basename}.{host}` (`sftpget.rs:52-62`) — and reports that
absolute path back to the caller (`sftpget.rs:75-95`).

Under `--transport stdio` that is fine: the server is a child process of the client on the
same machine, so the path the caller is handed is a path it can open.

Under `--transport http` it is not. That transport "serves many clients with per-client
session isolation" (`mtui-mcp/src/main.rs:11`), so "local" means **the server's**
filesystem. `get` then succeeds, and hands back an absolute path the caller has no way to
read. The transfer itself is reported honestly — per-host failures do fail the call
(`sftpget.rs:88-95`) — so this is not a silent failure; it is a success that yields an
unusable reference.

`put` has the mirror problem and is worse: it takes a **local** path
(`sftpput.rs:83-103`) and uploads it to every enabled host. A remote client cannot place a
file on the server to name, so it can only upload files that happen to already exist
there. There is no way to supply content over the wire at all.

Both are exposed by default: neither appears in the permanent deny-list (`deny.rs`) nor in
`CORE` (`profiles.rs`), so the default `full` profile advertises both to every client,
including under `http`.

## Second problem: the path has no per-client component

`{report_wd}/downloads/{basename}.{host}` derives only from `template_dir`, the RRID, the
remote basename, and the host — nothing session- or user-scoped. Under `http`, two clients
working the same RRID write into the same directory on the same server filesystem: client A
can read what client B downloaded off a reference host, and they overwrite each other's
files with no error. That is adjacent to the attribution gap in #412, but it is a
confidentiality/collision issue rather than an audit one.

## Why it matters

A tester driving mtui through a remote MCP server cannot retrieve a file from a reference
host at all — the one thing `get` exists for. The tool is offered, appears to work, and the
result is unreachable. On a shared server the artefacts also pile up in a place other
clients can see.

## Options

1. **Return the content in-band.** `testreport_logs` already establishes the pattern of
   returning file content in a tool result. This works for both transports and is what a
   model calling the tool actually wants. Needs a size cap, base64 for non-UTF-8 payloads,
   and an explicit truncation marker rather than silent clipping — a large binary would
   otherwise blow the caller's context.
   For `put`, the inverse: accept content in the call rather than a server-side path.
2. **Drop `get`/`put` from the surface under `http`.** They are already absent from `core`,
   so this is a small change and is honest — better an unavailable tool than one that
   returns a path to nowhere.
3. **Leave stdio as it is** either way; the current behaviour is correct there and REPL/CLI
   users are unaffected.

My preference is 1 for `get` (with the cap) and 2 as the immediate stop-gap if 1 is not
wanted soon, since the tool is currently advertised by default in a mode where it cannot
work. The per-client path collision above wants fixing regardless of which is chosen, if
`get` keeps writing files at all.

## Not claimed

This is **not** a defect under `stdio`, and not a defect for the REPL or CLI. An earlier
framing of mine said `get` over MCP was broken regardless of transport; that is wrong, and
the narrower claim above is the one supported by the code.

plusky:
Closes #434.

## Summary

Over `--transport http`, the MCP `get` tool downloaded to the **server's** disk and returned a path the remote client cannot read; `put` took a server-local path, so a remote client could not supply content at all. Both are now denied from synthesis (the `edit` → `testreport_*` precedent) and re-served under the same names as hand-written tools that carry content **in-band**, on both transports — in-band is strictly better on stdio too. MCP `get` no longer touches the server's disk at all, which also retires the `{report_wd}/downloads` per-client collision (#434's second finding) for MCP callers. The REPL `get`/`put` commands are byte-untouched (#399's literal-path contract included).

## Shape

- **mtui-hosts**: `Target::sftp_read` (over the existing `Connection::sftp_open`) plus `HostsGroup::{sftp_read,sftp_read_selected,sftp_put_bytes}` map fan-outs — one entry per *attempted* host, keyed by hostname (the `lock()` shape), deliberately not the never-cleared `last_download`/`last_upload` bookkeeping.
- **get**: per host `content` (UTF-8) or `content_b64` (binary), always with the full remote `size` and an explicit `truncated` flag. Per-host source cap at `[mcp] max_input_bytes`; equal share of `[mcp] max_output_bytes` on the wire, truncated on a char boundary with **no notice spliced into the content**. A source cap landing mid-codepoint trims to the last whole codepoint rather than reclassifying text as binary. `hosts` narrows/retries a subset; unknown or disabled names refuse.
- **put**: payload in the call (`content` xor `content_b64`), placed at the report's remote working directory under a validated bare `filename`; oversize payloads are refused rather than truncated.
- **No partial success**: any host failure fails the whole call, naming each failed host with its reason; zero enabled hosts refuses instead of succeeding vacuously.
- Both tools take the exclusive gate like synthesized fan-out commands, and restore the active-template pointer while **releasing the entry guard** afterwards — the `run_command` exclusive-arm contract (a lingering guard would make the next scoped call silently dispatch against the null report). Regression-tested.

## Review

Three read-only adversarial review agents (false-success paths / test quality / project contracts) ran over the uncommitted tree; all findings were fixed before the commits were created. The two that mattered most: the lingering-active-guard blocker above, and a `truncated` flag that stayed false on small wire-cap overruns (fixed by flagging from the pre-cap length and dropping the spliced notice entirely).

## Testing

- `cargo fmt --all --check`; `cargo clippy --workspace --all-targets -- -D warnings`
- `cargo test --workspace`; `cargo test -p mtui-mcp -F mcp` (24 unit + 53 integration in the `it` harness, 0 failures)
- `typos .`
- Red-first: the extended `mcp_denylist_is_consistent` pin was observed failing against the pre-flip list; the surface snapshots regenerate to drop the synthesized `get`/`put` rows.
- Server-level pins: the pair is advertised **and dispatched** on the wire (an unplugged dispatch arm fails the test), and `[mcp] tools_deny = ["put"]` removes the tool from both `tools/list` and dispatch.
- Mock hardening: `MockSftpOp::PutBytes` now records the payload bytes, so a same-length corrupted buffer is detectable; per-host fixtures use distinct content throughout so attribution is provable.

## Notes

- Commit 2 is an intentional intermediate state: the issue's own stop-gap (option 2) — the broken path-based pair leaves the surface for one commit before the in-band pair lands in commit 3.
- `Connection::sftp_open` buffers the whole remote file before the source cap applies (documented in the tool description); a streaming `max_bytes` variant is a possible follow-up if multi-GB `get`s become a real pattern.


mimi1vx (approved):
lgtm, its going ouside single thruth registy .. but on put/get this divergence is acceptable

Issue #400, opened 2026-07-30
update templates discard the patch command's exit code, so a failed patch with clean output passes


plusky:
The `update` templates discard the patch command's exit status, so a `transactional-update` or `zypper` patch that genuinely fails with clean output still passes the update check. Found while giving SL Micro and YUM hosts an update verdict at all (#393); deliberately left out of that PR because the fix changes a remote command string and widens the rollback blast radius.

## Mechanism

`SLM_UPDATE` and `ZYPPER_UPDATE` (`crates/mtui-testreport/src/update_workflow/actions/update.rs`) are multi-line scripts run as **one** remote `exec`. Their last line is the repo-cleanup loop:

```sh
zypper -n lr | awk ... | while read r; do zypper -n rr $r; done
```

The shell reports *that loop's* status. It is `0` whenever the loop body never runs — i.e. whenever there is nothing to clean up. The `transactional-update -n pkg in ... -t patch` / `zypper -n in -t patch` status is discarded before anything can read it.

## Consequences today

- The `("slmicro", true)` update check added in #393 can only be **marker-based** — lock strings, dependency error, RPM error, plus the `-1` never-ran sentinel. A `transactional-update` that exits non-zero with clean output passes.
- `checks/update.rs`'s `exitcode == 104` branch is close to dead in production for both zypper and slmicro, since the exit code it inspects is the cleanup loop's.
- The `("YUM", false)` check is narrowed to did-it-run for the same reason, plus the sibling-command stderr problem.

## Fix and its blast radius

Capture the patch command's status and re-exit with it at the end of the template. Three risks, which is why this wants maintainer sign-off rather than a drive-by PR:

1. **It changes a remote command string** that lands in `show_log` transcripts and in `/var/log/mtui.log`. AGENTS.md treats those as contract-ish.
2. **The templates use `SubstMode::Safe`**, so any `$?` must be written `$$?` to survive substitution. Easy to get wrong and easy to miss, since the wrong version still runs.
3. **It would activate the currently-dead `104` branch.** A zypper host whose patch exits 104 would newly fail the update *and* fire the group-wide rollback downgrade — reverting every host in the group. That is a behaviour change on a path that has been silently passing, so it needs a decision about whether 104 should fail at all, not just whether the exit code should be visible.

## Acceptance

[...]

Issue #397, opened 2026-07-30
No fixture exercises a populated SLFO packages envelope, so the SLFO parse path is untested


plusky:
Split out of #396, where it is the missing evidence that blocks root-causing a `prepare` no-op. It is also a coverage defect in its own right: **no test in the tree exercises a populated SLFO `packages` envelope**, so the suite cannot distinguish "SLFO package metadata parses" from "SLFO package metadata is silently dropped".

## What the fixtures actually cover

`base().packages` has one production writer, `JSONParser::parse` (`crates/mtui-testreport/src/metadata_parsers.rs:208-230`), which requires each entry to be exactly three whitespace-separated tokens (`"name <discarded> version"`) and drops anything else with **no log and no else branch**.

Every fixture that populates that map is a classic `SUSE:Maintenance` shape:

| fixture | `packages` |
|---|---|
| `crates/mtui-testreport/tests/fixtures/metadata/metadata.json:13-20` | `"15-SP3": ["sle-module-python2-release = 15.3-150300.59.4.1"]` |
| `crates/mtui-testreport/tests/lifecycle.rs:35-37` | `"15-SP3": ["sle-module-python2-release = 15.3-150300.59.4.1"]` |
| `crates/mtui-testreport/tests/metadata_parsers.rs:185` | `"test_prod": ["test_pkg 1.0 1.0"]` |

The only SLFO end-to-end fixture — `crates/mtui-testreport/tests/lifecycle.rs:453` — is:

```rust
"packages": {{}},
```

So the SLFO path is proven to *load*, and proven to load with nothing in it. Nothing proves it loads packages.

## Why the gap has teeth

`packages_for_map` (`crates/mtui-testreport/src/testreport.rs:246-260`) branches on the map's shape to seed each host's tracked packages:

```rust
if map.len() == 1 && map.contains_key("standard") {
    for (name, ver) in &map["standard"] { selected.insert(name, ver); }
} else if let Some(per_product) = map.get(base_version) {
```

The `"standard"` branch exists specifically for "a report that ships one product-agnostic set — e.g. SLFO metadata" (its own doc comment). But that correspondence is asserted **only by a unit test over a hand-built map**, never by a fixture representing real SLFO metadata. If real SLFO metadata is keyed by something else — `"6.1"`, a product name, anything not matching the host's parsed base version — seeding silently resolves to nothing (`crates/mtui-core/src/session.rs:974` skips `set_packages` without a warning), which is one of the paths behind #396's second defect.

[...]

PR #437, opened 2026-08-09
prepare/update refuse metadata-empty reports; export stops claiming never-checked packages are absent

### LINKED ISSUE #396: prepare reports success without installing anything, and export then claims installed packages are absent (SUSE:SLFO:1.1:418286)
Reported from a real validation session: `SUSE:SLFO:1.1:418286` (afterburn / afterburn-dracut security update) on two SL-Micro 6.1 transactional refhosts, aarch64 + x86_64.

`prepare` reported success and installed nothing. The operator only discovered this by checking `rpm -q` by hand; the package was then installed via `install`, landing directly on the staged build, and the exported testreport claimed `afterburn` was **not installed** on hosts where it was installed and running.

Two independent defects, both of the "reports success while doing nothing" shape.

## 1. `prepare` dispatches no command and still reports success

Call sequence (via `mtui-mcp`): `lock` → health check → `set_repo(add=true)` → `prepare(update=true)` → `install(...)` → `export`.

`prepare(update=true)` returned only:

```
prepare completed on cronos.qam.suse.cz, ulysse.qam.suse.cz
```

`list_packages` immediately afterwards still showed `afterburn: not installed` on both hosts. The per-host install log for the `prepare` phase contains exactly three commands:

```
# zypper -n ar -cfGkn 'issue-SL-Micro:6.1:p=1.1:418286' http://...
# zypper -n ar -cfGkn 'issue-SL-Micro-Extras:6.1:p=1.1:418286' http://...
# transactional-update --continue --non-interactive run zypper --gpg-auto-import-keys -n ref
```

That is `fanout_set_repo` and nothing else — no `transactional-update pkg in` anywhere in the phase. The package only appears later under the explicit `install` call.

**This is not the resolver declining to pull the package in.** The `slmicro` preparer template is an unconditional install of the whole list (`crates/mtui-testreport/src/update_workflow/actions/prepare.rs:63`):

```rust
command: format!("transactional-update -n pkg in -l {parameter} $package"),
```

Had it been dispatched it would have installed `afterburn` regardless of whether anything on the host depends on it. The command was never built.

### Why it can be never built, silently

`prepare_body` (`crates/mtui-testreport/src/reports/update_flow.rs:690`) guards the single-transaction dispatch and has **no `else`**:

```rust
} else if !pkgs.is_empty() {
    let joined = quote_args(pkgs);
    let cmd = build_prepare_map(targets, registry, Some(&joined), false);
    targets.run(Command::PerHost(cmd.clone())).await;
    note_dispatch(targets, &cmd, &mut dispatched, &mut inert);
}
```

An empty `pkgs` means repo-add + refresh, then `Ok(())`. This is currently *pinned as intended* by `perform_prepare_does_not_reboot_a_host_with_nothing_staged` (`update_flow.rs:3288`), which asserts `res.is_ok()` with the comment "nothing to prepare is not a failure". Not rebooting is right; returning a bare success with no word to the operator is the bug.

`pkgs` comes from `TestReport::get_package_list()` (`crates/mtui-testreport/src/testreport.rs:462`), which flattens `base().packages`, so an empty or unparsed package section in the loaded report produces an empty list with no error anywhere.

`build_prepare_map` (`update_flow.rs:814`) then has three more silent drops:

```rust
let Some((release, transactional)) = host_key(target) else { continue; };
let Some(doer) = resolve_doer(registry, Role::Prepare, &release, transactional) else { continue; };
...
if let Some(cmd) = rendered { map.insert(...); }
```

and the success line is built from the **selected** group, not from the dispatch map (`crates/mtui-core/src/commands/perform.rs:124`):

```rust
let hosts_label = selected.names().join(", ");
```

so `prepare completed on <hosts>` can name hosts that were never sent a command. `build_reboot_map` does not catch this either: it hard-errors on a missing doer but `continue`s on a `None` host key (`update_flow.rs:286`).

For this run the host key and doer must have resolved — the later `install` used `transactional-update -n pkg in -l`, and a missing preparer would have aborted `prepare` with `missing preparer` before locking. That leaves the empty package list as the mechanism.

### Where an empty package list comes from: two silent drops in the metadata parser

`base().packages` has exactly **one** production writer, `JSONParser::parse` (`crates/mtui-testreport/src/metadata_parsers.rs:208-230`), fed only from the checked-out `metadata.json`. (The template `log` text is parsed by `ReducedMetadataParser`, which touches `slack_review`/`hostnames`/`jira`/`bugs` and never packages. TeReGen and the QEM dashboard never write it.) `SUSE:SLFO:1.1:418286` resolves to `SlReport`, but no report type overrides `read` — SLFO and `SUSE:Maintenance:X:Y` share the identical population chain, so this is not SLFO-specific plumbing.

The parser requires each entry to be **exactly three whitespace-separated tokens**, `"name <discarded> version"`:

```rust
let mut tokens = entry.split_whitespace();
if let (Some(pkg), Some(_), Some(ver)) =
    (tokens.next(), tokens.next(), tokens.next())
{
```

**There is no `else` branch and no log.** A two-token entry (`"afterburn 5.9.0-…"`), a bare NEVRA (`"afterburn-5.9.0.git21.a73f509-slfo.1.1_1.1.x86_64"`), or a four-token entry all vanish without a trace. The sibling failure one line down — an invalid RPM name — *does* log (`error!(… "skipping invalid package name in metadata")`), so the shape mismatch is the one hole in an otherwise-guarded loop.

Then `packages.insert(prod.clone(), pkgs)` runs **unconditionally**, even when every entry was dropped. That produces a state worth naming, because it defeats the existing guard: `base.packages` is *non-empty* (it has the product key) while `get_package_list()` is *empty* (it flattens the inner keys). `list_packages -w`'s "No wanted package versions in the loaded report." check tests `base.packages`, so it stays silent in exactly this case — `crates/mtui-core/src/commands/listpackages.rs:195-198`.

The discriminating evidence is the on-disk `metadata.json` for this RRID (`packages` object: the product key, and whether each entry has three tokens). I could not check it myself — the report is no longer in the local cache — and no fixture in the tree answers it either. **Tracked separately as #397**, which also covers the coverage gap that let this reach production: the only SLFO end-to-end fixture ships `"packages": {}`, so the suite passes whether or not SLFO package metadata parses at all. If the real envelope turns out not to match the parser, that is this issue's root cause and #397 becomes its regression test.

One loose end I could not close: `export` did write `afterburn`/`afterburn-dracut` lines, which argues the host's tracked package list was *not* empty — and that list is seeded from the same metadata. Either a different exporter path wrote those lines, or the template already carried them. Whoever picks this up should reconcile that before assuming a single cause.

### A third silent no-op on the same data, worth fixing together

Host package seeding at `crates/mtui-core/src/session.rs:971-977`:

```rust
let seeded = mtui_testreport::testreport::packages_for_map(package_meta, &base_version);
if !seeded.is_empty() {
    target.set_packages(seeded);
    target.query_versions().await;
}
```

`packages_for_map` selects the sub-map keyed by the host's parsed base version, or the whole map if its only key is `"standard"` (`crates/mtui-testreport/src/testreport.rs:246-260`). SL-Micro's base version comes from the product XML `<version>` element, and the parser test asserts `SL-Micro` → `"6.0"`. So a metadata map keyed by anything other than the host's exact parsed version — or a lone `"standard"` — seeds **nothing**, with no warning, and `package_check` then has no packages to record before/after for. That is a plausible second contributor to defect 2 below.

### What should change

Regardless of which drop fired, the operator was told "completed". Suggested:

Ranked by how much silence each one removes:

1. **Log the dropped metadata entry.** Give the three-token `if let` an `else` that logs the rejected entry verbatim, matching the invalid-RPM-name arm right below it. This is the cheapest change and it converts the whole class from invisible to diagnosable.
2. **Do not insert an empty product sub-map** (or make the `list_packages -w` guard test `get_package_list()` rather than `base.packages`), so `base.packages` non-empty / package list empty stops being reachable.
3. **`prepare` with an empty resolved package list is a warning at minimum**, not a bare success — it means the loaded report told mtui nothing to install. Note this needs a decision about `perform_prepare_does_not_reboot_a_host_with_nothing_staged`, which currently asserts `is_ok()` for that case; the *reboot* half of that test should stay, the bare-success half is what is in question.
4. **The confirmation line should name the hosts actually dispatched to**, not the selected set, and each `continue` in `build_prepare_map` should log which host it dropped and why.
5. **Warn when host seeding resolves nothing** (`session.rs:974`) — silently skipping `set_packages` is what makes defect 2 possible.

## 2. `export` writes "is not installed" for packages that are installed

The exported testreport's per-arch block read, for both hosts:

```
before:
	package afterburn is not installed
	package afterburn-dracut is not installed
after:
	package afterburn is not installed
	package afterburn-dracut is not installed
```

`after` was false: the packages were installed and running the staged build, confirmed independently with `rpm -q`, `rpm -qi`, `rpm -V` and `afterburn --version`. This had to be corrected by hand via `testreport_patch`.

The mechanism is that before/after are recorded **only by the update flow**. `package_check` is called exactly twice, both inside `update_run_phase`'s surroundings (`update_flow.rs:1319` with `post=false`, `:1389` with `post=true`), and it is the only writer of those fields (`crates/mtui-hosts/src/target/hostgroup.rs:860-862`):

```rust
if post { pkg.set_after_version(current.clone()); } else { pkg.set_before_version(current.clone()); }
```

`install` never calls it. `export` then renders `None` as the literal claim that the package is absent (`crates/mtui-testreport/src/export/manual.rs:250-252`):

```rust
let new_line = match version {
    Some(v) => format!("\t{name}-{v}\n"),
    None => format!("\tpackage {name} is not installed\n"),
};
```

So an unknown version is printed as a **positive assertion of absence**, and the block is then auto-marked FAILED by the `before >= after` rule. A testreport is the artifact a release decision is read off; "we did not observe this" and "this package is not installed" are different statements and only the first is true here.

### What should change

`None` should render as unknown/not-recorded, distinct from an observed absence — or `export` should probe the host rather than trusting a snapshot the flow may never have taken. Ideally `export` warns when it is asked to write a per-arch block for a package whose after-state was never recorded.

## Note on `prepare -u` and the "released baseline"

Worth settling as part of this, because it affects the documented workflow: `perform_prepare` maps `testing` straight onto the repo operation (`update_flow.rs:598`):

```rust
let operation = if testing { RepoOp::Add } else { RepoOp::Remove };
```

So `prepare --update` **adds** the issue repo and then installs — meaning it would resolve to the highest available version, i.e. the staged build, not a released baseline. The QE validation guidance treats `prepare` as the step that establishes a *released* baseline before `update`; that property only holds for `prepare` *without* `-u`. If that reading is right, the flag whose name suggests "prepare for the update" is the one that destroys the baseline the before/after table depends on, which is worth either a doc fix or a warning.

## Environment

- mtui 26.1.1 / 26.2.0-dev (Rust workspace)
- Hosts: `cronos.qam.suse.cz` (aarch64), `ulysse.qam.suse.cz` (x86_64) — SL-Micro 6.1, transactional
- RRID: `SUSE:SLFO:1.1:418286`

plusky:
Closes #396.

## Summary

On `SUSE:SLFO:1.1:418286`, `prepare` reported success while installing nothing, and `export` then claimed the installed packages were absent. Root cause (established during #433's investigation, refuting the wrong-repo-layout hypothesis): the metadata envelope's packages either failed to parse or were empty, every silent-drop site on the chain converted "did nothing" into `Ok`, and export rendered never-observed state as a positive absence claim.

This fixes the whole chain, in causal order:

- **Parser**: unparsable package entries are logged and dropped (previously vanished); a product with no parsable entries no longer leaves an empty sub-map that defeats the `get_package_list()` guards; trailing tokens warn.
- **Prepare/update commands**: refuse up front on a report naming no package versions — no more `prepare completed on <hosts>` over a no-op (REPL and MCP). Not-loaded is checked first so the refusal never claims a report is loaded when none is.
- **Prepare flow**: a host `build_prepare_map` drops (unresolvable release, failed render) fails the run **by name** with mock-level proof it received no command; empty lists warn on every path.
- **Observation honesty** (`mtui-types`): `Package::{before,after}` are now a three-state `VersionCheck` — `NotChecked` / `NotInstalled` / `Installed(RPMVersion)` — so "checked, absent" and "never checked" are distinguishable; `required`/`current` stay plain options, and `before()`/`after()` keep their `Option<&RPMVersion>` signature. The `downgrade` rotation moves the whole check, so "never checked" survives it.
- **Export**: never-checked renders `package <name>: not checked (no version data recorded)`; a host with no recorded data keeps the scaffold and its `=> PASSED/FAILED` placeholder (never flips PASSED over unverified content); FAILED outranks unverified; `export` prints an operator-visible `WARNING:` line per unverified host (with a negative control pinning it doesn't fire for verified hosts).

## Review

Three read-only adversarial reviewers (false-success remnants / test vacuity / contracts) ran over the uncommitted tree; every finding was fixed before the commits were created — including two they rated medium: the guard-order bug (refusal claiming "the loaded report" with none loaded) and the downgrade rotation stamping a never-checked slot as checked. The one *confirmed non-regression*: `package_check` stamps observations unconditionally, so real update-flow exports keep saying "is not installed" for genuinely-absent packages — "not checked" appears only when no check ever ran.

## Testing

- Full gate: fmt, clippy `-D warnings`, workspace tests, `-p mtui-mcp -F mcp`, rustdoc gate, typos.
- Six red-first proofs, each observed failing against the mutated fix (with the test observably *running* — one initial "0 passed" false-negative was caught and re-proven against the right harness) and restored with `MUTANT`-marker greps.
- New pins: prepare/update refusal (+ no-success negative), per-host dispatch accounting (attribution via `err.host`, not substring), parser drop/keep matrix, the `VersionCheck` state machine across all four setters, four export-honesty tests plus the mixed-block both-line-kinds pin, export WARNING presence *and* absence.

## Known residuals (documented, out of scope)

- `package_check` can stamp "observed absent" when the *version query itself* never answered (host died between fan-out and check) — a pre-existing seam `VersionCheck` cannot see; noted for #397's fixture work.
- `export`'s WARNING covers the no-data-at-all host; a host with seeded-but-never-observed packages is visible in the report body ("not checked" lines + kept placeholder) but not as a display WARNING.
- The unreachable-cancel inventory log and #400/#406 interactions are sibling-issue scope.

---

**Review round 2 (2026-08-10).** Rebased onto `ae8b91db` (the #434 merge conflicted in `CHANGELOG.md` and `docs/src/mcp.md`; both resolved by keeping each side's entries), and @mimi1vx's point about the two `*_observed` booleans is addressed in `9a38923`: the flag beside the option became one field with three states (`VersionCheck`). The value and the fact that it was measured can no longer drift apart, and the downgrade rotation's guard is gone — moving the whole check carries "never checked" across on its own. `downgrade_verdict_rotation_keeps_an_unchecked_slot_unchecked` pins that, and fails against the naive value-only rotation.

### DIFF
--- a/crates/mtui-types/src/package.rs
+++ b/crates/mtui-types/src/package.rs
@@ -28,6 +28,12 @@ pub struct Package {
     after: Option<RPMVersion>,
     required: Option<RPMVersion>,
     current: Option<RPMVersion>,
+    /// Whether a before-version observation was ever recorded (#396): a `None`
+    /// [`before`](Package::before) with this set means *observed absent*, not
+    /// *never checked* — export must render the two differently.
+    before_observed: bool,
+    /// The after-side counterpart of [`before_observed`](Package::before_observed).
+    after_observed: bool,
 }
 
 impl Package {
@@ -40,7 +46,23 @@ impl Package {
             after: None,
             required: None,
             current: None,
+            before_observed: false,
+            after_observed: false,
         }
+    }
+
+    /// Whether a before-version observation was ever recorded — `None` from
+    /// [`before`](Package::before) then means "observed absent", not "never
+    /// checked".
+    #[must_use]
+    pub fn before_observed(&self) -> bool {
+        self.before_observed
+    }
+
+    /// The after-side counterpart of [`before_observed`](Package::before_observed).
+    #[must_use]
+    pub fn after_observed(&self) -> bool {
+        self.after_observed
     }
 
     /// The version of the package before an update, if known.
@@ -77,6 +99,7 @@ impl Package {
     /// as "clear" and never errors.
     pub fn set_before(&mut self, ver: Option<&str>) -> crate::error::Result<()> {
         self.before = parse_opt(ver)?;
+        self.before_observed = true;
         Ok(())
     }
 
@@ -86,6 +109,7 @@ impl Package {
     /// See [`set_before`](Package::set_before).
     pub fn set_after(&mut self, ver: Option<&str>) -> crate::error::Result<()> {
         self.after = parse_opt(ver)?;
+        self.after_observed = true;
         Ok(())
     }
 
@@ -98,14 +122,18 @@ impl Package {
         Ok(())
     }
 
-    /// Sets the [`before`](Package::before) version directly.
+    /// Sets the [`before`](Package::before) version directly, recording that
+    /// an observation was made (even a `None` one — observed absent).
     pub fn set_before_version(&mut self, ver: Option<RPMVersion>) {
         self.before = ver;
+        self.before_observed = true;
     }
 
-    /// Sets the [`after`](Package::after) version directly.
+    /// Sets the [`after`](Package::after) version directly, recording that an
+    /// observation was made (even a `None` one — observed absent).
     pub fn set_after_version(&mut self, ver: Option<RPMVersion>) {
         self.after = ver;
+        self.after_observed = true;
     }
 
     /// Sets the [`current`](Package::current) version directly.
@@ -147,8 +175,30 @@ impl std::hash::Hash for Package {
 
 #[cfg(test)]
 mod tests {
-    use super::*;
     use std::collections::HashSet;
+
+    use super::*;
+
+    /// #396: setting a `None` version still counts as an observation —
+    /// "observed absent" and "never checked" must stay distinguishable.
+    #[test]
+    fn set_version_none_counts_as_observed() {
+        let mut p = Package::new("bash");
+        assert!(!p.before_observed() && !p.after_observed());
+        p.set_before_version(None);
+        assert!(p.before_observed());
+        assert!(!p.after_observed());
+        p.set_after_version(None);
+        assert!(
+            p.after_observed(),
+            "set_after_version must mark the observation"
+        );
+        let mut p = Package::new("bash2");
+        p.set_after(None).unwrap();
+        assert!(p.after_observed());
+        assert!(!p.before_observed());
+        assert!(p.before().is_none() && p.after().is_none());
+    }
 
     #[test]
     fn new_has_name_and_no_versions() {
--- a/crates/mtui-types/src/package.rs
+++ b/crates/mtui-types/src/package.rs
@@ -8,11 +8,12 @@
 //!
 //! Each version field is stored as a single typed representation, with
 //! fallible setters that parse a `&str` rather than silently storing an
-//! unparsed string. `before` and `after` are [`VersionCheck`]s, not
-//! `Option<`[`RPMVersion`]`>`: those two are *measured* on a target, so they
-//! must be able to say "checked, not installed" and "never checked" apart
-//! (#396). `required` (declared by the update metadata) and `current` stay
-//! plain options.
+//! unparsed string. `before`, `after` and `current` are all [`VersionCheck`]s,
+//! not `Option<`[`RPMVersion`]`>`: all three are *measured* on a target, so
+//! they must be able to say "checked, not installed" and "never checked" apart
+//! (#396, #437 — a host the version query never reached must not read the
+//! same as one it confirmed the package absent on). `required` (declared by
+//! the update metadata, never measured) stays a plain option.
 //!
 //! A `Package` hashes and compares **by name only**, so it can live in a
 //! name-keyed set regardless of its version fields.
@@ -30,7 +31,7 @@ pub struct Package {
     before: VersionCheck,
     after: VersionCheck,
     required: Option<RPMVersion>,
-    current: Option<RPMVersion>,
+    current: VersionCheck,
 }
 
 /// The outcome of checking one of a package's versions on a target.
@@ -90,7 +91,7 @@ impl Package {
             before: VersionCheck::NotChecked,
             after: VersionCheck::NotChecked,
             required: None,
-            current: None,
+            current: VersionCheck::NotChecked,
         }
     }
 
@@ -130,9 +131,18 @@ impl Package {
     }
 
     /// The version currently installed on a target, if known.
+    ///
+    /// `None` covers both "checked, not installed" and "never checked" — use
+    /// [`current_check`](Package::current_check) when the difference matters.
     #[must_use]
     pub fn current(&self) -> Option<&RPMVersion> {
-        self.current.as_ref()
+        self.current.version()
+    }
+
+    /// The current-version check, including whether it ran at all.
+    #[must_use]
+    pub fn current_check(&self) -> &VersionCheck {
+        &self.current
     }
 
     /// Sets the [`before`](Package::before) version from an optional string,
@@ -191,9 +201,17 @@ impl Package {
         self.after = check;
     }
 
-    /// Sets the [`current`](Package::current) version directly.
+    /// Sets the [`current`](Package::current) version directly, recording a
+    /// completed check — `None` is *observed absent*, not *never checked*.
     pub fn set_current_version(&mut self, ver: Option<RPMVersion>) {
-        self.current = ver;
+        self.current = ver.into();
+    }
+
+    /// Sets the [`current`](Package::current) check outcome directly,
+    /// including [`NotChecked`](VersionCheck::NotChecked) — the version
+    /// query never having answered for this package at all (#437).
+    pub fn set_current_check(&mut self, check: VersionCheck) {
+        self.current = check;
     }
 }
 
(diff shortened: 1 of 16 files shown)


crates/mtui-types/src/package.rs
```diff
@@ -28,6 +28,12 @@ pub struct Package {
     after: Option<RPMVersion>,
     required: Option<RPMVersion>,
     current: Option<RPMVersion>,
+    /// Whether a before-version observation was ever recorded (#396): a `None`
+    /// [`before`](Package::before) with this set means *observed absent*, not
+    /// *never checked* — export must render the two differently.
+    before_observed: bool,
```
mimi1vx: this two booleas should not be needed, as after, required and current are `Option<>` type so its `Some<RPMVersion>` or `None`  sou I would look at parse_opt and maybe fix this ?

mimi1vx (changes requested):

crates/mtui-types/src/package.rs
plusky: Agreed on the two booleans — fixed in 9a38923, and following your pointer at `parse_opt`.

The three states are needed, though: an `Option<RPMVersion>` has no way to say *why* it is empty, and "the check ran and the package is not installed" and "no check ever ran" are the two facts #396 is about — the export renders them differently. So instead of a flag beside the option, `before`/`after` are now one field with three states:

```rust
pub enum VersionCheck {
    NotChecked,               // no check ran; version unknown
    NotInstalled,             // a check ran and found the package absent
    Installed(RPMVersion),    // a check ran and found this version
}
```

`parse_opt` still returns `Option<RPMVersion>`; the before/after setters feed it through `VersionCheck::from`, which reads `None` as *checked, absent*. `required` keeps the plain option — metadata declares a requirement, it does not observe one. `before()`/`after()` keep their `Option<&RPMVersion>` signature, so nothing that only wants the version had to change, and the export matches the enum directly instead of a `(version, observed)` pair.

It also removed code rather than adding it: the downgrade rotation had to guard the before/after move so an unchecked slot did not become an observation, and moving the whole check carries that across on its own — `pkg.set_before_check(pkg.after_check().clone())`, guard gone. New test `downgrade_verdict_rotation_keeps_an_unchecked_slot_unchecked` pins it; it fails against the naive value-only rotation.

Also rebased onto main (the #434 merge conflicted in CHANGELOG.md and docs/src/mcp.md — both resolved by keeping each side's entries).

mimi1vx (approved):

Issue #405, opened 2026-07-30
forced job_cancel strands the operation lock on every host in the group


plusky:
A forced `job_cancel` drops the worker future mid-`await`, so the operation's `group.unlock()` never runs and `/var/lock/mtui.lock` stays held on **every host in the group** until the session is evicted or someone runs `unlock`/`unload` by hand. A stranded lock blocks every other tester on those hosts.

Not a false success — the job does report `Cancelled` — but it strands remote locks at exactly the moment a cancel needs them released.

## Mechanism

`crates/mtui-mcp/src/session.rs:1281-1295` — `job_cancel` waits a 1s grace (`CANCEL_GRACE`, session.rs:117-127, "kept short so `job_cancel` stays responsive; not a config key") then aborts the worker `JoinHandle` directly:

```rust
token.cancel();
let mut forced = false;
if let Some(mut handle) = handle {
    if tokio::time::timeout(CANCEL_GRACE, &mut handle).await.is_err() {
        // Forced stage: the body never reached a checkpoint.
        forced = true;
        handle.abort();
        // Await the aborted task so cancellation has fully unwound
        // before we return; a `JoinError::Cancelled` is expected.
        let _ = handle.await;
    }
```

Aborting drops the body's future. On the concurrent dispatch path an `AbortOnDrop` guard (session.rs:136-142, used at session.rs:801) propagates the abort inward; on the exclusive path the dispatch runs inline (session.rs:836) and the future is dropped directly. Either way, every `.await` between the lock and the unlock is a point where the unlock is skipped.

## It falsifies a stated invariant

`crates/mtui-hosts/src/target/operation.rs:18` (module doc): `//! 4. **always** ``group.unlock()`` afterwards.`
`operation.rs:336`: `///  * ``unlock()`` unconditionally afterwards.`
`operation.rs:370-374`:

```rust
// Everything past the lock must reach `unlock()`, so this section stays
// free of `?`. It is infallible over the group seam today (run/reboot
// return `()`); a future fallible step must capture its error and fall
// through to the unlock rather than returning early.
```

The section is deliberately `?`-free so the unlock is guaranteed — but that guards against *early return*, not against *future drop*. Nothing survives a dropped future. The same shape is in the update flow (`crates/mtui-testreport/src/reports/update_flow.rs:1355-1372`, "run + check + reboot under the lock (unlock always)").

[...]

PR #443, opened 2026-08-10
REPL Ctrl-C cancels the running command cooperatively instead of killing the process

### LINKED ISSUE #441: REPL Ctrl-C during a host operation kills the process and strands a dead-pid operation lock on every host
Found while fixing #405 (PR #440); verified against the code at `49ad1bed`.

## Summary

A Ctrl-C in the REPL during a long host operation kills the mtui process outright, and the operation lock (`/var/lock/mtui.lock`) it had taken on every host stays behind with a **dead pid** — so unlike #405's in-process strand, the debris reads as *foreign* (`is_mine()` requires user AND pid to match) and neither a later mtui run nor session teardown can release it non-forced. It blocks every other tester on those hosts until someone runs `unlock --force` or the 24h stale reap (`lock_reap_stale`/`lock_stale_age`, `locks.rs:381-404`) fires on their next claim attempt.

## Mechanism

- reedline observes `Signal::CtrlC` only during `read_line` (`crates/mtui-cli/src/repl.rs:189-195` — "abort the current input and reprompt"); dispatch runs after `read_line` returns, awaited inline.
- Nothing installs a process-level SIGINT handler. `crates/mtui-core/src/commands/request_review.rs:223-225` documents this explicitly: "Nothing in the CLI installs a SIGINT handler today, so without this a Ctrl-C during a watch kills the process outright" — its `sleep_or_interrupt` is a bespoke, one-command mitigation, as is `regenerate.rs:136`'s.
- So the default SIGINT disposition terminates the process mid-`update`/`install`/`prepare`/`run`, between the flow's lock and its "unlock always" section. There is no `Drop`-based release and cannot be (the release is an SSH round-trip; `Drop` cannot await — the same reasoning as `provider.rs:355-357` for pool claims).

The cooperative-cancellation seam (`Session`'s `CancellationToken`, checked at the flows' step boundaries) already exists and is exactly what a SIGINT handler should trigger — today the REPL never cancels it; only MCP `job_cancel` does.

## Suggested direction

Install a SIGINT handler for the duration of a dispatched command (restore reedline's behaviour between commands): first Ctrl-C cancels the session token — the flows unwind at their checkpoints and run their own unlock discipline, exactly like a cooperative MCP `job_cancel`; a second Ctrl-C (or a short deadline) may hard-exit, accepting the stranded-lock outcome the first stage exists to avoid. `request_review`'s and `regenerate`'s bespoke hooks could then fold into the general mechanism.

## Acceptance

- [ ] Ctrl-C during `update`/`prepare`/`install`/`run` stops the flow at a checkpoint, runs the flow's unlock path, and returns to the prompt with a truthful "cancelled" report
- [ ] A test drives the cooperative path end-to-end (the seam + checkpoints already exist and are tested; the new part is the signal wiring)
- [ ] Double-Ctrl-C (or timeout) behaviour documented; the stranded-lock consequence named in the docs if hard-exit is kept
- [ ] `request_review`/`regenerate` bespoke hooks reconciled with the general handler

plusky:
Closes #441.

## The bug

reedline holds raw mode only inside `read_line`, so a Ctrl-C at the prompt is a key event — but during a dispatched command the terminal is cooked, the key is a real SIGINT, and the default disposition kills mtui outright. The flows' "unlock always" sections never run, and unlike #405's in-process strand, the leftover `/var/lock/mtui.lock` carries a **dead pid**: it reads as foreign to every later run, and only `unlock --force` or the 24h stale reap clears it. `quit`'s teardown (pool claims, host close) is skipped the same way.

A second, related defect fixed along the way: `request_review --watch` listened on `tokio::signal::ctrl_c()` directly, which arms tokio's process-permanent SIGINT handler as a side effect — so after one watch, a Ctrl-C during any later command was **silently swallowed** instead of doing anything. Ctrl-C behavior was non-deterministic across a session.

## The fix

First press cancels, second press force-quits — driving the cooperative seam that already exists (the same one MCP `job_cancel` uses):

- A forwarder task (one persistent SIGINT stream, armed when the REPL starts) feeds presses into a bounded channel. The dispatch site drains stale presses (a press from an `edit`/idle gap must not cancel the *next* command — and with the two-slot queue, a leftover press would otherwise consume the cooperative slot so a genuine press force-quits), installs a **fresh `CancellationToken` per dispatch** (the token is one-shot; without the per-line install, a cancelled token would fail every later command at the driver pre-flight — including `quit`'s teardown, re-stranding exactly these locks), and runs the dispatch under a **biased select that never drops the command future**: a press cancels the token and reports it; the flows unwind at their own checkpoints and release their locks; a second press escalates.
- The select is biased with the completion branch first, so a press racing completion can never force-quit a command that already finished (adversarial review measured the unbiased version doing exactly that ~50% of the time, with a false "locks may remain" warning on top).
- Escalation propagates out of the loop; `main` drops the editor first — **the session's shell history survives a force-quit** — then exits 130 with a warning naming `unlock --force`/`--pool` from a new session.
- Ctrl-D's teardown gets the same protocol minus the cancel: a press never cancels the teardown (it *is* the cleanup), but still counts toward force-quit — review found that with the handler armed, a teardown wedged on a black-holed host previously left the operator with no exit at all.
- The bespoke waits fold onto the seam: `sleep_or_interrupt` selects the session token (in the inter-poll sleep **and** the 429 backoff, which could block a cancel up to the 60s retry clamp), and `regenerate`'s should-stop predicate — previously wired to a spinner flag nothing ever set, making its Ctrl-C claim dead code — reads the live token. Bonus this buys: **MCP `job_cancel` now interrupts a running watch/regenerate wait at the poll boundary**, which the signal-only implementation never could.

## Honesty about what the first press does per command

Cancellation is cooperative and the flows differ (documented in the new FAQ entry): `update`, `prepare --installed-only`, non-transactional `downgrade`, and multi-template fan-outs stop at their checkpoints with `error: cancelled` and locks released (`update`'s last checkpoint sits before the point of no return; a press after dispatch is deliberately inert). `install`/`uninstall`/`run`/`reboot` and transactional downgrades have no mid-op checkpoint by design — they finish the current host operation, release locks normally, and report their result. The folded waits stop and report success in their own words.

**Known remaining window, stated rather than papered over:** the startup seeding load (`-a`/`--sut` before the first prompt — the slowest phase in the program) still has the old kill-with-strand behavior; arming the handler there without a consumer would make Ctrl-C a silent no-op for a minute, which is worse. Routing seeding through the same drain/token protocol is the follow-up.

## Tests

Twenty-one mutation proofs across two rounds, each observed red with the test demonstrably running. The adversarial round (three independent reviewers over the uncommitted tree) reshaped the change: the biased select, the teardown escalate arm, the history-preserving exit path, and five surviving mutants killed — including a drain `while`→`if` mutant whose real-world outcome (a stale press consuming the cooperative slot) would have been strictly worse than the original bug. Two proofs are documented as bounded rather than absolute: the biased-select pin is statistical (32 independent races per run; the unbiased mutant failed both ways when measured), and the one-line Ctrl-D arm wiring is unreachable from tests (TTY-bound) — narrowed structurally by the type change instead, and said so. No test raises a real signal; the interrupt source is channel-injected, and the thin `tokio::signal` forwarder plus `process::exit(130)` stay untested by design, per the repo's established pattern. Stress: 40 consecutive suite runs (~1280 select races), zero flakes.

## Residuals (documented)

- The startup seeding window (above — follow-up sketch in the impl notes).
- `release_pool_claims` has no timeout of its own (pre-existing; `CLOSE_TIMEOUT` wraps only the host close) — the teardown escalate arm is the operator's exit meanwhile.
- The first-press notice and the escalation warning are `tracing::warn!` — suppressible by `set_log_level error`, matching the `warn_on_unlock_failures` precedent; a conscious choice, noted.
- Non-active template groups can retain a stale cancelled token after a cancelled line (nothing reads it today; `activate` refreshes the active group every dispatch).

## Process

Three scouts (REPL/signal internals incl. vendored reedline/crossterm/tokio semantics, the per-command cancel-coverage map, the signal-test feasibility study), implementation, then three independent read-only adversarial reviews (signal semantics, test vacuity, contracts) over the uncommitted tree — thirteen findings, all fixed before the first commit existed.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -229,9 +229,12 @@ next actionable task before working on a subsystem.
   It installs a **fresh token per dispatched line, unconditionally** — the token
   is one-shot, so a cancelled one left in place would kill every later dispatch
   at the pre-flight check, `quit`'s teardown included, stranding exactly the
-  locks the cancel had to release. The Ctrl-D teardown gets the fresh token but
+  locks the cancel had to release. The teardown dispatch gets the fresh token but
   **no cancel arm**: a press there only escalates, because cancelling the
-  cleanup is what strands the locks. A new interrupt hook belongs on this seam,
+  cleanup is what strands the locks. That is a property of the *dispatch*, not of
+  the key — a typed `quit`/`exit` is routed to it by resolving the line's command
+  position through the registry, so it is protected exactly as Ctrl-D is.
+  A new interrupt hook belongs on this seam,
   never on its own `tokio::signal::ctrl_c` — the REPL arms SIGINT process-wide
   from the first prompt onward (startup seeding is still outside that window),
   and a headless tool call has no terminal to press Ctrl-C at, while a stdio
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -221,9 +221,27 @@ next actionable task before working on a subsystem.
   into its function's normal fall-through, never early-`return` past cleanup —
   and a real failure collected before the cancel always outranks it. MCP `job_cancel` cancels the job's token, waits a short grace for a
   cooperative stop, then hard-aborts the worker — its reply distinguishes the
-  two and never claims to cancel a job that had already finished. A flow that
+  two and never claims to cancel a job that had already finished. The REPL is
+  the seam's second producer: a Ctrl-C *during* a command (the terminal is
+  cooked then, so it is a real SIGINT rather than reedline's key event) is
+  forwarded onto the session token instead of killing the process, and a second
+  press force-exits 130 with a warning about the locks that may be left behind.
+  It installs a **fresh token per dispatched line, unconditionally** — the token
+  is one-shot, so a cancelled one left in place would kill every later dispatch
+  at the pre-flight check, `quit`'s teardown included, stranding exactly the
+  locks the cancel had to release. The Ctrl-D teardown gets the fresh token but
+  **no cancel arm**: a press there only escalates, because cancelling the
+  cleanup is what strands the locks. A new interrupt hook belongs on this seam,
+  never on its own `tokio::signal::ctrl_c` — the REPL arms SIGINT process-wide
+  from the first prompt onward (startup seeding is still outside that window),
+  and a headless tool call has no terminal to press Ctrl-C at, while a stdio
+  server that *does* share one would fire every listener at once, interrupting
+  work nobody asked to stop. A flow that
   stops on a cancel must surface as `CommandError::Cancelled`, not a generic
-  failure — and that verdict must come from the flow's own `cancelled` flag,
+  failure — unless stopping *is* its documented success (`request_review
+  --watch` returns `Ok` with "the request is still posted"; `regenerate`
+  abandons the wait and reports the state it last saw, since the server keeps
+  building) — and that verdict must come from the flow's own `cancelled` flag,
   never from sniffing the session token, which would mask a real host failure
   that merely coincided with a cancel (see
   `commands/perform.rs::map_flow_error`). New long-running command bodies
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -233,11 +233,12 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   fan-outs stop at their next checkpoint with their locks released, while
   `install`/`uninstall`/`run` finish the host operation already under way — and
   the session stays usable afterwards. A second press still force-quits (exit
-  130), now with a warning naming `unlock --force` as the remedy for whatever
-  locks it strands — and without discarding the session's command history, as
-  the old kill did. During the Ctrl-D/`quit` teardown a press cannot cancel
-  anything (the teardown is what releases the pool claims), but two presses
-  still force-quit, so a blackholed refhost can no longer wedge the exit.
+  130), now naming `unlock --force` as the remedy for whatever locks it strands
+  — and without discarding the session's command history, as the old kill did.
+  During the session teardown (Ctrl-D or a typed `quit`/`exit`) a press cannot
+  cancel anything, because the teardown is what releases the pool claims; two
+  presses still force-quit, so a blackholed refhost can no longer wedge the
+  exit, and that message names `unlock --pool` as well.
   Ctrl-C at the prompt is unchanged (it clears the line), and Ctrl-C during the
   initial `-a`/`-k`/`--sut` load still exits immediately without teardown.
   Two existing waits are folded onto the same seam: `request_review --watch`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -226,6 +226,27 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   loaded template. A *cooperative* cancel is unchanged: a body that unwound
   through its own flow ran its own unlock discipline. Nothing else is done at
   the hosts; as before, the remote command may still be running there.
+- Ctrl-C during a running command no longer kills the REPL outright (#441).
+  Killing it skipped every teardown, so each locked host was left with a
+  dead-pid `/var/lock/mtui.lock` that blocked the next tester. The first press
+  now cancels the command cooperatively — `update`, `prepare`, `downgrade`, and
+  fan-outs stop at their next checkpoint with their locks released, while
+  `install`/`uninstall`/`run` finish the host operation already under way — and
+  the session stays usable afterwards. A second press still force-quits (exit
+  130), now with a warning naming `unlock --force` as the remedy for whatever
+  locks it strands — and without discarding the session's command history, as
+  the old kill did. During the Ctrl-D/`quit` teardown a press cannot cancel
+  anything (the teardown is what releases the pool claims), but two presses
+  still force-quit, so a blackholed refhost can no longer wedge the exit.
+  Ctrl-C at the prompt is unchanged (it clears the line), and Ctrl-C during the
+  initial `-a`/`-k`/`--sut` load still exits immediately without teardown.
+  Two existing waits are folded onto the same seam: `request_review --watch`
+  stops on a cancel (and now also on an MCP `job_cancel`, which could never
+  interrupt it before), and `regenerate`'s wait for TeReGen finally honours the
+  cooperative-stop hook it always advertised. This also removes a
+  non-determinism: `request_review --watch` used to arm the process-wide SIGINT
+  handler as a side effect, after which Ctrl-C was silently swallowed for the
+  rest of the session.
 
 ### Security
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2374,6 +2374,7 @@ dependencies = [
  "shlex",
  "tempfile",
  "tokio",
+ "tokio-util",
  "tracing",
  "tracing-subscriber",
 ]
--- a/crates/mtui-cli/Cargo.toml
+++ b/crates/mtui-cli/Cargo.toml
@@ -42,6 +42,9 @@ mtui-types.workspace = true
 # startup seeding path (before the REPL loop begins).
 mtui-testreport.workspace = true
 tokio.workspace = true
+# `CancellationToken`: the REPL installs a fresh one per dispatched line so a
+# mid-command Ctrl-C cancels *that* command's flow (and only it).
+tokio-util.workspace = true
 tracing.workspace = true
 tracing-subscriber.workspace = true
 # Optional desktop-notification backend (feature `notify`); a headless no-op
--- a/crates/mtui-cli/src/lib.rs
+++ b/crates/mtui-cli/src/lib.rs
@@ -19,7 +19,7 @@ pub mod startup;
 
 pub use notification::notify_user;
 pub use prompt::MtuiPrompt;
-pub use repl::Repl;
+pub use repl::{Repl, ReplExit};
 pub use startup::seed_session;
 
 use std::io::Write;
--- a/crates/mtui-cli/src/main.rs
+++ b/crates/mtui-cli/src/main.rs
@@ -91,5 +91,16 @@ fn main() -> anyhow::Result<()> {
     let session = Arc::new(Mutex::new(session));
     let mut repl = Repl::new(registry, session);
 
-    runtime.block_on(repl.run())
+    let ending = runtime.block_on(repl.run())?;
+    // Drop the editor before honouring a force-quit: reedline persists its
+    // `FileBackedHistory` when it is dropped, and `std::process::exit` runs no
+    // destructors — exiting straight from the loop would throw the session's
+    // command history away on every double Ctrl-C. The exit itself stays here,
+    // thin and untestable, while the *decision* is made (and tested) in the
+    // loop.
+    drop(repl);
+    if let Some(code) = ending.exit_code() {
+        std::process::exit(code);
+    }
+    Ok(())
 }
--- a/crates/mtui-cli/src/main.rs
+++ b/crates/mtui-cli/src/main.rs
@@ -92,15 +92,19 @@ fn main() -> anyhow::Result<()> {
     let mut repl = Repl::new(registry, session);
 
     let ending = runtime.block_on(repl.run())?;
-    // Drop the editor before honouring a force-quit: reedline persists its
-    // `FileBackedHistory` when it is dropped, and `std::process::exit` runs no
-    // destructors — exiting straight from the loop would throw the session's
-    // command history away on every double Ctrl-C. The exit itself stays here,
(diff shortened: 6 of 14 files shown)


mimi1vx (changes requested):
Reviewed against a local worktree of the PR head (`b3878cc1`).

**Gates run locally, all green:** `cargo clippy --workspace --all-targets -- -D warnings`, `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features --document-private-items`, `cargo test -p mtui-cli -p mtui-core` (770 tests).

The core design is right: cooperative-first, biased select, fresh token per dispatch, dispatch future never dropped on the first press. I verified the parts that are easy to get wrong and they hold (details at the bottom). Three things need fixing before merge — one behavioural, two that record load-bearing constraints the code currently leaves undocumented or contradicted.

---

## Critical (must fix)

### 1. The teardown protection only covers Ctrl-D, not a typed `quit`/`exit` — and three documents claim otherwise

`step_teardown` (fresh token, escalate-only, *no cancel arm*) is reachable **only** from `Signal::CtrlD` (`crates/mtui-cli/src/repl.rs:287`). A typed `quit`, `exit`, or `exit reboot` is `Signal::Success(line)` (`repl.rs:208`) and goes through `step_interruptible`, which **does** `token.cancel()` on the first press and, on the second, emits the *mid-command* warning (`repl.rs:265`):

> `forcing exit mid-command; operation locks may remain on the update's hosts — release them with `unlock --force` from a new session`

In the exact scenario the escalate arm was added for — a blackholed refhost parking `release_pool_claims`, which has no timeout of its own — an operator who *typed* `quit` instead of pressing Ctrl-D force-quits and is told to run `unlock --force`, with **no mention of `unlock --pool`**. The RRID-based pool claim is genuinely stranded on the fleet and the message points nowhere near it.

The cancel itself is inert today (`Quit::call` never consults the seam, and neither `release_pool_claims` nor `HostsGroup::close` gate on it), so this is wrong operator guidance plus a latent contract hole rather than an immediate strand — but it *is* a hole: the moment anything on the close path gains a checkpoint, a typed `quit` cancels its own cleanup, which is precisely #441.

Three documents assert the fixed behaviour:

- `CHANGELOG.md:218` — "During the Ctrl-D/`quit` teardown a press cannot cancel anything"
- `docs/src/faq.md:208` — "Ctrl-C during the **Ctrl-D/`quit` teardown** behaves the same way but cancels nothing"
- `AGENTS.md:233` says "Ctrl-D teardown", which *is* accurate — so the contract file and the user-facing docs now disagree with each other.

Suggested fix, matching the existing `is_shell_line` / `is_edit_line` precedent already in this loop: resolve the line's first token through the registry and route a `Quit` hit to `step_teardown`; or give `step_interruptible` a `cancel_on_press: bool` (see Warning 3, which wants the same seam). Please add a test that a typed `quit` never observes a cancelled token — `a_press_during_the_teardown_does_not_cancel_it` mirrored onto `step_interruptible` with `SlowQuitCmd` is a two-line variant, and it should be observed red first.

### 2. `main.rs` — the reason this cannot become `fn main() -> ExitCode` is undocumented, and it is a deadlock rather than a style preference

The obvious later "modernise the binary" cleanup is:

```rust
fn main() -> anyhow::Result<ExitCode> { …; Ok(ending.into()) }
```

which runs destructors and lets the `drop(repl)` dance be deleted. **It would hang the force-quit.** Returning from `main` drops `runtime`, and `Runtime::drop` blocks until in-flight `spawn_blocking` tasks complete. There is one live on exactly this path: `Prompter::stdin()` (`crates/mtui-hosts/src/prompter.rs:73-90`), installed at the composition root (`crates/mtui-cli/src/main.rs:78`), performs a blocking `io::stdin().read_line()` on the blocking pool to back the SSH command-timeout question. If the operator double-presses Ctrl-C while `keep waiting? [Y/n]` is outstanding — the single most likely moment to force-quit — that read never returns, so `Runtime::drop` never returns. `std::process::exit(130)` is what escapes it.

The comment at `main.rs:95-100` justifies only `drop(repl)` (the history flush) and says nothing about *why `process::exit` rather than `ExitCode`*. By this repo's own rule — a load-bearing shape must not be left explained by a comment about something else — this constraint needs recording, or the next cleanup pass silently reintroduces a hang in the escape hatch. One sentence is enough:

> `process::exit`, not `main() -> ExitCode`: returning would drop the runtime, which blocks on in-flight `spawn_blocking` — including the stdin prompter that is very likely outstanding when someone force-quits.

### 3. A second exit-code vocabulary, and the documented exit-code contract left stale

`mtui_core::ExitStatus` (`crates/mtui-core/src/entrypoint.rs:42`) already exists, is exported, and its module doc carries a `## Exit-code contract` section presenting itself as the authority (`Ok=0`, `Failure=1`, `Usage=2`, with `From<ExitStatus> for i32`). It is consumed by `seed_session` (`ControlFlow<ExitStatus>`) and by `mtui-mcp/src/main.rs:28`.

After this PR `main` speaks two exit vocabularies nineteen lines apart:

```rust
std::process::exit(code.into());  // main.rs:85  — ExitStatus
std::process::exit(code);         // main.rs:104 — ReplExit / EXIT_INTERRUPTED
```

and `entrypoint.rs`'s contract doc never learns that 130 exists. A reader of the file that claims to define mtui's exit codes concludes mtui only ever exits 0/1/2 — which packaging, wrapper scripts, and CI will act on.

Preferred fix: add `ExitStatus::Interrupted => 130` and map `ReplExit::ForceQuit` onto it. 130 is a process-level convention, not a REPL-internal detail, and it belongs in the one place downstream will look (`ExitStatus`'s own doc line — "the process exit status a single non-interactive command run yields" — needs widening if you take this route). Minimum acceptable: extend the `## Exit-code contract` section to name 130 and cross-reference `ReplExit`.

---

## Warnings (should fix)

1. **[`repl.rs:264`, `repl.rs:299`] The escalation warning is `warn!`, so `set_log_level error` silences the one message that names the stranded locks.** The PR notes the first-press *notice* is suppressible and calls that a conscious choice — agreed, that one is informational. The escalation line is a different class of event: it is the only record that the process is abandoning locks, emitted at the moment an operator most needs it. `tracing::error!` survives the level operators actually set. Same for the mid-teardown variant.

2. **[`repl.rs:433-535`] `step_interruptible` and `step_teardown` are ~90% duplicated.** Drain, fresh token, `tokio::pin!`, biased select, `is_none` fall-through, escalate-on-second — identical; the only differences are `token.cancel()` and the warning text. One function with a `cancel_on_press: bool` would also make Critical 1 a one-line call-site change instead of a third copy. Two independent copies of a subtle select protocol is the shape that drifts.

3. **[`repl.rs:253-271`, `repl.rs:287-307`, `main.rs:94-105`] The two `Repl::run` escalate arms and the drop-then-exit ordering have no coverage.** TTY-boundness is a fair reason for the *signal source*, but the wiring here — which warning pairs with which arm, and `drop(repl)` preceding `process::exit` — is what the changelog makes a user-visible promise about ("without discarding the session's command history"). `FileBackedHistory`'s sync-on-drop makes that ordering genuinely load-bearing and nothing enforces it. Extracting `fn on_escalate(kind) -> ReplExit` would at least pin the arm→message pairing.

---

## Suggestions

- **[`main.rs:101`] `drop(repl)` drops far more than "the editor".** `Repl` is the sole owner of `Arc<Mutex<Session>>`, so this synchronously tears down the entire session graph — every SSH `Target`, the reqwest client, the openQA transport, the template registry — before `process::exit`, on the one path whose whole job is *get out now*. I checked: nothing in that chain blocks today (no `Drop` on `Session`/`Target`; `SpinnerGuard::drop` / `TtySpinner::drop` only `handle.abort()` plus std-mutex writes). But the blast radius is wider than the comment claims. What you actually want is "run exactly one destructor, then exit", which neither `ExitCode` nor `process::exit` expresses — worth naming, and worth narrowing (`Repl::into_history()`, or dropping only `line_editor`) so this path cannot grow a blocking destructor by accident.

- **[`repl.rs:68`, `repl.rs:90`] `i32` is the wrong domain for an exit code.** `WEXITSTATUS` is 8 bits, so `process::exit` truncates mod 256; `EXIT_INTERRUPTED: i32` / `exit_code() -> Option<i32>` admits values that silently change meaning. 130 is fine and `ExitStatus` has the same shape, so consistency partly excuses it — but if Critical 3 lands as `ExitStatus::Interrupted`, make the domain `u8` (or `std::process::ExitCode` right at the boundary, which is that type's one genuinely good use here even though it cannot be returned from `main`).

- **MCP semantics of "cancel is success."** Folding both waits onto the seam means a `job_cancel` on a `request_review --watch` or `regenerate` job now ends with the command returning `Ok`, i.e. `job_status` → `done`, not `cancelled`. AGENTS.md was amended to carve this out and that is defensible for the REPL, but a client that just called `job_cancel` and is told the job *succeeded* is a surprising surface. Worth a sentence beside the `job_cancel` bullet in `docs/src/mcp.md:293`.

- **[`request_review.rs:403`] The watch banner still reads `"watching for reactions (up to {}s, Ctrl-C to stop)"`.** Over MCP there is no Ctrl-C, and `job_cancel` now works — cheap to say "Ctrl-C / job_cancel".

- **[`repl.rs:345-388`] A non-TTY REPL now swallows SIGINT while idle.** With the handler armed from the first prompt, a piped session (no raw mode, so Ctrl-C at the "prompt" *is* a real SIGINT) queues the press and then discards it at the next drain — the process becomes un-interruptible by Ctrl-C between commands. Narrow, and arguably out of scope, but it is a behaviour change the FAQ's "Ctrl-C at the prompt is unchanged" does not cover.

---

## Verified sound

For the record, the parts I specifically tried to break and could not:

- The `biased;` ordering and its 32-round race pin — I traced the poll interleaving in `a_press_racing_completion_never_force_quits`; the presses really are queued in the same poll window as the completion, so the test would be a coin flip without `biased`.
- Unconditional fresh-token-per-dispatch against `Command::run`'s pre-flight `check_cancelled`: `run` calls `session.activate(...)` on **both** the single-template and fan-out paths, which re-pushes the fresh token onto the active `HostsGroup`, so `update_flow`'s five `targets.cancel_requested()` checkpoints observe the right token. The stale-token residual on non-active groups is genuinely inert — `Quit`, `close`, and `release_pool_claims` never read the seam, and `run_parallel` deliberately doesn't.
- `edit` and `shell` are intercepted *before* dispatch, so the "idle gap" drain rationale is accurate and neither can be cancelled mid-editor. tokio's `sigaction` handler is reset on `exec`, so child editors/shells keep default SIGINT; the `shell` bridge holds local raw mode, so Ctrl-C there is a byte to the remote PTY, not a signal.
- Drain loop terminates on both `Empty` and `Disconnected`; `mpsc::Receiver::recv` is cancel-safe in the select; the `press.is_none()` fall-through is unreachable-but-correct.
- `drop(repl)` before `process::exit` is genuinely required (reedline syncs `FileBackedHistory` on drop) and `runtime` correctly outlives it.
- No secrets reach the display or the log on any new path.

plusky:
All addressed in `7708d6a2`.

- **C1+W2 folded together**: `step_teardown` is gone — one `step_interruptible(..., OnPress::{Cancel, EscalateOnly})`; a typed `quit`/`exit`/`EOF` resolves its first token through the registry (the `is_shell_line` precedent) and gets the no-cancel protocol with the `--force`/`--pool` warning. `help quit`/`echo quit` don't match. Your `SlowQuitCmd` test exists and was observed red against the pre-review routing; the three docs now say "Ctrl-D **or a typed `quit`/`exit`**" truthfully.
- **C2**: your sentence, verbatim in spirit, at the exit site.
- **C3+S2**: `ExitStatus::Interrupted` (130) with the contract doc extended and `ReplExit` mapping through the one `From<ExitStatus> for i32`; `i32` kept for consistency with the existing conversion.
- **W1**: both force-quit records are `error!` (a test pins that they outrank `set_log_level error`); the first-press notice stays `warn` per your split.
- **W3**: `on_escalate(OnPress) -> ReplExit` extracted; a test pins the arm→message pairing.
- **S1**: `drop(repl.into_line_editor())` — exactly one destructor (the history sync) runs before `process::exit`; the session graph is deliberately leaked on the get-out-now path.
- **S3/S4**: mcp.md notes a cancelled watch/regenerate job ends `done`, not `cancelled`, and why; the banner and the `--watch` clap help both say "Ctrl-C / job_cancel" (hence one regenerated `cli.md` line + the matching schema-snapshot line).
- **S5**: documented in the FAQ (piped sessions queue-and-discard a between-commands Ctrl-C; `kill -INT` is the escape) and listed as a residual rather than changed.

Four new mutation proofs cover the routing, the token-position match, the pairing, and the `error!` level.

mimi1vx (approved):

PR #446, opened 2026-08-11
fix(testreport): report the patch's own exit status from the update templates

### LINKED ISSUE #400: update templates discard the patch command's exit code, so a failed patch with clean output passes

plusky:
Closes #400.

## The bug

`ZYPPER_UPDATE` and `SLM_UPDATE` are multi-line scripts run as one remote
command. Their last line is the repo-cleanup loop, whose status is `0` whenever
it finds nothing to remove — so the shell reported that, and the patch's status
was discarded before anything could read it. A patch that genuinely failed with
clean output passed the update check. (The post-state `zypper patches | grep`
clobbers `$?` even before the loop does.)

This is long-standing upstream behaviour, not a port regression: the Python
templates are byte-identical, trailing loop and all.

## The fix

Each template captures the patch's status on the line *after* the patch and
exits with it. The patch list is captured first and the patch runs only when
that list holds at least one word, so a host carrying none of the update's
products is a no-op rather than a package manager complaining about its own
arguments — a status that is now real.

The emptiness test is `set -- $mtui_patches` + `[ "$#" -gt 0 ]`, **not**
`[ -n "$mtui_patches" ]`: the latter is true for a list of whitespace, which
then word-splits to nothing and runs the patch with no operands. That was caught
in review, measured, and is covered by a test.

## The safety argument — please read this part

An `update` check failure drives the **group-wide** rollback downgrade: it is
handed the whole `HostsGroup` with no filtering, removes every host's issue
repos, downgrades every host, reboots the transactional ones, and rewrites the
report's version slots. A false failure therefore reverts hosts that did
everything right. So the check classifies the status instead of comparing it
against zero:

| codes | verdict |
|---|---|
| `0`, `100`, `101`, `102`, `103`, `106`, `107` | pass |
| `104`, `4`, `5`, `8` | "package not found" |
| any other non-zero | "Unknown Error", *unless* an output marker names it more precisely |

`102` is "reboot needed" — the routine outcome of patching a kernel.

**`107` was missing from that set in my first draft, and the review caught it.**
It is `ZYPPER_EXIT_INF_RPM_SCRIPT_FAILED`: a package's `%post` script failed
although the package itself was, per the man page, "successfully unpacked to
disk and registered in the rpm database". It would have failed the check and
rolled back the fleet over a scriptlet hiccup. The second commit fixes the
identical omission in the *install* check, where it is a live bug today.

The template change and the check change are one commit deliberately: a template
reporting the real status while the check still lacked the carve-out would newly
fail every host that merely needs a reboot.

## Two things I want your call on

1. **`4`/`5`/`8` report "package not found".** That grouping is inherited
   verbatim from the install check, and the reason strings are a stable contract,
   so I kept it rather than let one exit code carry two verdicts across two
   checks. But it is now on the *primary* failure path instead of dormant, and
   the label misdiagnoses: `8` is `ERR_COMMIT` (the most likely status for the
   failure this issue is about) and `5` is `ERR_PRIVILEGES`. Splitting them means
   minting a reason string. Happy to do it if you'd rather.
2. **`103` passes.** `RESTART_NEEDED` means zypper installed the patch that
   updates the package manager itself and *remaining* patches need a second run —
   so mtui reports the update applied while some patches were not installed.
   Inherited from the install check; the post-patch `zypper patches` line in the
   transcript is what shows them. Flagging because this change advertises it.

## SL Micro

The classification is in practice just "`0` passes". Upstream
`transactional-update` captures zypper's status, tolerance-tests it, and flattens
everything else to `EXITCODE=1` — it returns only `0` or `1`, verified against
`openSUSE/transactional-update` @ `aee1e1b5` (v6.1.3) and unchanged across nine
tags. So neither the informational codes nor the package-not-found ones can reach
the check on that key. This started as an inference and was checked rather than
assumed; the doc says so with the citation.

## Testing

The acceptance test renders each template and executes it under a real `/bin/sh`
with stub `zypper` / `transactional-update` executables on `PATH`, because the
suite's `MockConnection` scripts one exit code per command and cannot model the
shell's last-command-wins rule — the entire subject of this issue. Each case was
observed red against the old template first:

* patch exits 8, cleanup exits 0 → script must exit 8 (**before: 0** — the bug);
* patch exits 0, cleanup exits 3 → script must exit 0 (a cosmetic cleanup hiccup
  must not fail an update and revert the group);
* empty and whitespace-only patch lists → patch never invoked, script exits 0.

The stubs record their invocations so the tests cannot pass by never running
anything. A separate test asserts the fixture rows match under *this host's* awk,
so a non-GNU awk fails loudly with the real cause rather than silently skipping —
the template's `\>` is a gawk extension and the macOS CI leg runs these tests.

## Residuals, deliberately not fixed here

* The command substitution's own failure is discarded: a failed `zypper refresh`
  or a missing repo yields an empty patch list, skips the patch, and reports
  success having installed nothing. Same class as this bug, one step upstream —
  filed separately rather than widening a change that already moves the rollback
  boundary.
* `#[cfg(unix)]` on the shell tests: they vanish on a non-unix builder with a
  green `ok`.
* `perform_update_aggregates_multiple_host_failures_and_keeps_repos` still
  scripts its exit code via `with_default`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -144,13 +144,40 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   failed transaction. As on every other host, a failed update check on these
   also drives the group-wide rollback downgrade, which they could not reach
   before.
-  SL Micro is judged on the command's output markers rather than its exit code:
-  its update template ends with a repo-cleanup loop, so the status the shell
-  reports is that loop's and not the patch's. RHEL/YUM is judged only on
-  whether the command ran: that key covers every RHEL version, `yum` is `dnf`
-  on RHEL 8/9, and mtui passes the whole package list to hosts that carry only
-  a subset — so neither a non-zero exit nor an `Error:` line (the template also
-  runs `yum repolist` in the same shell) reliably means the update failed.
+  SL Micro is judged on the command's output markers and on the patch's exit
+  code. RHEL/YUM is judged only
+  on whether the command ran: that key covers every RHEL version, `yum` is
+  `dnf` on RHEL 8/9, and mtui passes the whole package list to hosts that carry
+  only a subset — so neither a non-zero exit nor an `Error:` line (the template
+  also runs `yum repolist` in the same shell) reliably means the update failed.
+- An `update` whose patch command fails now fails the update, on both zypper
+  and SL Micro hosts, even when the patch's output is clean. The two update
+  templates are multi-line scripts run as a single remote command, and their
+  last line was the repo-cleanup loop — whose status is `0` whenever it finds
+  nothing to remove — so the shell reported that and discarded the
+  `zypper -n in -t patch` / `transactional-update -n pkg in -t patch` status
+  before anything could read it. Each template now captures the patch's own
+  status and exits with it, and the remote command text visible in `show_log`
+  transcripts changes accordingly.
+  The exit code is classified rather than compared against zero. `100`-`103`,
+  `106` and `107` are zypper's *informational* results and still pass — `102`
+  is "reboot needed", the routine outcome of patching a kernel, and `107` means
+  a package's `%post` script failed although the package itself is installed
+  and registered. (`103`, "restart needed", also passes: zypper installed the
+  patch that updates the package manager itself, and any *remaining* patches
+  need a second run — the post-patch `zypper patches` line in the transcript is
+  what shows those.) `104`, `4`, `5` and `8` report "package not found", ahead
+  of the output markers. Any other non-zero reports "Unknown Error", but only
+  when the markers recognise nothing — a locked stack, a dependency problem or
+  an RPM error keeps its own, more precise reason.
+  Without the informational carve-out a host that patched perfectly would fail
+  its check and fire the group-wide rollback downgrade, reverting every healthy
+  host in the group. A host carrying none of the update's products still
+  passes: the patch command is skipped rather than run with no operands.
+  On SL Micro the classification is in practice just "`0` passes": upstream
+  `transactional-update` absorbs zypper's status and returns only `0` or `1`,
+  so neither the informational codes nor the package-not-found ones can reach
+  the check on that key.
 - An `update` that timed out, or whose connection dropped part-way, is no
   longer reported as successful on **any** host. The update check had no
   equivalent of the downgrade check's "timed out or failed to run" gate, so a
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -145,11 +145,14 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   also drives the group-wide rollback downgrade, which they could not reach
   before.
   SL Micro is judged on the command's output markers and on the patch's exit
-  code. RHEL/YUM is judged only
-  on whether the command ran: that key covers every RHEL version, `yum` is
-  `dnf` on RHEL 8/9, and mtui passes the whole package list to hosts that carry
-  only a subset — so neither a non-zero exit nor an `Error:` line (the template
-  also runs `yum repolist` in the same shell) reliably means the update failed.
+  code — and because `transactional-update` absorbs zypper's status and reports
+  a flat `1`, that key now fails on any non-zero exit, including failures that
+  are not the patch's own (an already-open transaction, a snapshot that could
+  not be created). RHEL/YUM is judged only on whether the command ran: that key
+  covers every RHEL version, `yum` is `dnf` on RHEL 8/9, and mtui passes the
+  whole package list to hosts that carry only a subset — so neither a non-zero
+  exit nor an `Error:` line (the template also runs `yum repolist` in the same
+  shell) reliably means the update failed.
 - An `update` whose patch command fails now fails the update, on both zypper
   and SL Micro hosts, even when the patch's output is clean. The two update
   templates are multi-line scripts run as a single remote command, and their
@@ -166,10 +169,11 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   and registered. (`103`, "restart needed", also passes: zypper installed the
   patch that updates the package manager itself, and any *remaining* patches
   need a second run — the post-patch `zypper patches` line in the transcript is
-  what shows those.) `104`, `4`, `5` and `8` report "package not found", ahead
-  of the output markers. Any other non-zero reports "Unknown Error", but only
-  when the markers recognise nothing — a locked stack, a dependency problem or
-  an RPM error keeps its own, more precise reason.
+  what shows those.) `104` reports "package not found" ahead of the output
+  markers. Every other failing code — `4`, `5`, `8` and anything else non-zero
+  — lets the markers speak first, so a locked stack, a dependency problem or an
+  RPM error keeps its own, more precise reason; with nothing recognised in the
+  output, `4`/`5`/`8` report "package not found" and the rest "Unknown Error".
   Without the informational carve-out a host that patched perfectly would fail
   its check and fire the group-wide rollback downgrade, reverting every healthy
   host in the group. A host carrying none of the update's products still
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -178,6 +178,12 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   `transactional-update` absorbs zypper's status and returns only `0` or `1`,
   so neither the informational codes nor the package-not-found ones can reach
   the check on that key.
+- `install` no longer fails when a package's `%post` script did. zypper exits
+  `107` (`ZYPPER_EXIT_INF_RPM_SCRIPT_FAILED`) in that case — an *informational*
+  code meaning the packages "were successfully unpacked to disk and are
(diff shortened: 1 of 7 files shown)


mimi1vx (changes requested):
Reviewed the full files (not just the diff) and ran the local gate — `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `RUSTDOCFLAGS="-D warnings" cargo doc ...` and `cargo test -p mtui-testreport` are all green.

The core template fix is correct: `mtui_status=$?` sits immediately after the patch invocation, `set -- $mtui_patches` + `[ "$#" -gt 0 ]` is the right emptiness test (field splitting rather than `-n`), the `$$`-escaping renders correctly under `SubstMode::Safe`, and the trailing `exit` is safe because `Connection::run` does a bare `channel.exec` with nothing wrapped or appended. `run_checks` reads `lastexit()` before `reboot_transactional`, and `add_op_history` uses `sftp_append`, so the snapshot the check inspects really is the patch script's own.

Requesting changes on one point below.

## Blocking

### `checks/update.rs:160-173` — `PackageNotFound` short-circuiting `markers` degrades the exact failure this PR exists to surface

With the new short-circuit, on the zypper update check:

- exit `8` (`ZYPPER_EXIT_ERR_COMMIT` — an rpm transaction failure, the most likely status for a genuinely failed patch) with `Error:` on stderr now reports **"package not found"** instead of **"RPM Error"**;
- exit `5` (`ERR_PRIVILEGES`) carrying `System management is locked` now reports **"package not found"** instead of **"update stack locked"**.

Both are the headline diagnoses for the two most common real-world patch failures, and both are now replaced by a message that is actively misleading — the package was found, the transaction failed.

`package_not_found_outranks_the_markers` pins this deliberately, but the stated justification (the reason strings are a frozen contract shared with the install check) does not hold on the *update* path:

- Nothing branches on the string. `rg '"package not found"'` matches only the check modules and their own tests — it is user-facing text, not a machine-readable contract.
- On `update`, exits `4`/`5`/`8` previously produced **no** verdict at all: they fell through to `markers` and passed. There is no prior behaviour to stay compatible with, so no consumer can be relying on `8 -> "package not found"` from an update.

A minimal fix that preserves `104`'s short-circuit exactly and gives the other three a real diagnosis:

```rust
ExitClass::PackageNotFound if args.exitcode == 104 => { log_failed(args); Err(...) }
ExitClass::PackageNotFound => {
    markers(args)?;
    log_failed(args);
    Err(UpdateError::new("package not found", ...))
}
```

The install check can keep its inline ordering unchanged.

## Non-blocking

### `actions/update.rs:83-89` — a failed `zypper -n refresh` now yields a clean `exit 0` with nothing installed

Scenario: the issue repo is unreachable or its GPG key is untrusted -> `refresh` fails -> `zypper -n patches` returns no matching row -> the guard skips the patch -> `mtui_status` stays `0` -> the update check passes.

The verdict is unchanged from before this PR (the old `zypper in` with no operands exited 3, which was masked, and "Required argument missing." carries no marker), so it is not a regression. But it is now the only path that reports success without patching, on a check this PR advertises as exit-code-faithful. The PR body files it as a residual — noting it here so the deferral is explicit in review as well.

### `checks/update.rs:122-125` — slmicro: any non-zero `transactional-update` status now triggers the group-wide rollback

`transactional-update` flattens everything to `1`, including failures that are not the patch's (an open transaction, snapshot creation/deletion problems). Those are legitimate failures and the CHANGELOG documents the change, but this key went from "never fails on exit code" to "fails on any non-zero", and a check failure downgrades and reboots *every* host in the group. Worth a deliberate ack rather than an implicit one.

### `checks/mod.rs:913-919` — `classify_exit(-1)` returning `Unknown` is a footgun the doc comment can only warn about

The doc says `-1` "is not classified here", but the function does classify it — as `Unknown`, i.e. "Unknown Error" -> `UpdateFailure::Check` -> a group-wide rollback on a host mtui never reached. That is precisely what `UpdateFailure::NotRun` exists to prevent, and it only holds because both current callers run `not_run` first. An `ExitClass::NotRun` variant would turn the omission into a compile error at the next call site instead of a comment someone has to read.

### `CHANGELOG.md:16-18` — reflow

`"...and on the patch's exit code. RHEL/YUM is judged only\n on whether the command ran"` leaves a short orphan line mid-sentence.

## Note on the tests

The `rendered_script` module is the right call here — `MockConnection` genuinely cannot express "the patch failed but the last line succeeded". The stub-invocation sentinels and `the_stub_rows_match_under_this_hosts_awk` close the two ways those tests could have been vacuous, and the awk-dialect guard passes on macOS.

plusky:
Thanks — the blocking point is right and my justification for the old ordering does not survive it. Pushed `77c2ebae` addressing all five items.

## Blocking

Applied as you suggested: `104` keeps its short-circuit, `4`/`5`/`8` run `markers(args)?` first and fall back to "package not found" only on a clean transcript. Both halves of your argument check out — a workspace-wide search for `==`/`matches!`/`contains`/`starts_with` against any of the six reason strings returns zero hits outside tests, and on `main` the update check handled only `104` and `106` before falling through, so there was no prior behaviour for the short-circuit to preserve.

`only_104_outranks_the_markers` replaces the old test. Restoring the blanket arm, deleting the arm's `markers(args)?`, or moving the guard constant each turns it red.

One thing your review implies that I want to be explicit about rather than leave to inference: **the install check now disagrees with `update` on the same transcript** — exit `5` with `System management is locked` is "update stack locked" on update and still "package not found" on install. You said install keeps its ordering, so it does; I pinned the asymmetry on the install side too. That turned out to matter: install's short-circuit was only ever exercised with an empty transcript, so hoisting its marker block above the not-found set broke no test at all. It does now.

## Non-blocking

**`classify_exit(-1)`** — added `ExitClass::NotRun`. While writing the justification I found that the footgun is narrower than either of us stated: `never_ran` in `update_flow.rs:1665` keys on the target's recorded `lastexit()`, not on anything a check returns, so a `-1` host wrapped as `UpdateFailure::NotRun` and skipped the rollback under the old fallthrough too. What the fallthrough actually broke was the *message* — an operator told "Unknown Error" about a host that was never contacted. The variant is still worth it for that and for the exhaustive match, and the docs and tests now say so instead of claiming a rollback that could not happen.

**slmicro any-non-zero** — acknowledged explicitly at the point of decision, with the blast radius and the reason it is still the better of the two options.

**`refresh` residual** — left deferred as you noted, but written into the module docs rather than living only in this conversation, including why the status cannot distinguish the two cases.

**CHANGELOG** — reflowed, and amended where the old text described the ordering this change replaces.

Also folded in three things found while re-reading: the summary paragraph on `classified` still described the behaviour you asked me to change, `UpdateError::reason`'s field doc still called the strings a contract, and the `(c): c` marker was the one branch that logged no command or stderr — invisible before, reachable now on exactly the codes an operator would be investigating.

Gates green locally: fmt, clippy `-D warnings`, rustdoc `-D warnings`, typos, 2466 workspace tests, 286 with `-F mcp`. Four mutations verified red before the commit.

mimi1vx (approved):

Issue #451, opened 2026-08-11
dead_probes cannot detect a broken downgrade probe: the pipeline reports awk's status, so a failed zypper se records 0 and the rollback silently runs zero commands


plusky:
Found while scoping #447. Same shape, different file — and this one sits in the
group-wide rollback path, so it is reachable exactly when an update has just
failed and mtui is trying to repair the fleet.

## The gate cannot catch what it says it catches

`crates/mtui-testreport/src/update_workflow/actions/downgrade.rs:23-29` —
`LIST_COMMAND`, shared by the zypper and slmicro downgraders:

```sh
zypper -n se -s --match-exact -t package $packages \
| grep -v "(System" \
| grep ^[iv] \
| sed "s, ,,g" \
| awk -F "|" '{ print $2,"=",$4 }'
```

`crates/mtui-testreport/src/reports/update_flow.rs:1090-1099` — the `dead_probes`
gate that is supposed to stop a broken probe from degrading into a silent no-op:

```rust
let dead_probes: BTreeSet<String> = list_map
    .keys()
    .filter(|hn| {
        targets.get(hn).and_then(mtui_hosts::Target::lastexit).is_some_and(|c| c != 0)
    })
    .cloned()
    .collect();
```

The comment above it (`:1085-1087`) states:

> The pipeline's exit status is awk's, so a recorded non-zero exit here always
> means the probe itself broke, never "package not found".

That sentence is **true**. But the gate depends on its **converse** — that a zero
exit means the probe worked — and that does not hold. A pipeline's status is the
last stage's, and `awk` exits 0 after a successful run over empty input. So when
`zypper -n se -s` itself fails (a ZYpp lock, broken repo metadata, an unreadable
rpmdb, zypper absent), the pipeline records **0**, the host never enters
`dead_probes`, and the very outcome the comment describes is what happens.

The gate therefore catches SSH-level death (the `-1` sentinel) and nothing else.

## What follows

With the probe recorded as healthy but empty:

* `versions` stays empty for that host, so the insert at `update_flow.rs:1117-1126`
  is skipped;
* non-transactional: the per-package loop's `let Some(ver) = versions.get(hn) else
  { continue }` (`:1152`) skips every package — zero commands run, and `run_checks`
  is never reached because the command map is empty (`:1175`);
* transactional: the host drops out of `combined` (`:1194`) and the reboot gate
  logs `no downgrade was staged; skipping reboot` (`:1262-1267`) — whose own
[...]

PR #452, opened 2026-08-11
fix(testreport): fail an update whose host could not work out what to patch

### LINKED ISSUE #447: update templates discard the patch-list substitution's status, so a failed refresh reports a successful update that installed nothing
Found while fixing #400, and deliberately left out of that PR: it is the same
class of bug, one step upstream, and closing it moves the rollback boundary
again.

## Summary

Both update templates open with

```sh
mtui_patches=$(zypper -n patches | awk -F "|" '/$repa\>/ { print $2; }')
```

The command substitution's own status is discarded — in POSIX the assignment's
status *is* the substitution's, and nothing reads it. So when the pipeline
fails, the patch list comes back empty, the patch is skipped, `mtui_status`
stays `0`, and mtui reports a successful update having installed **nothing**.

Failure modes that reach it:

* `zypper -n refresh` failed earlier in the script (its status is likewise
  discarded — no `set -e`);
* the update repo was never added, or was added under a different alias;
* `zypper -n patches` itself fails — exit `6` (no repositories defined) or `7`
  (ZYpp library lock), among others.

In each case the operator is told the update installed cleanly.

## Why it is not #400

#400 was "the patch ran and failed, and we discarded its status". This is "the
patch never ran, and we cannot tell that apart from having nothing to run". The
fix for #400 made the second case *explicit* — `mtui_status=0` on the skip path
— where it used to be an accident, which is what makes it worth stating now.

Note the skip path is also legitimate: a host carrying none of the update's
products genuinely has no patches, and must keep passing. So this is not a
matter of failing on an empty list; it needs the substitution's own failure
distinguished from a genuinely empty result.

## Suggested direction

Capture the substitution's status and route it separately from the patch's, e.g.

```sh
if ! mtui_patches=$(zypper -n patches | awk ...); then
    mtui_status=<sentinel>
fi
```

with the sentinel classified as its own verdict rather than reusing "package not
found" or "Unknown Error" — the operator needs "could not determine what to
patch", which is a different action from "the patch failed".

Two things to weigh before implementing:

1. **The blast radius.** A new failure class on this path fires the group-wide
   rollback downgrade, which reverts every host in the group. Whether "could not
   list patches" should roll back at all is a real question — arguably it should
   fail the host without reverting peers, like the `-1` never-ran gate does.
2. **A pipeline's status is the last stage's**, so `awk`'s success would mask a
   failing `zypper`. `set -o pipefail` is not POSIX; `${PIPESTATUS[0]}` is bash.
   The portable form needs thought, and the templates currently target plain
   `/bin/sh`.

## Related

The same shape exists in `downgrade.rs`'s `LIST_COMMAND`, whose consumer already
documents that the pipeline's status is awk's and therefore only ever means "the
probe itself broke" (`reports/update_flow.rs`). Worth fixing together.

plusky:
Closes #447.

Rebased onto `main` now that #446 has merged; this is standalone and its own two commits are the whole diff.

## The bug

The patch list was built by a pipeline, and a pipeline reports its **last stage's** status — awk's, which succeeds on empty input. zypper's status was not merely discarded, it was unobservable there. A ZYpp lock, a broken repo or an unreadable rpmdb gave an empty list, skipped the patch, and exited 0: an update reported as applied that installed nothing.

## The fix

Each probe runs on its own line so its status can be read, and the rows reach awk through the `printf` builtin. A failed probe prints a marker and exits with **its own** status — no invented sentinel, since a POSIX `exit` truncates to 0..255 and any chosen value would collide with the package manager's own space. There is no precedent in the tree for a template exiting with a status of its own choosing, and this does not add one.

**awk is guarded too.** It is the stage that actually produces the list, and it failed in exactly the same silent way — the first review round found #447 still reachable through it.

## Two judgement calls, both against the obvious choice

**The probe accepts `0`, and `106` only while the update's own repo is still listed.** Still deliberately narrower than the set the *patch's* status is read against — that one is argued from "the transaction committed", which is a claim about the patch, and on a metadata probe its members invert.

But rejecting `106` outright — which is what the first round of this PR did — was wrong for the same reason that leaves `refresh` unguarded below. `init_repos()` raises `106` for **any** skipped repository and cannot say which, and a refhost routinely carries one stale or unreachable repo alongside working ones. Rejecting it aborted exactly the hosts that leaving `refresh` unguarded was meant to protect: the two rules cancelled out and the same host failed one line later.

So the status is no longer asked to carry the verdict. On `106` the script asks the question the guard actually cares about — *is the update's own repo still there?* — via `mtui_issue_repo_present`, which greps `$repa` out of `zypper -n lr`, the same selector and the same listing the cleanup loop already matches on. Repo present: the skipped one was somebody else's, so note it in the transcript and carry on. Repo absent: the list cannot be trusted, and the probe fails.

Two properties keep that from becoming a false-positive machine, and both are pinned by tests:

- The alias question is asked **only** on the `106` path. A host carrying none of the update's products adds no repo, answers `0`, and is never asked — so the legitimate empty case stays the no-op it always was.
- `mtui_issue_repo_present` is a pipeline **on purpose**: awk's `END { exit !found }` *is* the verdict, so an `lr` that fails outright reads as "absent" and fails the probe. That is the conservative direction.

**`zypper -n refresh` is deliberately *not* guarded.** My first draft did guard it. Review caught that `refresh.cc:337-345` returns `4` whether one repository failed or all of them did, and that the explicit `refresh` command never returns `106` (that is set only by `init_repos` inside an operation command). So its status cannot separate "the issue repo went missing" from "an unrelated stale repo did", and guarding it would abort the update on a refhost that would have patched correctly — routine on QAM refhosts. The case is not lost: a refresh failure that really does hide the update resurfaces on the patches probe, which goes through `init_repos` and answers `106`, at which point the alias question above finds the repo missing or disabled.

## The failure does not roll back

A host that could not determine what to patch never ran a patch. Nothing is half-applied for a downgrade to undo, and the rollback is group-wide — it would revert every healthy host in the group. New `UpdateFailure::ProbeFailed`, selected from a typed flag on `UpdateError`, never from the reason text.

`ProbeFailed` rather than widening `NotRun` because the command *did* run to completion and *did* produce a verdict, so `NotRun`'s "absence of a verdict, state unknown" doc would have become false.

The predicate is now `any(repairable)` → `Check`, else `all(probe_failed)` → `ProbeFailed`, else `NotRun`. On any run with no probe failures this is provably identical to the old `all(lastexit == -1)` form, so the documented rule that a mixed run still rolls back on behalf of the host it can repair is unchanged. One case did change: a run mixing a probe failure with a lost host no longer rolls back, since neither host has anything to repair.

## Testing

Rendered templates are executed under a real `/bin/sh` with stubs — the suite's `MockConnection` scripts one exit code per command and cannot model a shell's last-command-wins rule. 40 mutations applied to the original design, observed red and reverse-restored; each red is credited by a panic message unique to its assertion. The `106` path was mutation-verified the same way after the redesign: collapsing the conditional acceptance back to a plain reject turns the new cases red on "106 with the update repo listed must still patch".

The legitimate empty case — a host carrying none of the update's products — is pinned as first-class at both script and check level. It stays a no-op, not a failure.

The probe-failure marker is matched **at the start of a line** rather than anywhere in the stream, so a `set -x` trace, a quoting log line or an indented copy of the text cannot forge a probe failure.

- The template's `\>` is a GNU awk extension. SUSE refhosts ship gawk, but a non-GNU awk would match nothing and skip silently; the module doc now says so, and says the fixture row is dialect-neutral so nobody mistakes the shell tests for awk coverage.
- Repos are added without `-f`, so `patches` does not autorefresh them: a stale-but-present cache still answers `0` with the patch absent. Closing that needs a freshness assertion, not a status test.
- zypper's exact `106` conditions are not documented anywhere I could verify; the design only assumes it means *some* repo was skipped, which is why it asks `lr` rather than trusting the code.
- No MCP client can distinguish "failed and rolled back" from "failed and left alone" — the variant collapses to a string at the boundary.
- `zypper -n patches` now runs twice per update, so the transcript's list is not literally the list that was patched.
- `SLM_UPDATE` has no refresh at all, so stale metadata on SL Micro is a silent no-op no guard can see.
- The same shape in the rollback path is filed separately as #451.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -160,7 +160,8 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   nothing to remove — so the shell reported that and discarded the
   `zypper -n in -t patch` / `transactional-update -n pkg in -t patch` status
   before anything could read it. Each template now captures the patch's own
-  status and exits with it, and the remote command text visible in `show_log`
+  status and exits with it — unless it never got as far as the patch, which is
+  the entry below — and the remote command text visible in `show_log`
   transcripts changes accordingly.
   The exit code is classified rather than compared against zero. `100`-`103`,
   `106` and `107` are zypper's *informational* results and still pass — `102`
@@ -182,6 +183,48 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   `transactional-update` absorbs zypper's status and returns only `0` or `1`,
   so neither the informational codes nor the package-not-found ones can reach
   the check on that key.
+- An `update` on a host that could not work out **what** to patch is no longer
+  reported as a successful update (#447). Both zypper templates decide whether
+  to patch from `zypper -n patches` and the `awk` that filters its output, and a
+  failure of either — exit `6` on a host with no repositories, `7` on a locked
+  ZYpp stack, an unreadable rpmdb, an `awk` that is broken or missing — produced
+  an *empty* patch list, which is indistinguishable from the legitimate "this
+  host carries none of the update's products". The patch was then skipped and
+  the script exited `0`, so mtui reported an update it had never installed.
+  Both are now captured with their own status read before anything can
+  overwrite it; the script prints
+  `mtui: could not determine what to patch: <command> exited <status>` and exits
+  with that command's own status. The check reads that line ahead of the
+  exit-code rules — the two statuses share one numeric space, so nothing else
+  could tell them apart — and reports the new reason **"could not determine
+  what to patch"**, which an operator (or an MCP client) sees instead of
+  "Unknown Error" or "package not found".
+  **Only exit `0` counts as an answer here**, deliberately narrower than the
+  rule the patch's own status is read against: `106`
+  ("some repository had to be disabled temporarily") is routine for a *patch*,
+  but on the listing step it may mean the disabled repository was the update's
+  own, in which case the missing patch would go unnoticed. The cost is that a
+  host carrying an unrelated stale or unreachable repository can now fail its
+  update with "could not determine what to patch" rather than patching; the
+  remedy is to fix or remove that repository. A host carrying none of the
+  update's products still passes — an empty list reached with a `0` status is an
+  answer, not a failure — and so does an update repo that was simply never
+  added, which zypper reports as a perfectly good list without the update's
+  rows.
+  `zypper -n refresh` is **not** judged: it returns the same status whether one
+  repository failed to refresh or all of them did, so on a refhost with an
+  unrelated broken repo it cannot distinguish a fatal problem from a routine
+  one. A refresh failure that really does hide the update surfaces on the
+  listing step instead.
+  Such a host fails the update but does **not** trigger the group-wide rollback
+  downgrade: it never ran a patch, so there is nothing half-applied to repair,
+  and rolling back would revert every healthy host in the group. A run mixing
+  one such host with a genuine check failure still rolls back, on behalf of the
+  host the rollback can actually repair; a run mixing one with a host the flow
+  lost (a timeout, a dropped connection) does not roll back at all, where it
+  previously did — neither host is one a downgrade could repair.
+  The remote command text changes again, so `show_log` transcripts change with
+  it.
 - `install` no longer fails when a package's `%post` script did. zypper exits
   `107` (`ZYPPER_EXIT_INF_RPM_SCRIPT_FAILED`) in that case — an *informational*
   code meaning the packages "were successfully unpacked to disk and are
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -199,18 +199,21 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   could tell them apart — and reports the new reason **"could not determine
   what to patch"**, which an operator (or an MCP client) sees instead of
   "Unknown Error" or "package not found".
-  **Only exit `0` counts as an answer here**, deliberately narrower than the
-  rule the patch's own status is read against: `106`
-  ("some repository had to be disabled temporarily") is routine for a *patch*,
-  but on the listing step it may mean the disabled repository was the update's
-  own, in which case the missing patch would go unnoticed. The cost is that a
-  host carrying an unrelated stale or unreachable repository can now fail its
-  update with "could not determine what to patch" rather than patching; the
-  remedy is to fix or remove that repository. A host carrying none of the
-  update's products still passes — an empty list reached with a `0` status is an
-  answer, not a failure — and so does an update repo that was simply never
-  added, which zypper reports as a perfectly good list without the update's
-  rows.
+  **Exit `0` counts as an answer here, and `106` only while the update's own
+  repository is still listed** — deliberately narrower than the rule the
+  patch's own status is read against. `106` ("some repository had to be
+  disabled temporarily") is routine for a *patch*, but on the listing step it
+  may mean the disabled repository was the update's own, in which case the
+  missing patch would go unnoticed. zypper raises it for any skipped
+  repository without saying which, so instead of judging the status the script
+  asks `zypper -n lr` whether the update's repository is there: if it is, the
+  skipped one was somebody else's and the update proceeds with a note in the
+  transcript; if it is not, the listing cannot be trusted and the update fails.
+  A refhost carrying an unrelated stale or unreachable repository therefore
+  keeps patching, as it does today. A host carrying none of the update's
+  products still passes — an empty list reached with a `0` status is an answer,
+  not a failure — and so does an update repo that was simply never added, which
+  zypper reports as a perfectly good list without the update's rows.
   `zypper -n refresh` is **not** judged: it returns the same status whether one
   repository failed to refresh or all of them did, so on a refhost with an
   unrelated broken repo it cannot distinguish a fatal problem from a routine
--- a/crates/mtui-testreport/src/reports/update_flow.rs
+++ b/crates/mtui-testreport/src/reports/update_flow.rs
@@ -455,7 +455,13 @@ fn aggregate_failures(op: &str, mut failures: Vec<UpdateError>) -> Result<(), Up
         // vanish on the other — and they are the declared routing contract,
         // not a detail of who reads them today.
         aggregate.probe_failed = failures.iter().all(|e| e.probe_failed);
-        aggregate.cancelled = failures.iter().all(|e| e.cancelled);
+        // `cancelled` is deliberately NOT propagated here. Every cancellation
+        // in this module is an early `return Err`, so no cancelled error ever
+        // reaches a `failures` vec — the line would be dead, and if it ever
+        // became live it would newly route a multi-failure aggregate to
+        // `CommandError::Cancelled`, letting a cancel mask a real host failure
+        // that merely coincided with it. That is a behaviour change, not part
+        // of this fix.
         Err(aggregate)
     }
 }
@@ -2532,6 +2538,10 @@ mod tests {
             "one host that did dispatch a patch clears the flag: {mixed:?}"
         );
 
+        // `cancelled` is deliberately not summarised. Every cancellation in
+        // this module is an early `return Err`, so no cancelled error reaches
+        // a `failures` vec; propagating it here would be dead code that, were
+        // it ever reached, would let a cancel mask a coincident real failure.
         let cancelled = aggregate_failures(
             "update",
             vec![
@@ -2541,8 +2551,8 @@ mod tests {
         )
         .unwrap_err();
         assert!(
-            cancelled.is_cancelled(),
-            "a summary of cancellations is still a cancellation: {cancelled:?}"
+            !cancelled.is_cancelled(),
+            "aggregation must not mint a cancel verdict: {cancelled:?}"
         );
     }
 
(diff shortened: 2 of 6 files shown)


plusky:
Rebased onto the updated #446 (`9fc823cb`). Three conflicts, all where the two PRs touch the same code:

- `UpdateError`'s doc — #446 retracted the "stable contract values consumed by callers" claim while this PR added `"could not determine what to patch"` to that same list. Kept both, and tied them together: the probe failure travels as a typed flag *because* the strings are not a contract.
- `classify_exit`'s `-1` doc — both branches had independently corrected the same stale claim about the rollback. Combined.
- `not_run_error`/`probe_failure` and the two new tests landed at the same offsets. The conflict tail left a duplicate copy of one test body under the other test's header; removed, and verified no duplicate function names remain.

Also dropped the refresh-residual paragraph #446 added to this module's docs: it duplicates the better-cited section already here, and on this branch it is partly wrong — the probe guard means a refresh failure that hides the issue repo now resurfaces on the patches probe instead of exiting `0`.

Gates on the rebased tree: fmt, clippy `-D warnings`, rustdoc `-D warnings`, typos, 2483 workspace tests, 286 with `-F mcp`, all clean.

mimi1vx (changes requested):
Reviewed the top commit `9fc823cb` only — the two below it are #446.

Verified locally in a worktree at `9fc823cb`: `cargo test -p mtui-testreport`
(299 lib + 151 it) passes, `cargo clippy -p mtui-testreport --all-targets -D
warnings` is clean, `cargo fmt --all --check` is clean. The new shell-level
tests do execute the rendered templates under a real `/bin/sh`.

The core fix is correct. The pipeline-status bug is real, per-line capture with
`$?` read on the following line is the right POSIX lever, and the routing
rewrite is equivalent to the old `all(lastexit == -1)` rule on any run without a
probe failure — `any(!probe_failed && lastexit != Some(-1))` reduces to
`any(lastexit != Some(-1))` when no error carries the flag, the `host: None`
case included. The `UpdateFailure` match in `perform_update_with_rollback` is
exhaustive, so the new variant cannot fall into a rollback arm by accident.

Two things to resolve before this lands.

## Rejecting `106` re-incurs the cost that leaving `refresh` unguarded avoids

`crates/mtui-testreport/src/update_workflow/actions/update.rs:100-112`

The module docs argue `zypper -n refresh` must stay unguarded because a refhost
routinely carries one unrelated stale/unsigned/unreachable repo alongside
working ones, and aborting there would kill updates that are fine. The same
docs then argue `106` must be rejected on the probe because `zypper -n patches`
goes through `init_repos()`.

But `init_repos()` sets `ZYPPER_EXIT_INF_REPO_SKIPPED` for *any* skipped repo
and cannot say which — as lines 80-82 state. So on exactly that refhost the
update now aborts one line later at `update.rs:190` with "could not determine
what to patch". The premises are in tension: leaving `refresh` unguarded buys
nothing if the next command hard-fails on the same condition.

Conditions for impact: any host with at least one repo whose metadata cannot be
loaded, where the issue repo itself is fine. Those hosts patch successfully
today and will need manual per-host repo cleanup after this change. The blast
radius is the refhost fleet rather than one code path, which is why this is
worth settling now.

The alternative the old module docs proposed resolves it without giving up the
detection: treat `106` as a warning, then assert the issue alias is actually
present (`repo_manager.rs::issue_alias` in `zypper -n lr`). That tests the thing
the guard cares about — the issue repo is there — instead of a status that
cannot distinguish it.

Note I could not verify zypper's exact `106` conditions from here; the
detection claim in the commit message depends on them holding as described.

## `aggregate.cancelled` is unreachable and unrelated to #447

`crates/mtui-testreport/src/reports/update_flow.rs:458`

```rust
aggregate.cancelled = failures.iter().all(|e| e.cancelled);
```

Every `UpdateError::cancelled` construction in this file is an early
`return Err(...)` — lines 553, 983, 1383, 1527, 1585. None is pushed into a
`failures` vec that reaches `aggregate_failures`. Check functions and
`host_command_failures` never set the flag either, so in production this
expression can only evaluate to `false`. Its only exercise is the new unit test
calling the private function directly (line 2535).

If it did become reachable, it would newly route a multi-failure aggregate to
`CommandError::Cancelled` through `map_flow_error` — a behavior change riding
in a bugfix commit, and one that conflicts with the guidance that a cancel
verdict must not mask a real host failure that merely coincided with it. Drop
the line or split it into its own commit. The `probe_failed` line beside it is
load-bearing and should stay.

## Suggestions

**The removed "residual" section is narrowed, not closed.**
`actions/update.rs:109` — "The case it would have caught is not lost" overstates
it. Repos are added with `zypper ar` and no `-f` (`mtui-hosts`
`target/repo_manager.rs`), so `patches` will not autorefresh them. If the issue
repo has a stale-but-present cache — a re-run of `update` after an earlier
successful `prepare`, with the repo now unreachable — `zypper -n patches`
answers `0` from cache, the list lacks the patch, and the script exits `0`: the
original silent no-op. The dominant case (freshly added repo, no cache,
unreachable → `106`) is genuinely covered. A one-line residual note would be
more accurate than dropping the section.

**Anchor the probe-failure match to line start.**
`crates/mtui-testreport/src/update_workflow/checks/update.rs:342` — the gate is
a substring match over all of stdout and stderr.
`args.stdout.lines().any(|l| l.starts_with(PROBE_FAILURE_MARKER))` costs the
same and removes the residual false-positive class the split-`printf` trick
does not cover: any host-side output reproducing the sentence mid-line.
`the_probe_marker_is_matched_in_full` already pins the prefix-weakening
direction; this pins the other one.

**Hedge the `ARG_MAX` claim.** `actions/update.rs:50-51` — "no `ARG_MAX` limit
sits between them" holds only while `printf` resolves to a builtin. True for
dash, bash and busybox, so fine in practice, but the doc states it
unconditionally. A `/usr/bin/printf` resolution with a large patch table gives a
false "could not determine what to patch" — fail-safe direction, worth one
hedging word.

**Rebase.** Stacked on #446; `assert_status_comes_from_the_patch` touches lines
#446 introduces, so expect a conflict there specifically.

plusky:
Both blocking items addressed in `c4dbb264`, and you were right on both. Also rebased onto the new `main` (#453) and onto the updated #446.

## `106` re-incurring the cost that leaving `refresh` unguarded avoids

Correct, and the tension was mine to resolve rather than argue away — the two rules cancelled out, and the fleet would have paid for it.

Implemented your alternative. On `106` the script no longer judges the status; it asks whether the update's own repository is still listed, with `mtui_issue_repo_present` greping `$repa` out of `zypper -n lr` — the same selector and the same listing the cleanup loop already matches on, so no new vocabulary. Repo present: the skipped one was somebody else's, note it in the transcript and patch. Repo absent: the list cannot be trusted, probe fails as before.

Two things I checked while implementing, both of which could have made this a false-positive machine:

- **The no-products host is not caught by it.** No repo is added for such a host, so `patches` answers `0`, no `106` arises, and the alias question is never asked. It is only ever asked on the `106` path, which is what keeps the legitimate no-op a no-op. Had the assertion been unconditional it would have failed every host that carries none of the update's products.
- **The presence test is a pipeline, deliberately.** Here the last-stage rule is what I want — awk's `END { exit !found }` *is* the verdict — so an `lr` that fails outright reads as "absent" and fails the probe. Conservative direction, and stated in the docs so it does not look like the bug this PR fixes.

Both directions are pinned at the end of `only_a_zero_probe_status_is_accepted`. Deleting the `106` arm fails the repo-present case; accepting `106` flat fails the repo-absent one. I ran both mutations and each went red on its own assertion — and caught that my first attempt at the delete-mutation had not applied at all, so its "pass" was meaningless.

The exhaustive refusal loop still lists `106` and still refuses it, because `Knobs::default()` leaves `repo_row` empty. I noted that in the comment so it does not read as a contradiction.

## `aggregate.cancelled`

Confirmed unreachable, independently: there is no `From<HostError> for UpdateError` in the tree, every `UpdateError::cancelled` is an early `return Err`, and the four `failures.push(e)` sites take check-function errors built with `UpdateError::new`, where the flag is false. Dropped, with the reasoning left at the site. The test now asserts the *opposite* — that aggregation does not mint a cancel verdict — so re-adding the line goes red rather than unnoticed.

## Suggestions

- **Line-anchored marker match** — applied, and pinned in both directions: a `set -x` trace or a log line quoting an earlier run no longer trips the gate, while a line that genuinely begins with the marker still does on either stream.
- **The residual** — you are right that "not lost" overstated it. Restored as an explicit residual: repos are added without `-f`, so `patches` does not autorefresh, and a stale-but-present cache still answers `0` with the patch absent. Closing that needs a freshness assertion, not a status test.
- **`ARG_MAX`** — hedged to the builtin `printf` it depends on, with the failure direction named.
- **Rebase** — done. The conflict was where you predicted, plus `UpdateError`'s doc and `classify_exit`'s `-1` doc, where this branch and #446 had independently corrected the same stale claim.

Gates: fmt, clippy `-D warnings`, rustdoc `-D warnings`, typos, 2486 workspace tests, 286 with `-F mcp`.

One thing I could not verify either: zypper's exact `106` conditions. The new design needs less from them than the old one did — it now only assumes `106` means *some* repo was skipped, not *which* — but if `init_repos()` can raise it without any repo actually being unusable, the note in the transcript would be noise on healthy hosts. Worth someone with a refhost confirming.

plusky:
Rebased onto `main` now that #446 has merged — the three commits below this PR dropped out as already applied, so it is down to its own two. I verified the rebase preserved the reviewed delta exactly: `git diff c5ba1b74..c4dbb264` and `git diff main..d5251ea3` are byte-identical.

I also corrected the PR description, which still described the design from before your review — it claimed the probe accepts `0` only. That is no longer true and the body now matches the code and the CHANGELOG.

mimi1vx (approved):

PR #454, opened 2026-08-15
fix(testreport): fail a rollback whose host could not work out what to downgrade

### LINKED ISSUE #451: dead_probes cannot detect a broken downgrade probe: the pipeline reports awk's status, so a failed zypper se records 0 and the rollback silently runs zero commands

plusky:
Closes #451.

## The bug

`downgrade` — and the automatic rollback `update` runs after a failed check — could report success having run zero downgrade commands.

The version list was produced by a single shell pipeline, and POSIX defines a pipeline's exit status as its **last** stage's. That stage is `awk`, and `awk` exits `0` after a successful run over empty input. So a failed `zypper -n se` — a ZYpp lock, broken repo metadata, an unreadable rpmdb, zypper missing — was not merely a discarded status, it was *unobservable*: the probe recorded exit `0` with an empty version map, every package was skipped for want of a version to roll back to, and the run finished by logging `done`.

`dead_probes` was written against that shape. Its comment was truthful — a recorded non-zero really did always mean the probe itself had broken — but the gate depends on the **converse**, that zero means it worked, and that did not hold. It could catch SSH-level death (the `-1` sentinel) and nothing else.

This is #447's defect in the path where it costs most. The rollback runs *after* an update has already failed on the group, so it is the last thing standing between a half-applied update and an operator who has just been told everything is fine.

## The fix

At the three layers the issue names.

**The template** (`update_workflow/actions/downgrade.rs`). #447's technique, deliberately the same one rather than a second dialect. Each producing command runs on its own line so its status can be read (an assignment's status is the substitution's — POSIX XCU 2.9.1); the rows reach the filter chain through the `printf` **builtin**, so no `ARG_MAX` limit sits between probe and filter; and `mtui_probe_ok` prints a start-of-line marker and exits with **the failed tool's own status**. No invented sentinel: a POSIX `exit` truncates to `0..=255`, so any value we picked would sit inside zypper's own space. The marker is assembled from `printf`'s format string *and* its argument, so the script's own text never contains the sentence it prints. `awk` is guarded too — it is the stage that actually produces the list, and `$?` after the filter pipeline is already awk's, so that capture is free.

**The gate** (`reports/update_flow.rs`). The `c != 0` predicate is unchanged; what changed is that it can now be believed in both directions. A non-zero is now either SSH-level death or the guard passing the tool's own status through — never "package not found", which the guard accepts as `104` and reports as `0`. And a zero genuinely means the probe answered, so an empty list at `0` is a host carrying none of these packages rather than a probe that failed unnoticed. A dead probe abandons that host's downgrade, and the command fails naming it (`h2: package version probe failed`).

**The verdict.** `downgrade_verdict` no longer logs `done` while any probe is dead; the unverified hosts are named at WARN instead. The per-host ERROR now says the host **was not downgraded** rather than that a step was *skipped* — the operator needs to know this host still carries the update, not that the flow omitted a step.

Accepted statuses are `0`, `104` and `106`; everything else refuses.

## Judgement calls

**A dead probe aborts that host, not the group.** This is the opposite of the update path's own probe failure, and deliberately so. There, nothing had been applied yet, so failing the whole group costs nothing. Here the run *is* the recovery from an update that already failed, so aborting it over one host's broken zypper would strand every healthy peer half-applied. The healthy hosts still roll back and still reboot; the error for the dead ones is raised at the end. All probes dead aborts immediately.

**No new failure variant.** `perform_downgrade` returns one `UpdateError`, nothing downstream routes on probe-ness, and on the rollback route the existing best-effort contract already keeps the original update error outranking it (the call site logs the downgrade error at WARN and re-surfaces the update's). Adding a variant would have bought a distinction no caller can act on. The stale doc on `perform_update_with_rollback` that claimed this worked because `perform_downgrade` "returns `()`" is corrected in the same commit — it has returned a `Result` for some time, and that Err path is now the carrier of this verdict, so the doc named the wrong mechanism at exactly the wrong moment.

**`106` is accepted, without the disambiguation the update path has.** `ZYPPER_EXIT_INF_REPO_SKIPPED` is raised by `init_repos()` for *any* skipped repository without saying which, and a refhost routinely carries one stale one. The update templates answer `106` by asking whether the update's own repo survived; that question is not merely hard here, it is unanswerable — `downgrade_body` calls `fanout_set_repo(RepoOp::Remove)` and removes the issue repository from every host *on purpose* before the probe runs, so its absence is the expected state. The released versions a rollback wants come from the host's ordinary repositories. `104` (`ZYPPER_EXIT_INF_CAP_NOT_FOUND`, an empty search result) is accepted for the plainer reason that it is the legitimate "this host carries none of these packages", routine in a mixed group.

**Withholding `done` is a second layer, not belt-and-braces.** `downgrade_verdict` flags a package only when the loaded report carries a `required` version to compare against. On a standalone `downgrade` there is none, so a dead probe produces exactly the same empty map a completed rollback does — the silent success of this issue restated one layer up, and the command failing does not stop `done` being printed above it in the same transcript.

**`commands/listversions.rs` is left alone.** It has the same pipeline shape, but it is advisory and read-only, and the issue asks for it separately.

**The second commit is a test-harness fix, not product code.** It is separate precisely so it can be reviewed as such.

## Testing

The template is pinned by executing the **rendered** script under a real `/bin/sh`, with a stub `zypper` on `PATH` and an awk-shadowing knob. That layer exists because `MockConnection` scripts one outcome per command string and cannot run a shell — it cannot express the last-command-wins rule that *is* the bug, so a mock-only test could not have caught this and cannot pin the fix. Asserting on the rendered text rather than the source also means a broken `$$`-escape is caught where a source-level look-alike would hide it.

Coverage across the two layers: a failed `zypper -n se` (marker present, versions empty, exits the tool's status); every one of the 17 refused codes rejected *and* `0`/`104`/`106` accepted, each half guarding the other against a trivially always-fail or always-pass guard; a broken `awk` (the discriminating assert is that the marker *names awk* — the old pipeline coincidentally also exits 2, so a status-only assert would have been vacuous); the healthy path; the `104`/`106` no-op paths; and at the flow layer, a probe exit of `7` routed as a dead probe (the `-1` sentinel is only one value of the predicate, so a predicate narrowed to `c == -1` would still pass the two pre-existing tests — this one pins the widened form), plus the `done`-withholding.

**How RED was observed.** Every assertion was driven red against unfixed or hand-broken code before being trusted, and the fixture was checked for its ability to express failure first:

- Reverting `LIST_COMMAND` to the old pipeline reds both render tests and the failed-`zypper`, exhaustive-code and broken-`awk` tests. Reverting the `downgrade_verdict` change reds the `done`-withholding test. Removing the `probe_dead` parameter breaks compilation at 7 call sites. No test in this diff passes with the feature deleted.
- The dead-probe flow test asserts `h2.commands().contains(&list)` — that the rendered probe really was dispatched — because an empty or mistyped `PATH` also yields "no versions", and without that liveness check the test would pass for the wrong reason.
- The probe is scripted per exact command string rather than through `with_default`. A mock answering every command identically leaves nothing attributable, which is how a headline test in #393 passed with the feature deleted.
- The `done` assert is `l == "done"`, not `contains("\"done\"")`. The capture layer renders a message unquoted, so the quoted form would have matched nothing at all and passed however the code behaved.
- In the final pass, three anchors of the `done`-withholding test were re-broken by hand and each watched go red: dropping the `probe_dead` guard around `done`, reverting the per-host wording, and deleting the unverified WARN.

That last exercise is what surfaced the second commit. The `done`-withholding test was **flaky** — failing about one run in three of `cargo test -p mtui-testreport --lib` while the code under test was correct. The test-only `capture_logs` helper scoped its subscriber with `tracing::subscriber::set_default`, whose guard is thread-local, but `tracing` caches callsite *interest* process-wide: a callsite first reached from a thread with no subscriber installed is cached as `Interest::never()` and stays silent for every later capture. Its sibling test drives the same `warn!`/`error!` callsites with no subscriber, so whichever reached them first decided whether the capture could see anything. `mtui-datasources`' `teregen` capture hit this same race and is already fixed the same way, so the second commit is this crate catching up with a pattern the workspace had settled: install the subscriber once, globally and permissively, and move the scoping to a thread-local sink. The captured set is unchanged — `set_default`'s guard was thread-local too, and the fan-out is `buffer_unordered` on the test's own single-threaded task. 20 consecutive lib runs green afterwards, and the three hand-breaks above were all re-verified *under the new helper*, one of them confirming the capture now holds the full transcript rather than the single line the race produced.

Full gate green: `fmt`, `clippy -D warnings`, `rustdoc -D warnings`, `cargo test --workspace` (37 suites, 2494 tests), `cargo test -p mtui-mcp -F mcp` (286 tests), the compile-only feature matrix both ways, and `typos`.

## Known limitations

- **The middle of the filter chain is still unguarded.** A failing `grep` or `sed`, or a missing binary among them, yields an empty list at awk's `0`, exactly as before — awk is the last stage and succeeds on no input. This is the same accepted residual the update templates carry; closing it needs `pipefail` (not POSIX) or `${PIPESTATUS[@]}` (bash-only), while these scripts must run under whatever `/bin/sh` a refhost provides.
- **The `106` residual is real and named rather than papered over.** If the skipped repository happens to be the one carrying the released versions, the list is silently short and those packages stay at the update version. On the rollback path `downgrade_verdict` catches it whenever the loaded update names `required` versions; on a standalone `downgrade` with none loaded, it is not caught. A status cannot carry that verdict — only a comparison against what should be installed can — which is what the transcript note exists to prompt.
- **The flow test does not pin the template, and the sh tests do not pin the flow.** The flow test passes with the template reverted, because `MockConnection` cannot express last-command-wins. That is the declared division of labour, not an oversight: the sh tests pin the template, the flow test pins the widened predicate, and reverting either reds its own side.
- **`show_log` transcripts change.** The remote command text is different, as it was for #447.
- **`commands/listversions.rs` still carries the same pipeline shape**, deliberately, for its own issue.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -302,6 +302,20 @@ The `tests/` fixtures are the authority for these formats; treat them as golden.
   lands with that prefix automatically; don't hand-name it otherwise.
 - **Gate real hosts/containers** behind `#[ignore]` + a CI env flag (sshd
   integration fixture); unit tests must run offline and fast.
+- **Capturing `tracing` output: install the subscriber globally, scope the
+  sink.** `tracing::subscriber::set_default`'s guard is thread-local, but
+  callsite *interest* is cached **process-wide**: a callsite first reached from a
+  thread with no subscriber installed is cached `Interest::never()` and stays
+  silent for every later capture, so with libtest running in parallel whichever
+  test got there first decides whether a log assertion can even fail. Install
+  once with `set_global_default` and move the scoping to a thread-local sink.
+  The pattern is settled and already written up three times — canonical copy:
+  `crates/mtui-datasources/tests/log_capture.rs`; also
+  `mtui-datasources::teregen` and `mtui-testreport::reports::update_flow`'s test
+  modules. Copy it, do not re-derive it. Known cost: an unfiltered `Registry`
+  reports no `max_level_hint`, so `LevelFilter::current()` goes to `TRACE` for
+  that whole test binary and every `debug!`/`trace!` in it starts evaluating its
+  arguments; bound the layer with a `LevelFilter` if that ever matters.
 - **A test that cannot fail is not evidence.** Twice a real regression has sailed
   through a green workspace run: once because the fixture disarmed the assertion
   (a `MockConnection` answering empty stdout builds no downgrade command, so
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -228,6 +228,32 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   previously did — neither host is one a downgrade could repair.
   The remote command text changes again, so `show_log` transcripts change with
   it.
+- `downgrade` — and the automatic rollback `update` runs after a failed check —
+  could report success while running zero downgrade commands (#451). The
+  package version probe was a shell pipeline, and a pipeline reports its last
+  stage's status: `awk`'s, which succeeds on empty input. So a failed
+  `zypper -n se` (a ZYpp lock, broken repo metadata, an unreadable rpmdb,
+  zypper missing) recorded exit `0` with an empty package list, the host was
+  taken for healthy, every package was skipped for want of a version to roll
+  back to, and the run finished with `done`. The probe now guards the commands
+  that produce the list, the same way the update templates do (#447): a failed
+  probe prints
+  `mtui: could not determine what to downgrade: <command> exited <status>` to
+  the transcript and exits with that command's own status, the host's downgrade
+  is abandoned, and the command fails (REPL and MCP) naming it. The healthy
+  hosts still roll back and still reboot — this run is usually the repair for
+  an update that already failed on the group — and the flow no longer logs
+  `done` while any host's state is unverified.
+  A host carrying none of the packages is still a no-op: `zypper search`
+  answers `104` for an empty result table, which stays an accepted, noted
+  answer. A skipped **unrelated** repository (`106`) is noted in the transcript
+  rather than failing the recovery — zypper raises it for any skipped
+  repository without saying which, and a refhost routinely carries one stale
+  one. The residual that leaves is named rather than papered over: if the
+  skipped repository was the one carrying the released versions, the list is
+  short and those packages stay at the update version — which the verdict
+  catches whenever the loaded update names required versions.
+  The remote command text changes, so `show_log` transcripts change with it.
 - `install` no longer fails when a package's `%post` script did. zypper exits
   `107` (`ZYPPER_EXIT_INF_RPM_SCRIPT_FAILED`) in that case — an *informational*
   code meaning the packages "were successfully unpacked to disk and are
--- a/crates/mtui-testreport/src/reports/update_flow.rs
+++ b/crates/mtui-testreport/src/reports/update_flow.rs
@@ -4222,6 +4222,27 @@ mod tests {
     /// the fan-out is `buffer_unordered` on the test's own task and
     /// `#[tokio::test]` is single-threaded, so no event under test is emitted
     /// off this thread.
+    ///
+    /// **Blast radius, accepted knowingly.** The `Registry` is unfiltered, so
+    /// it reports no `max_level_hint`; `set_global_default` reads that as "no
+    /// maximum" and `LevelFilter::current()` becomes `TRACE` for the *whole*
+    /// `mtui-testreport` lib test binary from the first capture onward. Every
+    /// `debug!`/`trace!` in the crate — previously dead, and none of them under
+    /// test here — then evaluates its formatting arguments and dispatches into
+    /// `CaptureLayer`, which drops it for want of a sink. The cost is argument
+    /// evaluation in tests only, and it buys the interest cache the race above
+    /// needs. Bounding the layer with a `LevelFilter` would keep the pinned
+    /// interest without the blast radius, and is the change to make if the lib
+    /// suite ever slows down noticeably.
+    ///
+    /// **This is the workspace's third copy of the pattern** — the others are
+    /// `mtui-datasources`' `tests/log_capture.rs` (the fullest write-up) and
+    /// `mtui-datasources::teregen`'s test module, and `mtui-core` gains a
+    /// fourth in #404/PR #459. They are copies rather than one helper because
+    /// each is a `#[cfg(test)]` module in a different crate and target
+    /// (unit-test modules cannot share an integration test's file); the
+    /// standing note in `AGENTS.md` § "Testing conventions" is what keeps the
+    /// next one from re-deriving the race from scratch.
     async fn capture_logs<T>(fut: impl std::future::Future<Output = T>) -> (T, String) {
         install_capture_subscriber();
         CAPTURE_SINK.with(|s| *s.borrow_mut() = Some(Vec::new()));
--- a/crates/mtui-testreport/src/reports/update_flow.rs
+++ b/crates/mtui-testreport/src/reports/update_flow.rs
@@ -202,9 +202,13 @@ where
 /// Every other failure installed nothing — no updater for a host's key, a
 /// prepare that could not run, a cancel before dispatch, a host the flow lost,
 /// or a host that could not determine what to patch — so each re-surfaces
-/// without a rollback attempt. The rollback is best-effort
-/// ([`perform_downgrade`] returns `()`), so it can never bury the original
-/// update error.
+/// without a rollback attempt. The rollback is best-effort, but not because it
+/// cannot fail: [`perform_downgrade`] returns a `Result`, and a host whose
+/// version probe never answered raises it (#451). The precedence is enforced at
+/// the call site, which logs that error at WARN and re-surfaces the original
+/// update error, so a failed rollback can never bury the failure it was trying
+/// to repair — do not simplify the call away, or a broken rollback goes silent
+/// again.
 pub async fn perform_update_with_rollback<R>(
     report: &R,
     targets: &mut HostsGroup,
@@ -1141,13 +1145,26 @@ async fn downgrade_body(
     }
 
     // A dead probe must abort that host's downgrade, not degrade it. When the
-    // probe dies (an SSH no-output timeout records exit -1) its stdout is
-    // empty, the version map below stays empty, and the flow would
-    // "complete" having run zero downgrade commands — leaving every package at
-    // the update version behind a success-looking run. The pipeline's exit
-    // status is awk's, so a recorded non-zero exit here always means the probe
-    // itself broke, never "package not found". Handled per host: the healthy
-    // hosts still roll back (and transactional ones still reboot); the error for
+    // probe dies its stdout carries no versions, the version map below stays
+    // empty, and the flow would "complete" having run zero downgrade commands —
+    // leaving every package at the update version behind a success-looking run.
+    //
+    // Non-zero is the whole signal, and the template is what makes it
+    // trustworthy in both directions (#451). It guards the commands that
+    // *produce* the list and exits with the failed tool's own status, so a
+    // non-zero status here is either SSH-level death (the `-1` sentinel) or the
+    // guard passing zypper's or awk's status through — never "package not
+    // found", which the guard accepts as `104` and reports as `0`. And zero now
+    // genuinely means the probe answered: an empty list at `0` is a host
+    // carrying none of these packages, not a probe that failed unnoticed. Left
+    // as one pipeline the status was the *last* stage's — awk's, and awk
+    // succeeds on empty input — so a failed `zypper se` recorded `0` and this
+    // gate could only ever catch the `-1`.
+    //
+    // Handled per host: the healthy hosts still roll back (and transactional
+    // ones still reboot), because this downgrade is often the repair for an
+    // update that already failed on the group, and aborting it over one host's
+    // broken zypper would strand the healthy peers half-applied. The error for
     // the dead ones is raised at the end. All probes dead aborts immediately.
     let dead_probes: std::collections::BTreeSet<String> = list_map
         .keys()
@@ -1164,7 +1181,7 @@ async fn downgrade_body(
         error!(
             host = %hn,
             exit = ?exit,
-            "package version probe failed; skipping downgrade on this host"
+            "package version probe failed; this host was not downgraded"
         );
     }
     if !dead_probes.is_empty() && dead_probes.len() == list_map.len() {
@@ -1351,7 +1368,7 @@ async fn downgrade_body(
             .map(reboot_error),
     );
 
-    let not_downgraded = downgrade_verdict(targets).await;
+    let not_downgraded = downgrade_verdict(targets, &dead_probes).await;
 
     // A per-host check failure aborts first (matches the pre-#336 aggregation).
     aggregate_failures("downgrade", failures)?;
@@ -1416,7 +1433,20 @@ async fn downgrade_body(
 /// map of packages still at or above the update version — empty on a fully
 /// completed rollback. Iterated in sorted hostname order (the group's own
 /// ordering) so the log is deterministic.
-async fn downgrade_verdict(targets: &mut HostsGroup) -> BTreeMap<String, Vec<String>> {
+///
+/// `probe_dead` names the hosts whose version probe never answered, and the
+/// all-clear is withheld while it is non-empty. Nothing was downgraded on those
+/// hosts, and this verdict cannot speak for them either: it flags a package only
+/// when the loaded report carries a `required` version to compare against, so on
+/// a standalone `downgrade` — or for a package outside the report's list — an
+/// unmeasured host produces the same empty map a completed rollback does.
+/// Logging `done` over it is the silent success of issue #451 restated one layer
+/// up. The hosts are named at WARN instead; the command's own failure is raised
+/// by the caller.
+async fn downgrade_verdict(
+    targets: &mut HostsGroup,
+    probe_dead: &BTreeSet<String>,
+) -> BTreeMap<String, Vec<String>> {
     // Query every host's versions concurrently via the shared fan-out, then
     // run the pure verdict scan below.
     targets.query_versions().await;
@@ -1448,8 +1478,20 @@ async fn downgrade_verdict(targets: &mut HostsGroup) -> BTreeMap<String, Vec<Str
         }
     }
 
+    if !probe_dead.is_empty() {
+        warn!(
+            hosts = %probe_dead.iter().cloned().collect::<Vec<_>>().join(", "),
+            "no package version probe answered on these hosts, so nothing was \
+             downgraded there and their package state is unverified"
+        );
+    }
+
     if not_downgraded.is_empty() {
-        tracing::info!("done");
+        // Only when every host was actually measured. A dead probe leaves this
+        // map empty for the same reason a completed rollback does.
+        if probe_dead.is_empty() {
+            tracing::info!("done");
+        }
     } else {
         for (hostname, names) in &not_downgraded {
             error!(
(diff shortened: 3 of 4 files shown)


mimi1vx (commented):
This is a good fix and the diagnosis is the valuable part: "a pipeline reports
its last stage's status, that stage is awk, and awk succeeds on empty input" is
the kind of root cause that is easy to paper over with a retry and hard to
actually find. The consequence — a rollback reporting `done` having run zero
downgrade commands, on the path that exists to repair an already-failed update —
justifies the size of the change.

The `rendered_script` harness is the right call and I would like to see more of
it. Running the *rendered* template under a real `/bin/sh` with a stubbed
`zypper` is the only way to test "which command's status the script reports",
and the module doc says exactly why `MockConnection` cannot express that state.
`only_zero_104_and_106_are_accepted_by_the_downgrade_probe` enumerating the whole
informational band rather than sampling it, and asserting the accepted three
separately so the refusal loop cannot pass vacuously, is a good shape. The
liveness assertion on the stub sentinel guarding the negative cases is the right
paranoia — a `PATH` mistake really would make every one of those vacuous.

Also good: `downgrade_verdict` taking `probe_dead` and withholding `done` rather
than inferring from an empty map. The doc explains precisely why the empty map is
ambiguous (no `required` version to compare against on a standalone `downgrade`),
which is the non-obvious half.

No blocking findings. A few notes.

### The `104`/`106` notes share a stream with the parser

`mtui_probe_ok`'s accepted-status notes go to **stdout**, which is the same
stream `parse_downgrade_versions` reads. It is safe today only because that
parser takes lines containing `" = "`:

```rust
if let Some((name, version)) = line.split_once(" = ") {
```

and neither note contains one. That is an implicit coupling between a shell
string and a Rust parser two modules away, and a future reword of the note (say,
`… exited 104, no package = no rollback`) would start injecting a bogus package
into the version map.

Either add a comment at the `printf` sites naming the constraint, or route the
notes to stderr. Stderr is not free here — `host_command_failures` treats any
stderr as a failure — so a comment is probably the cheaper answer. Worth a test
asserting the notes do not parse as versions, since that is the property you
actually depend on.

### Minor: unquoted glob

`grep ^[iv]` is subject to pathname expansion. Pre-existing and essentially
never going to fire, but you are rewriting the line anyway and `grep '^[iv]'` is
free.

### Test-subscriber note

The move from `set_default`'s thread-local guard to a permanent
`set_global_default` is correct, and the doc explaining the one-in-three flake
(callsite interest cached as `Interest::never` by whichever sibling reached it
first) is a genuinely useful piece of institutional knowledge — worth keeping
findable.

Two residuals, both non-blocking:

- An unfiltered `Registry` reports no `max_level_hint`, so `LevelFilter::current()`
  goes to TRACE for the whole `mtui-testreport` lib test binary once any capture
  runs. Every previously-dead `debug!`/`trace!` elsewhere now evaluates its
  formatting arguments. Harmless in practice; a `LevelFilter`-bounded layer would
  keep the interest rebuild without the blast radius.
- #459 adds the same pattern to `mtui-core`. Different test binaries so they
  cannot collide, but it is now a pattern rather than a one-off — worth a shared
  note somewhere, or it will be re-derived a third time.

### Cross-PR interaction with #457

This PR widens the all-probes-dead abort from "only the `-1` sentinel" to "any
non-zero probe status". #457 adds an `add_op_history(targets, "downgrade", …)`
call at that exact abort site. Together, the `:downgrade:<pkgs>` history row gets
written in strictly more cases where **zero** downgrade commands ran.

#457's justification (the issue-repo removal already landed, and this is the
rollback path where reconstructing host state matters most) still holds, but the
row's label over-claims more often once both are in. Worth a line in whichever
lands second. These two also touch adjacent lines in `downgrade_body`, so expect
a small conflict.

plusky:
Thanks — all four notes addressed. Three follow-up commits on top of `025aca61`, one per note, so each reply can be read against its own diff. Nothing amended or rebased: the branch is still fast-forward from what you reviewed.

### 1. The `104`/`106` notes share a stream with the parser — `92ad5997`

Took the comment option, not stderr, and added the test.

**The comment.** Two lines inside `LIST_COMMAND`, in the `case` block between the `0)` arm and `104)`:

```sh
  # Both notes print to stdout, which parse_downgrade_versions also reads. It
  # splits on space-equals-space, so no note may ever contain that separator.
```

One comment covering both `printf` sites rather than one at each: they sit four lines apart inside a twelve-line block, and the rendered template is remote command text that lands verbatim in `show_log` transcripts, so a second copy is operator-facing noise for no extra signal. It spells the separator out in words instead of writing it literally on purpose — the module already carries an assertion that *the script text must not contain the marker it prints*, for exactly this class of self-sabotage, and a comment containing a literal `" = "` would itself parse as a version row on any host running the command under `set -x`.

A comment is free to delete, so it is now pinned too, in the shared `assert_the_probe_is_guarded` (both downgraders):

```rust
assert!(
    listed.contains("parse_downgrade_versions"),
    "{name}: the notes must name the parser whose stream they share: {listed}"
);
```

**The test** — `rendered_script::the_probe_notes_never_parse_as_versions`. It renders the real template, runs it under `/bin/sh` with the stubbed `zypper`, and feeds the resulting real stdout through the real `parse_downgrade_versions` rather than through the harness's own `Ran::versions()`. Pinning the local copy of the `" = "` rule would pin the copy, not the contract; that needed `parse_downgrade_versions` to become `pub(crate)` (`reports::update_flow` is already a `pub(crate) mod`, so nothing new leaves the crate). The note text is never copied into the test, so a reword is caught because it *parses*, not because it stopped matching a literal.

Three cases: a refused status (marker only), `104` (note only), and `106` **beside a genuine version row** — the discriminating one, since an injected entry then shows up as a second map entry rather than as the difference between empty and non-empty. That case had no coverage at all before: `a_host_with_none_of_the_packages_stays_a_noop` already happened to catch a `104` reword through `Ran::versions()`, but nothing looked at the `106` note's output. An anti-vacuity guard asserts a note was printed, matched on the `mtui: note: ` prefix rather than on wording the script is free to change.

Observed red, each mutation applied to the template alone and reverted:

| mutation | observed failure |
|---|---|
| `104` note reworded to `… exited 104, no package = no rollback` | `the notes must contribute no package to the version map` — `left: [("mtui: note: zypper -n se exited 104, no package", "no rollback")]`, `right: []` |
| `106` note reworded to `… exited 106, skipped repo = short version list` | same assertion — `left: [("mtui: note: … skipped repo", "short version list"), ("pkg-a", "1.0-1")]`, `right: [("pkg-a", "1.0-1")]` |
| the `104` `printf` deleted, arm left as `return 0` | anti-vacuity guard fires: `the script printed no note at all, so nothing is being tested:` |
| the constraint comment deleted from the template | `the notes must name the parser whose stream they share` (both downgraders) |

**One factual correction, which does not change your conclusion but does change its reason.** "Stderr is not free here — `host_command_failures` treats any stderr as a failure" is true of `host_command_failures` itself (`update_flow.rs:485`, `let bad_err = !target.lasterr().is_empty();`) but it is not what judges this step. Inside `downgrade_body` (`update_flow.rs:1108-1418`) `host_command_failures` is called exactly once, at line 1121, against the **issue-repo removal** that runs immediately before the probe. The probe's own `targets.run(Command::PerHost(list_map))` at 1144 is judged by `dead_probes` at 1169, on the exit status alone:

```rust
.and_then(mtui_hosts::Target::lastexit)
.is_some_and(|c| c != 0)
```

There is no `host_command_failures` call between the list run and the next fan-out, so a note on stderr *would* survive today. I still did not route them there, for the reason your conclusion points at from a different angle: surviving is a property of one call site's choice, not of the flow, and the `104` arm exists precisely to stop that status failing the command — putting the note one snapshot field away from the rule that would refuse it buys nothing the comment and the test do not. The module doc records the refined version rather than the original claim.

### 2. Unquoted glob — `093b2a16`

`grep '^[iv]'` in the template. Two places pinned the old form and moved with it: the verbatim four-step block in `assert_the_probe_is_guarded` and `FILTERED_ROWS`' doc comment. Observed red by reverting the template hunk alone while leaving the assertion quoted — `the version list must be captured, guarded, filtered, guarded`, failing both `list_command_probes_all_packages_in_one_call` and its `slmicro` twin. So a *future* accidental unquoting is caught, not just this one.

`crates/mtui-core/src/commands/listversions.rs:21` carries the same shape (`grep -e ^[iv]`) and is deliberately left alone: your justification was "you are rewriting the line anyway", which does not apply to a file this branch does not touch. Named as a known sibling in the commit message — happy to do it as a separate one-liner if you would rather it went in now.

### 3. Test-subscriber residuals — `1491965f`

Documented, not changed — both, since you flagged them as non-blocking.

- **The `LevelFilter` blast radius.** Verified before writing it down: `CaptureLayer` implements no `max_level_hint` and `Registry`'s is `None`, so the layered subscriber reports no maximum and `LevelFilter::current()` resolves to `TRACE` for the whole `mtui-testreport` lib test binary from the first capture onward. `capture_logs`' doc now states the cost (argument evaluation, in tests only) and records your suggested remedy — a `LevelFilter`-bounded layer — as the change to make if the suite ever slows noticeably. Not making it now: it is a behaviour change to a harness stabilised two commits ago, and the residual is not the thing under repair.
- **The count is worse than you had it, which strengthens the point.** It is not the second copy but the **third**: `crates/mtui-datasources/tests/log_capture.rs` (the fullest write-up) and `mtui-datasources::teregen`'s test module (`teregen.rs:562`, also `set_global_default`) already carry it, and #459 makes a fourth.
- **Where the shared note went.** A doc comment in one of four copies is not shared, so the durable note is a bullet in `AGENTS.md` § "Testing conventions" — naming `log_capture.rs` as the canonical copy, the interest-cache race, and the `LevelFilter` cost — with the per-site rationale kept in `capture_logs`. That bullet is the one file outside `mtui-testreport` this branch now touches; it is a single append, so it should not fight the other open branches.

### 4. Cross-PR interaction with #457

No line added here. The note belongs to whichever of the two lands second, and I would rather write it once against the shape that actually results than pre-write it into a branch that may land first.

On the conflict you predicted: these three follow-up commits do not widen it. They touch `update_flow.rs` only at `parse_downgrade_versions` (~1522) and inside the test module (~4222) — both outside `downgrade_body`'s 1108-1418, which is where #457's `add_op_history` call lands.

### No changelog entry

The `[Unreleased]` entry for #451 already says "The remote command text changes, so `show_log` transcripts change with it" and describes the `104`/`106` notes; nothing here contradicts it. What these three commits add on top is a shell comment, a quoted glob, and test/doc prose — implementation detail under `CONTRIBUTING.md` § Changelog, with no new user- or MCP-visible behaviour.

### Gate

Full DoD gate re-run on `1491965f`, all observed green:

- `cargo fmt --all --check` — clean
- `cargo clippy --workspace --all-targets -- -D warnings` — clean
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features --document-private-items` — clean
- `cargo test --workspace` — 2495 passed, 0 failed, across 37 test binaries
- `cargo test -p mtui-mcp -F mcp` — 286 passed, 0 failed, across 4 test binaries
- `cargo build --workspace --no-default-features` and `--all-features` — both compile
- `typos .` clean; `git status --porcelain` empty

mimi1vx (approved):
LGTM. The diagnosis is exactly right — a pipeline's status is its last stage's,
`awk` exits 0 over empty input, so a failed `zypper -n se` was not a discarded
status but an unobservable one, and `dead_probes` could only ever catch the `-1`
SSH sentinel. Reusing #447's technique rather than inventing a second dialect is
the right call, and widening this gate on the rollback path — the recovery from
an update that has already failed on the group — is where it matters most.

I traced the rendered script by hand and am satisfied:

- `$?` after `mtui_versions=$(… | awk …)` is the pipeline's status, i.e. awk's.
- `mtui_probe_ok` exiting with the tool's own status sidesteps the
  sentinel-collision problem a POSIX `exit` truncation would create.
- Neither accepted-status note contains `" = "`, so `parse_downgrade_versions`
  cannot mint a phantom package from them.
- Per-host abort rather than group abort is the correct asymmetry against the
  update path: there nothing had been applied, here aborting over one host's
  broken zypper would strand every healthy peer half-applied.
- Quoting `grep '^[iv]'` is a real drive-by fix; the bare form was
  glob-expandable.

Two doc corrections, non-blocking:

1. The `ARG_MAX` argument depends on `printf` being a **builtin**. POSIX
   classifies it as a regular utility, not a required builtin — it happens to be
   builtin in bash, dash, ash and ksh, so this is safe on any realistic refhost,
   but the module doc states it as a property of POSIX. Please reword to "a
   builtin in every shell mtui targets" so nobody later leans on the stronger
   claim.
2. The "no `\" = \"` in these strings" rule is stated only on the shell side.
   Hoist it into `parse_downgrade_versions`' rustdoc as well, so the coupling is
   visible from both ends of it.

Please make sure the `commands/listversions.rs` follow-up issue actually exists
and is linked from the code, or that deliberate omission will rot.

Merge order: this touches `downgrade_body` / `downgrade_verdict` in
`reports/update_flow.rs`, which #457 and #458 also rewrite. Suggest landing this
third (after #459 and #456), then #458, then #457 — #457 has to place its
downgrade-abort history row inside this PR's widened abort branch.

Issue #431, opened 2026-08-05
Credentialed datasource URLs can leak via reqwest's Display through the doubly-transparent `HttpError::Request` chain


mimi1vx:
Audited at `9279bf30`. Line numbers are from that commit; the quoted snippets identify the sites unambiguously if they drift.

## Summary

`HttpError::Request` is `#[error(transparent)]` over `reqwest::Error`, and the per-datasource `…Error::Http` variants are `#[error(transparent)]` over `HttpError`. Rendering any of them therefore renders **reqwest's own `Display`**, which appends the request URL verbatim:

```rust
// reqwest-0.13.4/src/error.rs:280
if let Some(url) = &self.inner.url {
    write!(f, " for url ({url})")?;
}
```

No redaction. `AGENTS.md` states configured datasource URLs may legitimately embed credentials (`scheme://user:pass@host`) and must always go through `mtui_datasources::sanitize_url` before reaching a log or an error. This path bypasses that.

The most misleading instances are the sites that interpolate the error *next to* a sanitized URL — they read as safe while the raw URL rides along inside `{e}`.

## Confirmed leak: OBS send path

`crates/mtui-datasources/src/obs/client.rs:209-217`

```rust
builder.send().await.map_err(|e| {
    if is_ssl_verification_error(&e) {
        let host = host_of(url);
        tracing::error!("{}", ssl_verification_hint(host.as_deref()));
        tracing::debug!("OBS TLS error detail: {e}");        // :213 — no sanitizing at all
    } else {
        tracing::error!("OBS {method} {} failed: {e}", sanitize_url(url));  // :215 — sanitized url, raw {e}
    }
    ObsError::Http(e.into())                                  // :217 — transparent, leaks on Display
})
```

`e` is a raw `reqwest::Error` from `send()`, and reqwest attaches the URL on exactly that path — `async_impl/client.rs:3100` (`crate::error::request(e).with_url(self.url.clone())`), plus `:3077`/`:3085` for timeouts. So both log lines and the returned `ObsError` can carry `user:pass@host`.

## Same shape, error-kind-dependent

- `crates/mtui-datasources/src/gitea.rs:437` — `format!("{method} - {}: {e}", sanitize_url(url))`, `e: HttpError` from `read_body_capped`
- `crates/mtui-datasources/src/slack.rs:309` — same shape, same source

[...]

PR #456, opened 2026-08-15
fix(cli,mcp,core): hold the HTTP transport at INFO when debug is enabled

### LINKED ISSUE #439: Debug logging prints credential-bearing pool authorities from hyper-util; carve third-party transport targets out of the level directives
Found while reviewing #438 (the #431 fix); verified against hyper-util 0.1.20 and mtui's log wiring at `7f9446c9`.

## Summary

`hyper-util`'s connection pool logs the pooled connection key — `(scheme, authority)`, where `http::uri::Authority` includes userinfo — at DEBUG:

- `hyper-util-0.1.20/src/client/legacy/pool.rs:268` — `debug!("reuse idle connection for {:?}", key)`
- `pool.rs:401` — `debug!("pooling idle connection for {:?}", key)`

reqwest strips first-hop userinfo at builder construction (`extract_authority`), so an operator's own configured credential never reaches the pool key. But a redirect `Location: https://user:pass@host/…` is never re-stripped (`TowerRedirectPolicy::on_request` only handles sensitive headers and the Referer), so an attacker-supplied credentialed authority lands in hyper's pool key and is printed verbatim when DEBUG is enabled — observed as `pooling idle connection for ("http", alice:s3cret@127.0.0.1:9)` during #438's test work.

## Reachability

Both user-reachable log paths enable third-party DEBUG wholesale:

- `set_log_level debug` reloads a bare global directive (`crates/mtui-cli/src/lib.rs:136-141` via `level_directive`), no per-target carve-out; same for `-d/--debug` (`lib.rs:107-108`) and `RUST_LOG=debug`.
- `mtui-mcp` pins only `rmcp::service=warn` (`crates/mtui-mcp/src/runner.rs:237-238`, `:260-262`).

Impact is bounded — the leaked userinfo is attacker-supplied (from the hostile redirect), not the operator's credential, and reqwest strips `Authorization` on cross-host redirects — but "enable debug logging to diagnose a problem" should not print any credential-shaped data from transport internals.

## Suggested fix

Append `hyper_util=info,hyper=info,reqwest=info` to the level directives in both wiring sites (`level_directive` in mtui-cli; the mtui-mcp runner's filter), so a user-driven `debug`/`trace` applies to `mtui_*` targets without switching on third-party transport internals. An explicit `RUST_LOG` override should still win, as it does today, for anyone who genuinely needs hyper's view.

## Acceptance

- [ ] `set_log_level debug` (and `-d`) no longer emits `hyper_util`/`hyper`/`reqwest` DEBUG lines
- [ ] A test pins the directive string (both surfaces), so the carve-out cannot silently vanish
- [ ] `RUST_LOG=hyper_util=debug` still works as an explicit override

plusky:
Closes #439.

## The bug

`hyper-util`'s connection pool logs its pooled key — `(scheme, authority)` — at DEBUG (`pooling idle connection for …`), and an `http::uri::Authority` retains userinfo. `reqwest` strips first-hop userinfo when the client is built, but it never re-strips a redirect's `Location: https://user:pass@host/…`, so a hostile or misconfigured redirect puts a credential-shaped authority into the pool key. Every user-reachable way of turning DEBUG on then printed it:

- `mtui -d/--debug`
- a runtime `set_log_level debug` in the REPL
- `mtui-mcp -d/--debug`

All three built a bare `debug` (or `debug,rmcp::service=warn`) directive string, and a bare level raises *every* target — mtui's own and the third-party transport stack's alike.

This is the same class as #431 one layer lower: #431 stopped mtui from rendering credentialed URLs in its own errors and log lines; this stops mtui from switching on the log level at which the transport crates print them themselves.

## The fix

A new `mtui-core::log_filter` module holding one constant — `TRANSPORT_LOG_CARVE_OUT = "hyper_util=info,hyper=info,reqwest=info"` — and the policy for composing it with `$RUST_LOG`. It is appended to every `debug` directive string mtui builds for itself:

- **mtui-cli**: `level_directive` returns `debug,<carve-out>` for `LogLevel::Debug`. Both `-d` at startup and the `set_log_level` reload sink now go through it. `init_tracing` previously inlined its own `"debug"`/`"info"` literal — exactly the shape that lets one path drift from the other.
- **mtui-mcp**: `default_directives` appends the same constant to its `-d` arm.
- **both**: `resolve_log_directives` layers the same cap on top of an explicit `RUST_LOG`, and both entrypoints now resolve their startup filter through the shared `startup_filter` seam rather than `EnvFilter::try_from_default_env()` directly.

The cap is INFO, not `off`, so real transport warnings and errors still reach the operator.

**`RUST_LOG` is no longer honoured wholesale**, because that is the knob almost everyone actually reaches for and leaving it alone would have left the leak wide open behind a `-d` fix. Precisely:

- A `RUST_LOG` carrying a *global* `debug`/`trace` gets the cap layered on top, for each of the three targets its own directives do not already govern. `RUST_LOG=debug` therefore raises mtui and leaves the transport at INFO.
- **Naming a transport target** (`RUST_LOG=hyper_util=debug`, `RUST_LOG=hyper=trace`, …) remains an informed opt-in and is honoured verbatim — the cap is added only where the operator's own directive still wins under `EnvFilter`'s longest-match resolution. This is now the only setting that can print a credential-shaped authority, so mtui prints a one-line notice (`TRANSPORT_DEBUG_NOTICE`) to stderr at startup when one is in force.
- Everything else is left exactly as written. `RUST_LOG=error`, `RUST_LOG=off` and a `RUST_LOG` naming only unrelated targets get nothing added: the cap is only ever applied where it *lowers* a target, never where it would raise one.

## Judgement calls

- **Only the `debug` arm carries the carve-out.** Appending it at `error`/`warn` would *raise* the transport to INFO, above the level the operator chose. The `info` arms need nothing — their base already caps the transport at INFO. Both string pins assert the coarser arms stay bare, so a later "just append it everywhere" cannot land quietly.
- **INFO rather than `off`.** Silencing the transport entirely would hide the connection errors operators actually use when a refhost or datasource misbehaves. INFO sits below the DEBUG line that carries the pool key and above nothing worth keeping.
- **`hyper_util` is named even though `hyper=info` already covers it.** `EnvFilter` matches a directive's target as a prefix, so `hyper` alone already catches `hyper_util::client::legacy::pool`. It is named anyway so the intent survives a crate rename or a narrowing of the `hyper` entry — recorded in the constant's rustdoc so nobody simplifies it back out without knowing what they are removing.
- **The constant lives in `mtui-core`, not duplicated per binary.** Both surfaces depend on it; two literals in two crates is precisely the drift the CLI's own inlined `"debug"` had already demonstrated.
- **`rustls` and `tower_http` are deliberately absent.** Neither is a direct dependency logging an authority at DEBUG here: the tree has no `TraceLayer`/`tower_http` use at all, and `rmcp::service` is already pinned to `warn` on the MCP side.
- **A runtime `set_log_level` still discards `RUST_LOG`'s per-target directives**, including an explicit transport opt-in, because the reload replaces the whole filter. That is pre-existing behaviour; this change documents it rather than silently altering it.
- **The `RUST_LOG` trigger is a *global* `debug`/`trace`, not any `debug` anywhere.** Appending `hyper=info` to `RUST_LOG=error`, `RUST_LOG=off` or `RUST_LOG=mtui_core=trace` would push the transport **up** to INFO, above the level the operator chose — the same trap the coarser arms of the built-in defaults avoid. With a global `debug`/`trace` present the added directive can only lower a target.
- **"Named" means prefix-or-equal, matching `EnvFilter`'s longest-match rule.** `RUST_LOG=hyper=debug` hands over the whole hyper stack, `hyper_util` included, rather than having a `hyper_util=info` cap fight the request; `RUST_LOG=debug,hyper_util=warn` still gets `hyper` and `reqwest` capped, because a `hyper=info` beside it cannot overrule the longer `hyper_util=warn`.
- **The directive scan is deliberately lenient.** It only decides whether to *add* a cap, and the binary parses the real string afterwards, so an input it misreads costs at worst a redundant directive. The one direction it must never get wrong — reading a verbose directive as non-verbose — is closed by treating "no explicit level" as `TRACE`, which is what `EnvFilter` does with it.
- **The notice goes to `eprintln!`, not `tracing::warn!`.** The opt-in that triggers it is typically target-scoped (`RUST_LOG=hyper_util=debug`), which enables no `mtui_*` target at all, so a `WARN` event would be swallowed by the very filter it is warning about.

## Testing

Ten tests. Two per surface for the built-in defaults — one byte-pin and one behavioural filter test each — plus the `RUST_LOG` composition matrix on the `mtui-core` policy and through `mtui-cli`'s `startup_filter` (each `#[serial(env)]`, since `$RUST_LOG` is process-global and the crate's tests share one process).

The four covering the built-in defaults:

1. `level_directive_pins_bare_levels_and_debug_transport_carve_out` (mtui-cli) — exact equality on all four directive strings, with the `debug` one spelled out as a literal rather than rebuilt from `TRANSPORT_LOG_CARVE_OUT`, so emptying the constant cannot green both sides at once.
2. `set_log_level_debug_does_not_enable_transport_debug` (mtui-cli) — drives the real sink shape (`handle.reload(EnvFilter::new(level_directive(level)))`) on a scoped subscriber writing to a local buffer, and asserts three things: hyper-util's DEBUG pool line (carrying the unique token `s3cret`) never reaches the writer; an mtui DEBUG line does; a transport INFO line does. Arms 2 and 3 are the anti-vacuity guards — without arm 2 the test would also pass with the filter stuck at `info` (i.e. with `set_log_level debug` outright broken), and without arm 3 it would pass with the transport over-silenced to `warn`/`off`.
3. `default_directives_pin_rmcp_service_warn_and_transport_carve_out` (mtui-mcp) — the same literal pin for both arms.
4. `debug_default_filter_suppresses_transport_debug` (mtui-mcp) — the same three-arm behavioural test, building the real fallback filter with `EnvFilter::try_new(...).expect(...)` so an unparseable directive panics instead of degrading silently (`EnvFilter::new` drops a malformed directive without a word).

**How RED was observed.** Each production helper was hand-broken back to its pre-fix one-liner and the tests re-run:

- `level_directive` returning bare `base` for `Debug` → both mtui-cli tests FAILED: `left: "debug"` / `right: "debug,hyper_util=info,hyper=info,reqwest=info"`, and the behavioural test's dump showed the leaked line verbatim — `DEBUG hyper_util::client::legacy::pool: pooling idle connection for ("http", alice:s3cret@example.test:9)`. `2 failed; 87 filtered out`.
- `default_directives` returning `format!("{base},rmcp::service=warn")` → both mtui-mcp tests FAILED the same way under `-F mcp`. `2 failed; 232 filtered out`.

The filtered-out counts are recorded on purpose: a run that matched no tests would report `0 filtered out` with 0 running, and could not be mistaken for evidence.

**Mutation note.** The one mutation the behavioural tests cannot kill is respelling `hyper_util` as `hyper-util`: `EnvFilter` parses it, and the `hyper=info` prefix match still covers the target, so the leak stays closed by accident. That is exactly what the two literal string pins catch, and the test's own rustdoc says so.

Full gate green on the final commit: `fmt --check`; `clippy --workspace --all-targets -D warnings`; `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --all-features --document-private-items`; `cargo test --workspace` (2487 passed / 0 failed across 28 test binaries); `cargo test -p mtui-mcp -F mcp` (287 passed / 0 failed); both feature-matrix builds (`--no-default-features`, `--all-features`, compile-only); `typos`.

## Known limitations

- **The two `init_tracing` startup blocks are not unit-observable.** They install a *process-global* subscriber, so a test cannot stand up a competing one to inspect. Both now take their default from the tested builders (`level_directive` / `default_directives`), so what the tests pin is the exact string those blocks use — but someone re-inlining a literal there would be caught by review, not by the suite.
- **`RUST_LOG=hyper_util=debug` (or any explicitly named transport target) still prints the pool authority.** That is the deliberate escape hatch, not an oversight — it is the only way to get the transport's own view — and it is the reason for the stderr notice.
- **The `RUST_LOG` scan reimplements a slice of `EnvFilter`'s directive grammar.** It is lenient by construction and a misread only costs a redundant cap, but it has to track `tracing-subscriber`'s grammar across upgrades; nothing pins the version it was written against.
- **`h2` is excluded on a source read, not on an enforced invariant.** Its DEBUG lines are frame shapes and protocol breadcrumbs and its `Debug for Headers` omits the header block (h2-0.4.15), so nothing header-bearing is rendered below TRACE. A future h2 that changes that would not be caught automatically.
- **The carve-out is a denylist of three crate roots.** A future *direct* HTTP dependency that logs authorities at DEBUG would have to be added by hand; nothing enforces that automatically.
- **This does not redact the authority.** It stops mtui from turning on the level that prints it. Redacting hyper-util's own log line is not ours to do.

---

_Description updated by a maintainer at merge time: the `RUST_LOG` layering, the `log_filter` module and the startup notice landed after the original write-up, which still described the `-d`-only version of the fix._


mimi1vx (changes requested):
Correct fix for a real leak, and routing both startup and the runtime
`set_log_level` through the same `level_directive` helper is the right shape —
that is the bug class ("the carve-out applies to one path and not the other")
closed by construction rather than by discipline.

The test pair is well built. Pinning the directive strings as literals *and*
driving the assembled filter through a real subscriber is exactly right, since
`EnvFilter::new` swallows a malformed directive and a byte-pin alone cannot tell
a working carve-out from an inert one. The anti-vacuity assertion ("mtui debug
reaches the log") and the "cap is `info`, not `warn`" assertion are both load-
bearing. Verified the doc's claim about prefix matching: `EnvFilter` compares
targets with a raw `starts_with`, so `hyper=info` alone would indeed already
cover `hyper_util::client::legacy::pool`, and naming `hyper_util` anyway is a
reasonable hedge.

One residual worth resolving before this lands under a **Security** heading.

### `RUST_LOG=debug` bypasses the carve-out entirely

Both binaries resolve `EnvFilter::try_from_default_env()` first, so any
`RUST_LOG` replaces the defaults wholesale. The docs frame that as the escape
hatch:

> An explicit `RUST_LOG` (e.g. `RUST_LOG=hyper_util=debug`) overrides the
> defaults entirely for anyone who needs the transport's own view.

`RUST_LOG=hyper_util=debug` *is* an informed opt-in. But `RUST_LOG=debug` is not
— it is the single most common way anyone turns on debug logging, it names no
transport target, and it silently re-enables the credential-bearing pool key
this PR exists to suppress. An operator who reads the CHANGELOG entry and
concludes "debug logging is safe now" gets the leak back the moment they reach
for the env var instead of `-d`.

The threat model here is a hostile redirect putting userinfo into the pool
authority, which does not care which knob enabled DEBUG.

Options, roughly in order of preference:

1. Apply the carve-out **after** a parsed `RUST_LOG` unless it names one of the
   transport targets explicitly. `EnvFilter` can be built up with
   `add_directive`, so "user's directives, then the caps, unless the user
   mentioned hyper/hyper_util/reqwest" is expressible.
2. Emit a one-line `WARN` at startup when the resolved filter would enable DEBUG
   for the transport, naming the risk. Cheap, no behaviour change, and it makes
   the residual visible at the moment it matters.
3. If neither is wanted, say so plainly in the CHANGELOG — something closer to
   "`RUST_LOG=debug` still enables the transport's DEBUG output and can print
   credential-shaped authorities" rather than the current framing, which reads as
   though only a deliberate transport opt-in does.

Happy with (2) as a minimum; (1) is the one that actually closes it.

---

### Non-blocking

**Placement.** `TRANSPORT_LOG_CARVE_OUT` lives in `mtui-core/src/session.rs`,
which is otherwise `Session` state. It is a logging constant consumed by two
binaries. Not worth a new module on its own, but a sentence on *why* it lives
here (shared by both entrypoints, `mtui-core` is the common crate) would age
better than the current placement-by-convenience.

**Scope of the carve-out.** `hyper_util`/`hyper`/`reqwest` covers the known leak.
Worth a moment's thought on whether anything else in the transport stack logs a
URL or authority at DEBUG — `h2` is mostly TRACE-level frame logging, so probably
nothing, but if you have already checked, saying so in the constant's doc turns
"these three" from a list into a conclusion.

**`mtui-mcp` has no `set_log_level` sink**, so the MCP-exposed `set_log_level`
tool is a no-op there and only `default_directives` matters. Pre-existing and
correctly documented in `runner.rs::init_tracing` — just confirming it is not a
gap this PR opens.

plusky:
Both points are addressed. Two commits on top of `084db3cf`:

- **`fb05d06f`** — layers the carve-out onto `RUST_LOG` (the blocking point, the CHANGELOG reframe, and both non-blocking points).
- **`79ec52c7`** — a test-hygiene follow-up found while re-reading the new tests (last section).

## Blocking: `RUST_LOG=debug` bypasses the carve-out

Implemented option (1) **and** option (2); (2) is layered on, not instead of.

**(1)** Both entrypoints now resolve through one `mtui-core` helper, `resolve_log_directives` (`crates/mtui-core/src/log_filter.rs`), called from `startup_filter` in `crates/mtui-cli/src/lib.rs` and `crates/mtui-mcp/src/runner.rs`. `RUST_LOG=debug` now yields `debug,hyper_util=info,hyper=info,reqwest=info` on both binaries. There is one place that decides, not one per entrypoint, so the property you praised stays true by construction.

**(2)** The same parse produces `LogDirectives::transport_debug_opt_in`, so the startup notice fell out of it. It is not made redundant by (1) — after (1) an explicit transport directive is the *only* remaining way to print a pool authority, so announcing it is exactly what keeps the residual visible.

### Deviation 1 (deliberate): the cap is not applied unconditionally

It is applied only when the operator's `RUST_LOG` carries a **global** `debug`/`trace`. Appending `hyper=info` to `RUST_LOG=error`, `RUST_LOG=off` or `RUST_LOG=mtui_core=trace` would **raise** those targets — from the operator's chosen `error`/`warn`, or from silence, up to `INFO`. That is the same trap this PR's own `level_directive` already avoids for the coarse levels: *"The coarser levels stay bare: appending the carve-out there would raise those targets to INFO above an operator's chosen error/warn"* (`crates/mtui-cli/src/lib.rs`).

The narrow trigger loses no coverage. `EnvFilter` matches a target-ful directive by prefix only, so a *target-less* directive is the only way an un-named target such as `hyper_util::client::legacy::pool` reaches DEBUG; with no global verbose directive the transport is not enabled at all and there is nothing to cap. Pinned by `coarse_and_silent_globals_are_left_alone`, `unrelated_target_directive_alone_adds_nothing`, `empty_rust_log_stays_empty`.

### Deviation 2 (deliberate): "names a transport target" is asymmetric

The cap stands down for target *T* only when the operator's target is a prefix of (or equal to) *T* — the only shape where `EnvFilter`'s longest-match resolution would let our `<T>=info` **overrule** them. Verified in tracing-subscriber 0.3.23: `impl Match for Directive` is a raw `meta.target().starts_with(&target)`; `Ord for Directive` orders by target length; and equal-target directives replace one another in `DirectiveSet::add`, so an appended `hyper=info` would silently overwrite an operator's `hyper=warn`.

So `RUST_LOG=hyper=debug` hands over the whole hyper stack, `hyper_util` included, rather than having the cap fight the request; while `RUST_LOG=debug,hyper_util=warn` still caps `hyper` and `reqwest`, because our shorter `hyper=info` loses to their longer `hyper_util=warn`. The *notice* uses the broader both-directions relation, since a longer target (`hyper_util::client::legacy::pool=debug`) beats the cap for the events it matches even though the cap still goes on. Pins: `a_shorter_prefix_names_the_longer_transport_target`, `naming_the_longer_target_does_not_shield_the_shorter_one`, `a_longer_target_is_announced_and_still_takes_the_cap`.

### Three smaller implementation notes

- The notice is `eprintln!`, not `tracing::warn!`. The opt-in that triggers it is usually target-scoped (`RUST_LOG=hyper_util=debug`), which enables no `mtui_*` target, so a WARN event would be filtered out by the very filter it is warning about. stderr is already both binaries' log channel; `mtui-mcp`'s stdout is the JSON-RPC stream.
- A malformed `RUST_LOG` still falls back to the (capped) entrypoint defaults exactly as `try_from_default_env()` did, and the discarded opt-in is not announced — `malformed_rust_log_falls_back_to_the_capped_defaults`, on both binaries.
- The helper has two public entry points on purpose: `resolve_log_directives(defaults)` reads `$RUST_LOG`; `resolve_log_directives_from(rust_log, defaults)` is pure and carries the policy unit tests. The binaries call the env-reading one and the per-binary `#[serial(env)]` tests set the real variable, so a wrong variable name cannot hide behind green pure-function tests.

## Required: CHANGELOG reframing

Done in `fb05d06f`, by rewriting this PR's own not-yet-merged `[Unreleased]` entry (preferable to leaving a contradiction in the file). It no longer claims an explicit `RUST_LOG` "overrides the defaults entirely"; it states the three precise cases: a global `debug`/`trace` gets the cap layered on top; naming a transport target stays an informed opt-in that mtui never overrules and now announces on stderr; `error`, `off` and unrelated-only values are left exactly as written, because the cap is only ever added where it *lowers* a target.

## Non-blocking: placement of the constant

Answered by moving it, not by a sentence. `TRANSPORT_LOG_CARVE_OUT` left `crates/mtui-core/src/session.rs` (which is otherwise `Session` state) for the new `crates/mtui-core/src/log_filter.rs`, which also holds the resolver. The module doc records both reasons: `mtui-cli` and `mtui-mcp` are sibling leaves, so the composition root is the only place they can share a security decision about logging; and the policy stays a directive *string* because `mtui-core` must not depend on `tracing-subscriber`. The public path `mtui_core::TRANSPORT_LOG_CARVE_OUT` is unchanged.

## Non-blocking: does `h2` belong on the list?

Checked in the version actually in the tree (h2 0.4.15, `Cargo.lock`), deliberately left off, and the conclusion is now written into the constant's doc:

- The frame DEBUG lines are `tracing::debug!(frame = ?item, "send")` (`src/codec/framed_write.rs:218`) and `tracing::debug!(?frame, "received")` (`src/codec/framed_read.rs:418`). Both render through `impl fmt::Debug for Headers` (`src/frame/headers.rs`), which stops at `stream_id`/`flags`/`stream_dep` and carries the comment ``// `fields` and `pseudo` purposefully not included``. `PushPromise`'s `Debug` says the same. So no `:authority`, no `authorization`, no header value.
- Every other DEBUG site is protocol-error breadcrumbs and counters — flow-control and header-size errors in `src/proto/streams/recv.rs`, connection errors in `src/proto/connection.rs`, hpack's "Decoder changed max table size" with sizes only.
- The only header-bearing lines are TRACE, and the sole header *value* among them is the TE header.

Your "mostly TRACE-level frame logging" is right in conclusion; the precise reason is the `Debug` impl. Adding `h2=info` would cap something that leaks nothing and would blur the constant's "these three are the whole surface" claim, so it stays out.

## Non-blocking: the MCP `set_log_level` sink

Left untouched, and here is what the code shows. `mtui-mcp` installs **no** log-level sink at all: `init_tracing` in `crates/mtui-mcp/src/runner.rs` builds a plain stderr subscriber with no reload handle, and `Session::apply_log_level` (`crates/mtui-core/src/session.rs`) invokes the sink only if one was installed, returning `false` otherwise. So an MCP `set_log_level debug` cannot change that process's filter — there is no second path to the transport there. On the REPL side the sink rebuilds the filter from `level_directive`, which already carries the carve-out at `debug` (`084db3cf`), so that path was closed by the first commit.

## The third suggested test, substituted

`RUST_LOG=mtui_core=trace` "still keeps the carve-out" is not assertable as written under the narrow trigger: that value needs no carve-out (an un-named transport target is not enabled at all), and it cannot be observed red against the pre-fix code, which already suppressed the transport for exactly that value. I kept the intent — *an unrelated `RUST_LOG` must not be mistaken for a transport opt-in* — as `RUST_LOG=mtui_cli=trace,debug` / `mtui_mcp=trace,debug`: an unrelated target plus the global `debug` that does reach the transport. That version is red pre-fix (below), and it also asserts the operator's own TRACE still flows.

## How the tests were observed red

Five mutation rounds against the final code; every failure below was observed, then reverted and the suites re-run green.

1. **Cap never layered on** (i.e. pre-fix "`RUST_LOG` replaces the defaults"). Red: `global_debug_gets_the_full_carve_out` (`"debug"` vs `"debug,hyper_util=info,hyper=info,reqwest=info"`), `global_trace_gets_the_full_carve_out`, `unrelated_target_directive_does_not_stand_the_carve_out_down`, `a_shorter_prefix_names_the_longer_transport_target`, `a_longer_target_is_announced_and_still_takes_the_cap`, `a_transport_target_named_at_a_coarse_level_is_not_an_opt_in`, `naming_the_longer_target_does_not_shield_the_shorter_one`; plus, on **both** binaries, `rust_log_debug_still_holds_the_transport_at_info` and `rust_log_unrelated_target_keeps_the_transport_capped`, each printing the leak line ``DEBUG hyper_util::client::legacy::pool: pooling idle connection for ("http", alice:s3cret@example.test:9)``.
2. **Cap applied unconditionally** (the literal version of the guidance). Red: `coarse_and_silent_globals_are_left_alone` (`"error"` → `"error,hyper_util=info,hyper=info,reqwest=info"`), `unrelated_target_directive_alone_adds_nothing`, `empty_rust_log_stays_empty` (`""` → `",hyper_util=info,…"`), `explicit_transport_opt_in_is_untouched_and_announced`. This is the evidence for deviation 1.
3. **"Named" made symmetric.** Red: `naming_the_longer_target_does_not_shield_the_shorter_one` (`hyper` lost its cap) and `a_longer_target_is_announced_and_still_takes_the_cap`. Evidence for deviation 2.
4. **Notice fired regardless of level.** Red: `a_transport_target_named_at_a_coarse_level_is_not_an_opt_in` and `naming_the_longer_target_does_not_shield_the_shorter_one` — the false-*positive* direction of the warning is pinned too, not only the false-negative one.
5. **`reqwest=info` dropped from `TRANSPORT_LOG_CARVE_OUT`.** Red: `carve_out_string_caps_exactly_the_listed_targets`, plus both byte-pins — `level_directive_pins_bare_levels_and_debug_transport_carve_out` and `default_directives_pin_rmcp_service_warn_and_transport_carve_out`.

Two of the fourteen resolver tests were red in **no** round, and I am not claiming otherwise: `unset_rust_log_keeps_the_entrypoint_defaults_verbatim` and `parse_directive_reads_targets_levels_and_implicit_trace`. Both are direct `assert_eq!` on a pure function, pinning the pass-through and the parser that none of the five mutations touch; there is no fixture in them that could disarm an assertion.

The test you had already read, `debug_default_filter_suppresses_transport_debug`, keeps its name and all three assertions — only its inline writer/event block moved into the shared `probe()` helper the new tests reuse. Same on the CLI side; no assertion was weakened.

## Follow-up found while re-reading the tests (`79ec52c7`)

`mtui-cli` had ended up with two disjoint exclusion domains over one process-global: the new `RUST_LOG` tests serialise on `#[serial(env)]`, while `edit`'s two `$EDITOR` tests guarded the same environment with a private `EDITOR_ENV` mutex, and `notification`'s real-environment probe had no guard at all. Two domains over one global exclude nothing from each other — `run_edit_uses_editor_env_and_path` can be spawning `$EDITOR`, a child that inherits the whole environment, at the moment `with_rust_log` calls `set_var`. That is the shape edition 2024 made `set_var` `unsafe` for, and it made `with_rust_log`'s own SAFETY comment ("no other test observes or mutates the environment concurrently") untrue while every test stayed green.

`79ec52c7` deletes the private mutex and puts all eight env-touching tests in the crate on `#[serial(env)]` — the rule AGENTS.md already states for a process-global under a single test binary.

No red evidence for this one, deliberately: the failure mode is a data race between tests, not a value an assertion can read. The check on it is structural — the crate now has exactly one env domain (five `startup_filter` tests, two `edit` tests, one `notification` test, no private env lock left), and `mtui-cli`'s suite is green.

## Not done

No `docs/src` change: `grep -rn RUST_LOG docs/ README.md` returns nothing, so no prose documents the variable; the mdBook's only debug references are generated clap help, which is unaffected.

## Gate

Full local gate observed green on `79ec52c7`: `cargo fmt --all --check`; `cargo clippy --workspace --all-targets -- -D warnings`; `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features --document-private-items`; `cargo test --workspace` 2506 passed / 0 failed; `cargo test -p mtui-mcp -F mcp` 291 passed / 0 failed; `cargo build --workspace --no-default-features` and `--all-features`; `typos` clean. The `Cargo.lock` delta is the two `serial_test` dev-dependency entries only.

mimi1vx (approved):
LGTM on the security analysis and on the shape of the fix. `hyper-util`'s pool
key does retain userinfo, `reqwest` does not re-strip a redirect's `Location`
userinfo, and putting the policy in `mtui-core` rather than duplicating a literal
per binary is right — the CLI's own inlined `"debug"` had already demonstrated
that drift.

**Please update the PR description before this merges.** It no longer describes
the diff, and the mismatch is in the security direction:

- Body: *"An explicit `RUST_LOG` still replaces the defaults wholesale (both
  binaries resolve `EnvFilter::try_from_default_env()` first)"*
- Body, Known limitations: *"`RUST_LOG=debug` still enables transport DEBUG."*

The diff does the opposite — it intercepts `RUST_LOG`, layers the cap on top via
the new 447-line `log_filter` module, and prints a stderr notice on the opt-in
path. The CHANGELOG entry is correct; only the prose is stale. Body also says
"four tests, two per surface" against roughly ten.

Non-blocking, but worth doing:

1. `resolve_log_directives_from` is a hand-rolled reimplementation of
   `EnvFilter`'s directive grammar. I worked the table by hand —
   `hyper_util=debug`, `debug,hyper_util=warn`, `debug,hyper=warn`,
   `mtui_cli=trace,debug`, `[span]=debug`, `target=` — and the prefix /
   longest-match reasoning holds, and it is lenient-by-design so a misread only
   costs a redundant cap. But it now has to track `tracing-subscriber`'s grammar
   forever. Please pin the `tracing-subscriber` version it was written against in
   the module docs, and add a test asserting the *composed* string parses under
   `EnvFilter::try_new` for each shape — otherwise a grammar change degrades
   silently into the defaults fallback instead of going red.
2. The `h2` exclusion rests on a source read of `h2-0.4.15
   src/frame/headers.rs`. Nothing enforces that across a dependency bump; put the
   version in the comment so the justification visibly expires.
3. The notice fires only on the opt-in path. Someone debugging a refhost HTTP
   problem with `RUST_LOG=debug` now gets a silently capped transport and no
   hint why. Consider a note in the capped direction too.
4. `RUST_LOG` is documented nowhere in `docs/src/`. This change alters the
   meaning of a near-universal convention; it belongs in a short "Logging"
   section in `docs/src/invocation.md`, not only in a CHANGELOG line that will
   scroll away.

Merge order: this is disjoint from the three `update_flow.rs` PRs, so it can go
straight after #459.

Issue #409, opened 2026-07-30
three log-and-swallow sites on the update path


plusky:
Three places on the operation paths where a failure is logged and the flow continues as if nothing happened. Same shape as the install/uninstall silent-no-op already fixed, all lower severity, all small.

## 1. A failed prepare does not stop the update

`reports/update_flow.rs:1310-1317`:

```rust
if !noprepare {
    // Runs prepare with default flags (remove-repo prepare). Prepare is
    // best-effort within the update flow, so a failure is logged, not
    // returned.
    if let Err(e) = perform_prepare(targets, report, packages, false, false, false).await {
        warn!(error = %e, "prepare before update failed");
    }
}
```

The flow then proceeds to `package_check(false)` (`:1319`), `update_lock()` (`:1321`), `fanout_set_repo(RepoOp::Add, …)` (`:1327`) and the patch dispatch. A host that failed prepare — including one that is *down* — gets the issue repo added and a patch fired at it.

"Best-effort" is a defensible design, but it composes badly: with #406's missing prepare check on slmicro/YUM keys, a prepare failure on those hosts is neither judged nor propagated, which is a clean false success. Worth deciding whether this should abort rather than warn.

The `--newpackage` sibling at `:1383-1387` has the same shape. (A note in my backlog claimed that one had been promoted to `error!`; I checked, and it has not — both are `warn!`. There is no asymmetry in the current tree.)

## 2. The success path silently skips repo cleanup

`reports/update_flow.rs:1391-1398`:

```rust
// Success: remove the test update repositories.
if targets.update_lock().await.is_err() {
    warn!("could not lock hosts to remove update repositories; skipping repo cleanup");
    return Ok(());
}
targets.fanout_set_repo(RepoOp::Remove, report).await;
targets.unlock().await;
Ok(())
```

`.is_err()` discards the error value, so the log cannot even say *why* the lock failed. The caller gets `Ok(())` while the test update repos remain configured on every host — a refhost left pointing at a maintenance repo is a trap for whoever picks it up next.

Compare the *failure* path at `:1374-1381`, which handles the same situation properly: it warns, names the remedy (`remove later with 'set_repo --remove'`), and returns `Err(e)`. The success path does neither. At minimum it should keep the error in the message and say what state the hosts are in.

## 3. The `unlock` seam discards the aggregate verdict

`mtui-hosts/src/target/hostgroup.rs:1566-1568`:

```rust
async fn unlock(&mut self) {
[...]

PR #457, opened 2026-08-15
fix(testreport,hosts): record the history rows the aborts owe

### LINKED ISSUE #407: three abort paths leave side effects with no /var/log/mtui.log row
The CHANGELOG claims "a run that dispatched a command and then failed still leaves exactly one row" in `/var/log/mtui.log`. Three abort paths falsify it — each returns after real side effects have landed, and one of them can leave packages installed with no record anywhere.

`add_op_history` is now called from `reports/update_flow.rs:1431` (inside `update_run_phase`, after the dispatch, before the reboot) and `:1122` (in `downgrade_body`); install/uninstall write theirs in `mtui-hosts/src/target/operation.rs:382-384`. All three paths below return before reaching any of those.

## 1. The `perform_update` cancel gate — the one with teeth

`update_flow.rs:1348-1358`:

```rust
if targets.cancel_requested() {
    tracing::info!("cancelled: stopping before the update command was dispatched");
    targets.fanout_set_repo(RepoOp::Remove, report).await;
    targets.unlock().await;
    return Err(UpdateFailure::Cancelled(UpdateError::cancelled(
        "cancelled before the update command was dispatched",
    )));
}
```

By this point, with `!noprepare`, `update_flow.rs:1310-1317` has already run a **full `perform_prepare`** — which installs packages — and `:1327` has added the issue repo. The repo add is undone here. The prepare-installed packages are not.

So: `job_cancel` during a multi-minute prepare leaves packages installed on every host and **no history row anywhere**. The message says "before the update command was dispatched", which is true of the update command and misleading about the host's state. `/var/log/mtui.log` is how the next tester finds out what was done to a refhost; this is the case where it says nothing.

## 2. The `MissingUpdater` abort

`update_flow.rs:1331-1341` — same position, after the repo add and the optional prepare. For the update op itself this is correct behaviour (no updater command was dispatched, so no row is owed). The residue is again the prepare install when `!noprepare`.

There is a test asserting no history for the analogous install case (`update_flow.rs:1928-1950`, `install_writes_no_history_when_no_doer_resolves`), but nothing covering the prepare residue.

## 3. `downgrade_body`'s all-probes-dead abort

`update_flow.rs:947-952`:

```rust
if !dead_probes.is_empty() && dead_probes.len() == list_map.len() {
    return Err(UpdateError::new(
        "package version probe failed",
        dead_probes.iter().cloned().collect::<Vec<_>>().join(", "),
    ));
}
```

`update_flow.rs:889` has already run `fanout_set_repo(RepoOp::Remove, …)`, stripping the issue repo from **every host**. Returns at 948, before `add_op_history` at 1122. This one is likely to fire precisely on the rollback path, i.e. when the fleet is already in a bad state — and it leaves no row.

The later partial-failure path (`update_flow.rs:1138-1143`) does reach the history write, so the abort is the only gap.

## Decision needed

Either narrow the CHANGELOG claim to what actually holds, or write a row on these paths. My read is that (1) genuinely wants a row — packages were installed — while (2) and (3) are arguably correct to stay silent about the *operation* and want the CHANGELOG narrowed instead. Worth a maintainer's call, since "exactly one row" is the kind of claim other tooling may lean on.

plusky:
Closes #407.

## The bug

`/var/log/mtui.log` is what the next tester reconstructs a refhost's state
from — one `timestamp:user:field…` line per operation per host, read back by
`list_history`. Three abort paths returned *after* real side effects had
landed on the hosts, and none of them left a row:

1. **`update` cancelled at its pre-dispatch gate.** By the time the gate runs,
   the initial `prepare` has installed packages on every host. The gate undoes
   the repo add, releases the locks, and returns `Cancelled` — with no row
   anywhere, and a message (`cancelled before the update command was
   dispatched`) that reads as if the hosts were untouched.
2. **`update` aborted for a missing updater** (`MissingUpdater`), after the
   same prepare. Same residue, same silence.
3. **`downgrade` where every version probe died.** `fanout_set_repo(Remove)`
   has already stripped the issue repo from every host by then, and this abort
   fires precisely on the `update` rollback path — the moment where knowing
   what was done to a refhost matters most.

The root cause of 1 and 2 is the same: `prepare` had **no history site at
all**, although installing packages is exactly what it does. 3 is a return
that jumps over `downgrade`'s single post-dispatch site.

## The fix

- **`prepare` writes its own row** (`update_flow.rs::prepare_body`), once at
  least one host actually got an install command dispatched. Placement follows
  the rule `update` and `downgrade` already use: *after* the dispatch, so a row
  never claims work that never started, and *before* `reboot_transactional`,
  because a host that does not come back can no longer be written to. One site
  covers the standalone command, the prepare `update` runs before patching, and
  the `--newpackage` prepare after it — so both `update` aborts above are
  covered where the side effect is produced, not at each abort.
- **Gated on a real dispatch:** an empty package list, a cancel before the
  first package, or a prepare for which no command could be built for any host
  leaves no row. A row claiming an install that never started would be worse
  than none.
- **Scoped to the hosts a command actually reached**, via a new host-scoped
  fan-out `HostsGroup::add_history_for` (and its `add_op_history_for` wrapper).
  `prepare` builds a *per-host* command map, and `build_prepare_map` drops a
  host whose release key does not resolve or whose template does not render —
  the very hosts `prepare_body` then fails with "nothing was installed". A
  group-wide write would hand exactly those hosts a `:prepare:` row
  contradicting their own verdict, in a file the project treats as an interop
  contract.
- **The row names the dispatched subset, not the whole set.** A cancel at
  package `i` of the `--installed-only` loop dispatched `pkgs[..i]` and nothing
  after, so that is what the row records. The error message names the progress
  too, but it is transient; the log line is what an operator coming back to the
  host later reconstructs from.
- **`downgrade`'s all-probes-dead abort writes its row before returning.** The
  post-dispatch site below it is unreachable from that branch, so there is no
  double write.
- **The pre-dispatch cancel message owns the residue:** with a prepare behind
  it, it now says the installed packages are left in place and points at
  `list_history`. Under `--noprepare` it is unchanged.
- **`list_history -e prepare`** now selects the new rows (the value parser
  rejects anything outside `EVENTS`). Additive MCP schema change: the
  `list_history` tool's `event` enum gains `"prepare"`; the slim-profile
  snapshot and the generated CLI reference move with it, since both derive from
  the same array.
- A test in `mtui-hosts` pins that the history fan-out is **never** gated on
  cancellation — these rows are written after `job_cancel` has already
  cancelled the token, so a gate there would suppress the record exactly when
  an operator needs it.

Net effect: a full `update` run now leaves one row per operation that
dispatched — `prepare` then `update`, plus a second `prepare` row when
`--newpackage` is used.

## Judgement calls

- **The `prepare` row names the dispatched prefix; the other ops' rows still
  name the whole set.** This was the issue's "Decision needed". `prepare` can
  answer it because its loop already tracks what it dispatched, so a cancel at
  package `i` records `pkgs[..i]` rather than claiming the rest. `update` and
  `downgrade` are unchanged — narrowing them means threading progress through
  flows that do not track it, and the history format has no field for "how far
  it got" in any case. The accurate applied/not-attempted split is in the error
  the operator sees.
- **The `downgrade` row is written for an abort whose downgrade command never
  dispatched.** Arguably it over-claims. The argument for writing it: the row
  records an *operation that changed the host* — the issue repo was removed
  from every host before the probe ran — and this is the rollback path, where a
  silent record is most damaging. The alternative (a new event label for "repo
  removed") widens the wire contract every mtui on the fleet parses, for a
  strictly rarer case.
- **The message is gated on the flow's own `noprepare` flag, not on sniffing
  the session token** — the same rule the cancellation verdicts follow, so a
  message can never claim a prepare residue that could not exist.
- **`uninstall` is still missing from the `-e/--event` values** although
  uninstall rows exist. Pre-existing gap, unrelated to this abort; left for its
  own change rather than widened into this one.
- **Two doc comments were corrected in passing**, both in the family this issue
  is about: `add_op_history`'s list of ops that reach it (`prepare` was
  missing, and `install`/`uninstall` reach `add_history` through
  `OperationGroup` instead), and `UpdateFailure`'s "nothing was ever installed"
  — which is precisely the claim #407 disproves. The rollback verdicts are
  unchanged; only the prose is.

## Testing

Full DoD gate on the final tree, all green:

| gate | result |
| --- | --- |
| `cargo fmt --all --check` | exit 0 |
| `cargo clippy --workspace --all-targets -- -D warnings` | exit 0 |
| `RUSTDOCFLAGS="-D warnings" cargo doc …` | exit 0 |
| `cargo test --workspace` | 37 suites, 2495 passed, 0 failed, 0 ignored |
| `cargo test -p mtui-mcp -F mcp` | 4 suites, 286 passed, 0 failed, 0 ignored |
| `cargo build --workspace --no-default-features` | exit 0 |
| `cargo build --workspace --all-features` (compile-only) | exit 0 |
| `typos .` | exit 0 |

**How RED was observed.** Every new assertion was checked against a concrete
mutation of the production code, and each run reports its denominator so a
filter that matched nothing cannot pass for a green run (the nine
`update_flow` tests run as `9 tests`, `297 filtered out`):

| mutation | tests that went RED |
| --- | --- |
| delete the `prepare` history write | `prepare_records_history_once_dispatched`, `prepare_records_history_for_a_host_lost_to_its_reboot`, `update_cancelled_at_the_pre_dispatch_gate_records_the_prepare_row` (3/9 failed) |
| move that write *below* `reboot_transactional` | `prepare_records_history_for_a_host_lost_to_its_reboot` (1/9) — existence and placement are pinned separately |
| drop the `!dispatched.is_empty()` gate | `prepare_writes_no_history_when_cancelled_before_any_dispatch`, `prepare_writes_no_history_for_an_empty_package_list` (2/9) |
| delete the `downgrade` abort-site row | `downgrade_all_probes_dead_still_records_history` (1/9) |
| let that row escape its `if` (write on every downgrade) | `perform_downgrade_partial_dead_probe_rolls_back_healthy_host` (1/9) — it counts exactly one row |
| cancel message never mentions the residue | `update_cancelled_at_the_pre_dispatch_gate_records_the_prepare_row` (1/9) |
| cancel message *always* claims a residue | `update_cancelled_at_the_gate_under_noprepare_records_nothing` (1/9) |
| revert `EVENTS` to its five previous values | `prepare_is_a_selectable_event` (1/1) |
| gate `HostsGroup::add_history` on cancellation | `add_history_still_writes_after_the_token_is_cancelled` (1/1) |

Assertions are anchored so they cannot pass on the wrong token: the row asserts
are tail-anchored on `":prepare:pkg-a pkg-b\n"` rather than
`contains("prepare")` (which a package named `prepare-*`, or an `:update:`
row's payload, would satisfy), the negative command assert uses the literal
`:p=42:7` selector `repa_for` builds, and the cancelled-before-dispatch test
also asserts `commands().is_empty()`, so "no row was written" is attributable
to the cancel landing at package 0 rather than to a fixture that never ran
anything. The MCP schema snapshot is compared against
the synthesised schema, so its `"prepare"` entry is evidence about the schema,
not a hand-edit.

## Known limitations

- **`update` and `downgrade` rows are still group-level.** Only the new
  `prepare` row is host-scoped; the pre-existing two are written to every
  enabled host once the op dispatched, so a host the fan-out could not build a
  command for still receives one. Such a host is already named as a failure to
  the operator (#396), but the log line alone does not distinguish it. Bringing
  those two onto `add_history_for` is a behaviour change to existing rows and
  belongs in its own commit.
- **A `prepare` row cannot say a package failed**, only that it was
  dispatched. The wire format has no field for an outcome, so a host whose
  install was dispatched and then failed carries the same row as one where it
  succeeded; the verdict lives in the error the operator sees, not in the log.
- **The write stays best-effort.** `Target::add_history` swallows an append
  failure, so a host with a read-only or full `/var/log` still yields no row
  and no error. That is #409's territory and is deliberately not changed here.
- **The `downgrade` row does not distinguish the abort from a real downgrade.**
  Reading the log alone, an all-probes-dead abort looks like a downgrade that
  ran; the wire format has no place to say otherwise without extending it.
- **Expect a merge conflict with the #451 branch**, which rewrites the same
  `dead_probes` abort in `downgrade_body`. The resolution is mechanical — keep
  this branch's `add_op_history` call at the top of the abort branch — but it
  will not auto-merge.

---

_Description updated by a maintainer at merge time: the host-scoped `add_history_for` and the dispatched-prefix row landed after the original write-up, which still described the group-wide version._

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -416,6 +416,24 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   only that: a rollback that timed out, lost its connection, or never reached
   the host leaves the packages at the update version while the flow ends
   looking done.
+- Aborts that had already changed a host no longer leave `/var/log/mtui.log`
+  silent (#407). `prepare` now records its own history row once it has
+  dispatched an install command — standalone, and for both prepares that run
+  inside `update` (the initial one and `--newpackage`'s after the update) — so
+  an `update` cancelled at its pre-dispatch gate, or aborted for a missing
+  updater, no longer leaves the packages its prepare installed unrecorded on
+  every host. A prepare that dispatched nothing (cancelled before its first
+  package, an empty package list, no command built for any host) still writes
+  no row. A full `update` run therefore now leaves one row per operation that
+  dispatched: `prepare` and `update`. `list_history` gained `prepare` as an
+  `-e/--event` filter value (**MCP schema note:** the `list_history` tool's
+  `event` enum gained `"prepare"`). A `downgrade` aborted because every version
+  probe died now writes its row before aborting: its issue-repo removal had
+  already run on every host, and that abort fires precisely on the rollback
+  path, where reconstructing what was done to a refhost matters most. The
+  cancel message on the update's pre-dispatch gate now says that a completed
+  prepare's packages are left in place, instead of only that the update
+  command never dispatched.
 
 ### Security
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -422,11 +422,19 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   inside `update` (the initial one and `--newpackage`'s after the update) — so
   an `update` cancelled at its pre-dispatch gate, or aborted for a missing
   updater, no longer leaves the packages its prepare installed unrecorded on
-  every host. A prepare that dispatched nothing (cancelled before its first
-  package, an empty package list, no command built for any host) still writes
-  no row. A full `update` run therefore now leaves one row per operation that
-  dispatched: `prepare` and `update`. `list_history` gained `prepare` as an
-  `-e/--event` filter value (**MCP schema note:** the `list_history` tool's
+  every host. The row goes only to the hosts a prepare command actually
+  reached, and names only the packages that were actually dispatched: a host
+  the flow could build no command for keeps its "nothing was installed"
+  failure with no row to contradict it, and a prepare cancelled at package 3
+  of 10 records those three, not all ten. A prepare that dispatched nothing
+  (cancelled before its first package, an empty package list, no command built
+  for any host) still writes no row. On a transactional host the row records
+  what was *staged*: it is written before the reboot that activates the
+  snapshot, deliberately, because a host that never comes back would otherwise
+  leave those packages unrecorded entirely. A full `update` run therefore now
+  leaves one row per operation that dispatched: `prepare` and `update`.
+  `list_history` gained `prepare` as an `-e/--event` filter value
+  (**MCP schema note:** the `list_history` tool's
   `event` enum gained `"prepare"`). A `downgrade` aborted because every version
   probe died now writes its row before aborting: its issue-repo removal had
   already run on every host, and that abort fires precisely on the rollback
--- a/crates/mtui-core/src/commands/listhistory.rs
+++ b/crates/mtui-core/src/commands/listhistory.rs
@@ -10,7 +10,14 @@ use crate::error::{CommandError, CommandResult};
 use crate::session::Session;
 
 /// The event types `-e/--event` accepts.
-const EVENTS: [&str; 5] = ["connect", "disconnect", "install", "update", "downgrade"];
+const EVENTS: [&str; 6] = [
+    "connect",
+    "disconnect",
+    "install",
+    "update",
+    "downgrade",
+    "prepare",
+];
 
 /// Lists the history of mtui events recorded on the reference hosts.
 ///
@@ -123,6 +130,31 @@ mod tests {
         assert_eq!(
             history_command(10, &["install".to_owned()]),
             "tac /var/log/mtui.log | grep -m 10 -e \":install\" | tac"
+        );
+    }
+
+    #[test]
+    fn prepare_is_a_selectable_event() {
+        // `prepare` writes its own `/var/log/mtui.log` rows (#407), so it must
+        // be selectable like every other op: the value parser rejects any
+        // event not in `EVENTS`, and `matches` panics on a rejected argv.
+        let args = matches(&ListHistory, &["-e", "prepare"]);
+        let events: Vec<String> = args
+            .get_many::<String>("event")
+            .expect("the event arg parsed")
+            .cloned()
+            .collect();
+        assert_eq!(events, vec!["prepare".to_owned()]);
+        assert_eq!(
+            history_command(10, &["prepare".to_owned()]),
+            "tac /var/log/mtui.log | grep -m 10 -e \":prepare\" | tac"
+        );
+        // And it is offered by tab-completion, where the prefix is ambiguous
+        // with nothing else in the list.
+        let (session, _buf) = empty_session();
+        assert_eq!(
+            ListHistory.complete(&session, "prep", "list_history -e prep"),
+            vec!["prepare".to_owned()]
         );
     }
 
--- a/crates/mtui-hosts/src/target/hostgroup.rs
+++ b/crates/mtui-hosts/src/target/hostgroup.rs
@@ -2099,6 +2099,44 @@ mod tests {
         );
     }
 
+    /// The history fan-out is never gated on cancellation.
+    ///
+    /// The flows write their `/var/log/mtui.log` row *after* the work landed on
+    /// the host, which for a cancelled operation means after the token was
+    /// cancelled — an `update` stopped at its pre-dispatch gate still owes a row
+    /// for the packages its prepare installed (#407). A cancel check here (or in
+    /// `run_fanout`, whose own contract forbids one) would suppress the record
+    /// exactly when an operator needs it: after a `job_cancel`.
+    #[tokio::test]
+    async fn add_history_still_writes_after_the_token_is_cancelled() {
+        let conn = echo("h1");
+        let handle = conn.clone();
+        let mut g = HostsGroup::new(
+            vec![Target::with_connection(
+                "h1",
+                TargetState::Enabled,
+                Box::new(conn),
+            )],
+            false,
+        );
+        g.cancel_token().cancel();
+        assert!(g.cancel_requested(), "the group observes the cancel");
+
+        g.add_history(&["prepare".to_owned(), "pkg-a".to_owned()])
+            .await;
+
+        let written = String::from_utf8(
+            handle
+                .file_contents("/var/log/mtui.log")
+                .expect("a cancelled group still records its history row"),
+        )
+        .unwrap();
+        assert!(
+            written.ends_with(":prepare:pkg-a\n"),
+            "history line: {written:?}"
+        );
+    }
+
     // --- sftp_read / sftp_put_bytes map fan-outs (#434) --------------------
 
     /// Per-host DISTINCT content: identical fixtures could pass with the
--- a/crates/mtui-hosts/src/target/hostgroup.rs
+++ b/crates/mtui-hosts/src/target/hostgroup.rs
@@ -35,7 +35,7 @@
 //! deterministically ordered by hostname — order-sensitive operations always
 //! iterate in sorted order, without an insertion-order dependency.
 
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, BTreeSet};
 use std::path::Path;
 use std::sync::Arc;
 
@@ -963,6 +963,37 @@ impl HostsGroup {
             max_parallel,
             Some("add_history"),
             |_t| true,
+            |t| {
+                let fields = fields.to_vec();
+                Box::pin(async move { t.add_history(&fields).await }) as actions::BoxTargetFut<'_>
+            },
+        )
+        .await;
+    }
+
+    /// Appends a history entry to *only* the named hosts' remote history files.
+    ///
+    /// The host-scoped counterpart to [`add_history`](Self::add_history), for a
+    /// flow whose work reached some of the group but not all of it. The
+    /// group-wide fan-out is the right default for an op whose command map
+    /// covers every enabled host, but `prepare` builds a per-host map and
+    /// deliberately drops a host whose release key does not resolve or whose
+    /// template does not render — and that same host is then failed with
+    /// "nothing was installed". Writing it a `:prepare:` row anyway would put a
+    /// false entry in `/var/log/mtui.log`, which is an interop contract other
+    /// tools read (#407).
+    ///
+    /// `hosts` is matched against [`Target::hostname`]; a name that is not in
+    /// the group is ignored, and the per-target
+    /// [`enabled`](mtui_types::enums::TargetState) check still applies on top.
+    pub async fn add_history_for(&mut self, hosts: &BTreeSet<String>, fields: &[String]) {
+        let (is_repl, max_parallel) = (self.is_repl, self.max_parallel);
+        actions::run_fanout(
+            &mut self.data,
+            is_repl,
+            max_parallel,
+            Some("add_history"),
+            |t| hosts.contains(t.hostname()),
             |t| {
                 let fields = fields.to_vec();
                 Box::pin(async move { t.add_history(&fields).await }) as actions::BoxTargetFut<'_>
@@ -2134,6 +2165,46 @@ mod tests {
         assert!(
             written.ends_with(":prepare:pkg-a\n"),
             "history line: {written:?}"
+        );
+    }
+
+    /// `add_history_for` writes to the named hosts and to no others.
+    ///
+    /// The distinction the group-wide [`HostsGroup::add_history`] cannot make:
+    /// `prepare` builds a per-host command map, so a host it never dispatched
+    /// to must not be handed a row claiming an install (#407). Both directions
+    /// are asserted — a predicate stuck at `true` fails on h2, one stuck at
+    /// `false` fails on h1.
+    #[tokio::test]
+    async fn add_history_for_writes_only_to_the_named_hosts() {
+        let (c1, c2) = (echo("h1"), echo("h2"));
+        let (h1, h2) = (c1.clone(), c2.clone());
+        let mut g = HostsGroup::new(
+            vec![
+                Target::with_connection("h1", TargetState::Enabled, Box::new(c1)),
+                Target::with_connection("h2", TargetState::Enabled, Box::new(c2)),
+            ],
+            false,
+        );
+
+        let only_h1: BTreeSet<String> = ["h1".to_owned()].into_iter().collect();
+        g.add_history_for(&only_h1, &["prepare".to_owned(), "pkg-a".to_owned()])
+            .await;
+
+        let written = String::from_utf8(
+            h1.file_contents("/var/log/mtui.log")
+                .expect("the named host gets its row"),
+        )
+        .unwrap();
+        assert!(
+            written.ends_with(":prepare:pkg-a\n"),
+            "history line: {written:?}"
+        );
+        assert!(
+            h2.file_contents("/var/log/mtui.log").is_none(),
+            "an unnamed host must not be written to: {:?}",
+            h2.file_contents("/var/log/mtui.log")
+                .map(|b| String::from_utf8_lossy(&b).into_owned())
         );
     }
 
--- a/crates/mtui-mcp/tests/snapshots/it__slim_profile__slimmed_command_tool_schemas_snapshot.snap
+++ b/crates/mtui-mcp/tests/snapshots/it__slim_profile__slimmed_command_tool_schemas_snapshot.snap
@@ -397,7 +397,8 @@ expression: pretty
               "disconnect",
               "install",
               "update",
-              "downgrade"
+              "downgrade",
+              "prepare"
             ],
             "type": "string"
           },
--- a/crates/mtui-testreport/Cargo.toml
+++ b/crates/mtui-testreport/Cargo.toml
@@ -41,6 +41,11 @@ thiserror.workspace = true
 [dev-dependencies]
 insta.workspace = true
 tempfile = "3"
+# `test-util` for `#[tokio::test(start_paused = true)]`: the prepare loop's
+# cancellation checkpoint is a boundary between two awaited fan-outs, and
+# virtual time is the only way to land a cancel *between* two of them without a
+# wall-clock race.
+tokio = { workspace = true, features = ["test-util"] }
 wiremock.workspace = true
 # Perf baseline bench: metadata JSON parse hot path.
 criterion.workspace = true
--- a/crates/mtui-testreport/src/reports/update_flow.rs
+++ b/crates/mtui-testreport/src/reports/update_flow.rs
@@ -49,11 +49,19 @@ type UpdateMaps = (BTreeMap<String, String>, BTreeMap<String, String>);
 /// and re-surfaces the error untouched.
 ///
 /// "No" is not one situation, and the variants keep the differences because an
-/// operator needs them: nothing was ever installed (`MissingUpdater`,
-/// `Prepare`, `Cancelled`, `ProbeFailed`), or something may well have been
-/// installed but the flow cannot reach the host to undo it (`Reboot`), or it is
-/// unknown whether anything was (`NotRun`). All of them collapse to a single
-/// [`UpdateError`] at the command boundary.
+/// operator needs them: no update patch was ever dispatched, so there is
+/// nothing for a downgrade to undo (`MissingUpdater`, `Prepare`, `Cancelled`,
+/// `ProbeFailed`), or the patch may well have applied but the flow cannot
+/// reach the host to undo it (`Reboot`), or it is unknown whether it did
+/// (`NotRun`). All of them collapse to a single [`UpdateError`] at the command
+/// boundary.
+///
+/// "No patch was dispatched" is not the same as "the host is untouched": an
+/// abort that follows a *completed* `prepare` — the pre-dispatch cancel gate
+/// and `MissingUpdater` — leaves that prepare's packages installed. That is
+/// still a "no" for the rollback, which exists to undo a patch that never ran
+/// here, but the host did change, so the prepare writes its own
+/// `/var/log/mtui.log` row rather than leaving the abort silent (#407).
 #[derive(Debug, PartialEq, Eq)]
 pub enum UpdateFailure {
     /// One or more hosts failed the `updater` check after the command ran.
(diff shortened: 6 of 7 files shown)


mimi1vx (changes requested):
The core of this is right — an abort that changed a host owes a record, and both
the placement (after dispatch, before `reboot_transactional`) and the
"dispatched nothing → no row" inverse are well argued and well pinned. Two
things at the new `prepare_body` site need fixing before this lands, both
instances of the rule the PR itself states: *a row claiming an install that
never started is worse than no row.*

### 1. The row is written to hosts that were never dispatched to

`update_flow.rs`, the new block:

```rust
if !dispatched.is_empty() {
    add_op_history(targets, "prepare", None, pkgs).await;
}
```

`dispatched` is a **per-host** `BTreeSet` built by `note_dispatch`, but the gate
is a group-wide `is_empty()` and `HostsGroup::add_history` fans out to *every
enabled host*. `build_prepare_map` deliberately drops a host whose release key
does not resolve or whose template does not render, and the very next block in
this function fails exactly those hosts:

```rust
failures.push(UpdateError::new(
    "no prepare command could be built for this host; nothing was installed",
    host.to_owned(),
));
```

So on a mixed group, a host can simultaneously be told *"nothing was installed"*
and have `<ts>:<user>:prepare:pkg-a pkg-b` appended to its `/var/log/mtui.log`.
That is a false entry in a format the project treats as an interop contract, and
it is the precise failure mode `prepare_writes_no_history_when_cancelled_before_any_dispatch`
exists to prevent — just at per-host rather than per-group granularity.

The information to do this correctly is already in hand. Either gate the fan-out
on `dispatched` membership, or add a host-filtered variant of `add_history`.

A test that would catch it: two hosts, one with an unresolvable release key, and
assert the unresolvable one's `HISTORY_LOG` is `None` while the healthy one's
is written.

(`update`/`downgrade` share the whole-group fan-out, but they never compute a
per-host dispatch set, so this is the one site where the gap is both knowable
and newly introduced.)

### 2. The row over-claims on a partially cancelled prepare

Same site: the row names `pkgs` — the **whole** prepare set — even when the
per-package loop broke at `cancelled_at = Some(3)` of 10. The comment
acknowledges this:

> The row names the whole prepare set, as `update`/`downgrade` do; a per-package
> cancel names its own progress in the error.

But the error is transient and this log line is the durable artifact an operator
reconstructs host state from — that is the whole justification for the PR. On a
cancelled prepare the error is precisely what the operator no longer has when
they come back to the host later. Writing the dispatched subset costs nothing
here, and `cancelled_at` already tells you where to slice.

---

### Non-blocking notes

**`list_history` grep pattern.** `grep -m N -e ":prepare"` matches anywhere in
the line, so an `:update:` row whose payload names a package like
`prepare-tools` would match. Pre-existing for every event value and not made
worse here — just noting it.

**Transactional hosts.** On slmicro the row is written before
`reboot_transactional`, so it records *staged*, not *active*. That is the right
trade (a host that does not come back cannot be written to) and
`prepare_records_history_for_a_host_lost_to_its_reboot` pins it deliberately.
Worth one sentence in the block comment so a future reader does not read the row
as proof of an activated snapshot.

**Contract-compat.** Adding `prepare` as an event label is additive — older mtui
releases parsing `/var/log/mtui.log` still split the line fine and simply never
filter for it. The MCP `list_history` `event` enum gaining a value is likewise
additive. Both look safe; the CHANGELOG calls out the schema note, which is the
right call.

**Merge order.** This conflicts textually with #458, which rewrites the same
`prepare_body` hunk (it removes the post-loop `run_checks` call and edits the
"Surface any per-host command failure" comment this PR inserts above). It also
brushes #454 in `downgrade_body`'s all-probes-dead region. Worth agreeing an
order.

There is also a semantic interaction with #454: that PR widens the
all-probes-dead abort from "only the `-1` sentinel" to "any non-zero probe
status", so the `:downgrade:` row this PR adds at that site will be written in
strictly more cases where zero downgrade commands ran. The justification (the
issue-repo removal already landed) still holds, but the label over-claims more
often once both are in. Worth a line in whichever lands second.

plusky:
Both blocking points are correct — no pushback on either. Fixed in `6d005369`, with the #454 note in `1df3cbe2`. Both are new commits on top of `3dace329`; nothing already pushed was rewritten.

## 1. A row on a host the prepare never dispatched to (blocking) — fixed in `6d005369`

The contradiction was reachable in a single run, exactly as you described. `dispatched` is a per-host set (`note_dispatch`, `update_flow.rs:1087`, inserts `cmd.keys()`), while `HostsGroup::add_history` fanned out with `|_t| true`. `build_prepare_map` deliberately drops a host whose release key does not resolve (`update_flow.rs:1116`) or whose template does not render (`update_flow.rs:1140`), and the block right below fails those same hosts with `no prepare command could be built for this host; nothing was installed` (`update_flow.rs:1019`). On a mixed group, one host could be told nothing was installed on it and, in the same run, get `<ts>:<user>:prepare:pkg-a pkg-b` appended to its `/var/log/mtui.log`.

You offered two shapes; I took the host-filtered variant rather than gating the fan-out inline, so the filter stays atomic with the row construction:

- `HostsGroup::add_history_for(&BTreeSet<String>, &[String])` (`crates/mtui-hosts/src/target/hostgroup.rs:989`) differs from `add_history` (`hostgroup.rs:958`) only in the `run_fanout` `should_run` predicate — `|t| hosts.contains(t.hostname())` instead of `|_t| true` (`hostgroup.rs:996`). The per-target `Enabled` gate in `Target::add_history` (`crates/mtui-hosts/src/target/mod.rs:832`) still applies on top, so the two compose and disabled hosts stay excluded as before.
- `add_op_history_for` (`update_flow.rs:323`) routes the prepare row through it. The field list is factored out into `op_history_fields` (`update_flow.rs:335`) and shared with `add_op_history`, so the two spellings of the row cannot drift in wire format.
- The write is now `add_op_history_for(targets, &dispatched, "prepare", None, recorded)` (`update_flow.rs:967`). It is keyed off the same `dispatched` set that feeds the "nothing was installed" block and the reboot filter, so the two verdicts cannot disagree by construction rather than by coincidence.

## 2. Over-claiming the package set on a partial cancel (blocking) — fixed in `6d005369`

The row now names `pkgs[..i]` when `cancelled_at == Some(i)` and `pkgs` otherwise (`update_flow.rs:962-965`). No extra bookkeeping was needed: `cancelled_at` is set at the checkpoint *before* package `i` is dispatched, so the slice is exactly the dispatched prefix, and the `--installed-only` loop is the only path that can set it (the single-transaction branch dispatches once or not at all, so `recorded == pkgs` is right there).

`cancelled_at == Some(0)` is unreachable at the write, because `dispatched` is empty then and the `!dispatched.is_empty()` gate already covers it — so no empty-payload row exists. The error message keeps naming the progress as well; the difference is that the error is transient and the log line is what an operator reconstructs from later.

## 3. Transactional hosts: staged, not active (non-blocking) — documented in `6d005369`

Stated in the block comment at the write site (`update_flow.rs:928-934`) and in the CHANGELOG entry, phrased as the reason for the placement rather than as a caveat: the row is written before `reboot_transactional` deliberately, because a host that never comes back would otherwise leave those packages unrecorded entirely. The existing `prepare_records_history_for_a_host_lost_to_its_reboot` (`update_flow.rs:2783`) already pins that placement, so it is not duplicated.

## 4. Interaction with #454 (non-blocking) — documented in `1df3cbe2`, with one correction

Carried here as a separate docs commit at the abort site (`update_flow.rs:1258-1266`), since it is a different concern in a different function.

One correction to the mechanism, though the conclusion is right. The `lastexit != 0` test is not what #454 changes: that gate is already non-zero on `main` and appears verbatim as context in `git diff main...fix/451-downgrade-probe-status` — the only edits in that region are a log-message reword and the new `downgrade_verdict(targets, &dead_probes)` argument. The widening happens in the probe *template* (`crates/mtui-testreport/src/update_workflow/actions/downgrade.rs` on that branch), which starts guarding each stage and exiting with the failing tool's own status. Today the pipeline's status is awk's, and awk exits `0` over empty input, so in practice only the `-1` SSH-death sentinel ever reaches this gate; afterwards a broken `zypper se` reaches it too. So your conclusion holds exactly as stated — this row gets written in strictly more zero-commands-ran cases — and that is the point of writing it: those are the aborts that leave a host with its issue repo removed and nothing rolled back. The committed comment says it that way.

## 5. `list_history`'s grep pattern (non-blocking) — left as is

Pre-existing, and not made worse by this change: the new `prepare` event uses the same `:<event>:` shape as `update`/`downgrade`, so it is neither a new instance of the pattern nor an aggravation of it. Better as its own change with its own test than folded into this one.

## New tests, and how each was observed red

Each was re-run against a hand-broken tree (the fix reverted in place, everything else identical) and observed failing before being claimed:

| Test | Mutation | Observed failure |
| --- | --- | --- |
| `add_history_for_writes_only_to_the_named_hosts` (`hostgroup.rs:2179`) | predicate → `\|_t\| true` | `an unnamed host must not be written to: Some("…:prepare:pkg-a\n")` |
| same test, other direction | predicate → `\|_t\| false` | `the named host gets its row` (the `expect` on h1's file) |
| `prepare_records_history_only_on_the_hosts_it_dispatched_to` (`update_flow.rs:2634`) | write reverted to group-wide `add_op_history` | `a host nothing was dispatched to must have no prepare row: Some("…:prepare:pkg-a\n")` |
| `prepare_cancelled_mid_loop_records_only_the_dispatched_packages` (`update_flow.rs:2707`) | `recorded` → whole `pkgs` | `the row must name only the dispatched packages: "…:prepare:pkg-a pkg-b pkg-c pkg-d\n"` |

The fixtures can express those mutations because the two hosts are distinguishable (h2 has no parsed system, so `host_key` returns `None`) and each has its own `MockConnection` handle — a shared mock could not fail these. The absence assertions are backed by a positive control in the same test (the healthy host's row *is* written), so a fixture that silently wrote nothing at all could not pass. The history assertions use `ends_with(":prepare:pkg-a pkg-b\n")` plus `lines().count() == 1` rather than `contains("prepare")`, which would also match a package named `prepare-something` or another op's payload. The mid-loop test additionally pins the interleaving itself with a full-string `assert_eq!` on `err.reason` and `handle.commands().len() == 2`, so a change that moved the cancel to a different package fails loudly instead of silently re-aiming the history assertion.

## New dev-dependency

`tokio = { workspace = true, features = ["test-util"] }` in `crates/mtui-testreport/Cargo.toml`, with a comment saying why. The partial-cancel test has to land a cancel *between* two awaited per-package fan-outs; every existing cancel test in this module cancels at index 0, and `MockConnection` has no callback hook, so wall-clock sleeps were the only alternative and would be flaky under a loaded parallel run. `#[tokio::test(start_paused = true)]` with a 10 ms `with_run_delay` and a 15 ms canceller makes the interleaving exact — 25 consecutive runs of that test were green. `mtui-hosts` already declares the same dev-feature (`ssh.rs` uses `start_paused`) and the workspace `tokio` is already `features = ["full"]`, so no `Cargo.lock` change resulted.

## Sibling sites I audited but did not change

Since the defect class is "a group-wide row over a per-host dispatch map", I checked the other three history sites:

- **`install` / `uninstall`** — cannot have it. `OperationTemplate::run` resolves plans first (`operation.rs:380`), and `HostsGroup::plans` (`hostgroup.rs:1669`) is all-or-nothing: the first host with no release or no doer makes the whole call `Err` and nothing runs anywhere. `collect` then emits one command per plan (`operation.rs:334-346`), so the row at `operation.rs:403` covers exactly the dispatched set. There is also exactly one write per run — no double.
- **`update`** — same shape: `build_update_maps` (`update_flow.rs:1905`) `?`s out on the first host it cannot render for, so the map covers every host or the flow aborts before dispatch. The row at `update_flow.rs:1804` is therefore exact.
- **`downgrade`'s happy-path row** (`update_flow.rs:1442`) — this one *is* group-wide over maps that can skip a host (a host whose probe resolved no versions for any requested package never enters `cmd`/`combined`). That row is pre-existing on `main` (`main:…/update_flow.rs:1345`) and is untouched by this PR, so I have left it alone rather than widen the reviewed diff; flagging it here so it is on the record rather than found later.

## Gate

Full DoD gate, observed green on `1df3cbe2` after the mutations were reverted:

- `cargo fmt --all --check` — clean
- `cargo clippy --workspace --all-targets -- -D warnings` — clean
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features --document-private-items` — clean
- `cargo test --workspace` — 2498 passed, 0 failed
- `cargo test -p mtui-mcp -F mcp` — 286 passed, 0 failed
- `cargo build --workspace --no-default-features` and `--all-features` — both build
- `typos .` — clean; `git status --porcelain` — clean

## CHANGELOG

The `[Unreleased]` entry this PR added in `3dace329` was edited in place rather than appended to. The old text said the row goes to "every host" and named the whole package set, which the fix makes false; a second entry would have contradicted the first. No released section was touched. The entry now carries the per-host scoping, the dispatched-subset rule with your 3-of-10 example, and the staged-not-active sentence.

plusky:
On the merge order both this review and #458's asked about:

**Proposal: #454 → #458 → #457** (with #456 and #459 independent, any time).

- #454 first so the widened all-probes-dead abort is in before the history row lands on top of it — #457 already carries the doc line describing that interaction (`docs(testreport): note how #454 widens the downgrade abort that writes this row`), so it is written for the world where #454 is already in.
- #458 before #457 because #458 rewrites `prepare_body`'s check plumbing while #457 only inserts into it — rebasing the insertion over the rewrite is the small conflict; the other direction rebases a rewrite over an insertion.

I'll rebase whichever of these is left after each merge, as usual.

mimi1vx (dismissed):
LGTM. `/var/log/mtui.log` is an interop contract every mtui on the fleet reads,
and `prepare` genuinely had no history site at all despite being the op that
installs the packages — so all three abort paths returned after real side effects
with no record. The placement rule is right and matches what `update` and
`downgrade` already do: after the dispatch so no row claims work that never
started, and before `reboot_transactional` because a host that does not come back
can no longer be written to.

The `mtui-hosts` test pinning that the history fan-out is **never** gated on
cancellation is exactly the guard this needed — these rows are written after
`job_cancel` has already fired the token, so a gate there would suppress the
record precisely when an operator needs it.

**Please update the PR description before this merges.** Known limitations still
says:

> The row is group-level, not per-host. A host that the fan-out could not build a
> command for still receives a row naming the package set.

The diff does the opposite: it adds `HostsGroup::add_history_for` and scopes the
write to `dispatched`. Same for the mid-loop `--installed-only` cancel, which now
names `pkgs[..i]` rather than the whole set — better than the body claims, and
worth saying in the CHANGELOG since that is the accurate-progress behaviour an
operator would rely on. The body also does not mention the new public
`HostsGroup` API.

Non-blocking:

1. Adding `"prepare"` to `EVENTS` is additive for the MCP schema and leaves the
   *line* format unchanged, so older mtui releases on the same fleet still parse
   the file — but their `list_history -e prepare` will reject the value. One
   CHANGELOG line so fleet operators are not surprised.
2. The `:downgrade:` row written for an abort where no downgrade command
   dispatched does over-claim. I agree the alternative — a new event label
   widening the wire contract — is worse, and the argument that the issue repo
   *was* removed so the host did change is sound. But this is the one judgement
   call in the batch I would rather sign off explicitly than accept by default,
   so: noted and accepted.
3. `uninstall` is still missing from `-e/--event` although `uninstall` rows
   exist. Correctly out of scope; please file it.

Merge order: this is the hardest of the three `update_flow.rs` PRs to rebase, so
it should go last. Its downgrade-abort row has to land inside #454's widened
abort branch, and its `prepare_body` history write has to be placed relative to
#458's `note_check` / `dispatched` bookkeeping — that second resolution is the
non-mechanical one. Your Known-limitations note flags the #454 conflict but not
the #458 one.

plusky:
Rebased onto main after #459's merge — the conflict was CHANGELOG-only (both entries kept, order preserved). The reviewed delta is unchanged: `git diff` of old-base..old-tip vs main..`66ef3a92` differs only in the CHANGELOG hunk's line offset, zero code differences. The push dismissed the approval; re-requesting.

mimi1vx (dismissed):
Re-approving: the force-push was a pure rebase onto `main` after #459 landed —
I diffed the old and new heads and the only delta is the CHANGELOG hunk offset,
with no change to any source file. All checks green on the new head.

My earlier review stands unchanged, including the outstanding items in it.

plusky:
Rebased onto main after #458 — this one was a real conflict, not offset-only: #458 rewrote `prepare_body`'s failures collection (per-fan-out `note_check` + the `check_failed` attribution skip), and this PR's history-row block had to re-seat into that shape. Resolution: the row block sits between the dispatch loop and #458's filtered collection; this branch's obsolete pre-#458 post-loop `run_checks` line was dropped in favour of main's per-fan-out collection.

Verification: the patch payload — every `+`/`-` line of the branch delta — is byte-identical to what you approved (762 lines, compared mechanically); only hunk offsets and context lines changed. Local gate re-run on the result: fmt clean, clippy 0, workspace 37 suites / 2553 passed / 0 failed. The push dismissed the approval again; re-requesting — the one spot worth a look is the resolved region in `prepare_body`.

mimi1vx (approved):
Re-approving after the 09:43 rebase onto `main` (post-#454, post-#458). This was
the non-mechanical resolution of the three, so I verified it rather than trusting
`MERGEABLE`:

- **`prepare_body`** — `note_dispatch` and `note_check` both stay inside the
  per-package loop (#458), the `add_op_history_for(&dispatched, …)` write lands
  after the loop and before `reboot_transactional` (#407's placement rule), and
  #458's `check_failed` filter on `host_command_failures` plus
  `inert.extend(check_failed)` survive intact.
- **`downgrade_body`** — the history row sits at the top of the all-probes-dead
  abort branch, inside #454's widened `dead_probes` gate, with
  `downgrade_verdict(targets, &dead_probes)` preserved.

Local gate on the rebased tree: `fmt --check` clean, `clippy --workspace
--all-targets -D warnings` clean, `cargo test --workspace` 37/37 suites green.
CI green on `1d2be45e`.

Description updated to match the diff (the host-scoped `add_history_for` and the
dispatched-prefix row were still described as group-wide).

**One follow-up, not blocking.** The rebase left a stale rationale in
`downgrade_body`:

> Merge note (#451/PR #454): … Today the pipeline's status is awk's, and awk
> exits `0` over empty input, so this gate can in practice only catch the `-1`
> SSH-death sentinel; afterwards a broken `zypper se` reaches it too. This row is
> therefore written in strictly more zero-commands-ran cases **once #454 lands**.

#454 has landed — this comment now describes the pre-#454 world as the present
and is actively misleading about what the gate catches. Per AGENTS.md a retired
rationale gets replaced, not left standing: it should become a plain statement
("the probe template exits with the failing tool's own status, so this gate
catches a broken `zypper se` as well as the `-1` sentinel"). Happy to take it as
a one-line docs PR if you'd rather not respin.

PR #458, opened 2026-08-15
fix(testreport): give prepare/install/uninstall and downgrade a verdict on slmicro and YUM

### LINKED ISSUE #406: prepare/install/uninstall have no check entry for slmicro or YUM keys
`update` gained checks for `("slmicro", true)` and `("YUM", false)` in #393. The sibling roles still have none, so a transactional or RHEL host contributes no `prepare`, `install` or `uninstall` verdict at all.

What makes this more than a coverage note: **the check modules' own tests currently assert the holes exist.**

- `crates/mtui-testreport/src/update_workflow/checks/prepare.rs:96` — `assert!(prepare_check("YUM", false).is_none());`
- `crates/mtui-testreport/src/update_workflow/checks/install.rs:123` — `assert!(install_check("slmicro", true).is_none());`

So closing the gap means deliberately inverting two assertions, which is the right way round — but worth knowing before someone treats a red test as a regression.

## The gaps

**`prepare_check`** — `checks/prepare.rs:43-48`:

```rust
pub(crate) fn prepare_check(release: &str, transactional: bool) -> Option<CheckFn> {
    match (release, transactional) {
        ("11", false) | ("12", false) | ("15", false) | ("16", false) => Some(Box::new(zypper)),
        _ => None,
    }
}
```

The *action* table has both missing keys (`actions/prepare.rs:87-88`), so the gap is reachable: mtui will happily run a prepare it cannot judge. #393's prepare reboot gate therefore leans on the non-zero-exit term rather than a check verdict. Prepare's templates *are* single commands, so the exit code is genuine there — but an exit-0-with-lock-message failure is still missed.

**`install_check`** — `checks/install.rs:54-59`: same four zypper keys, while `actions/install.rs:34-35` and `actions/uninstall.rs:32-33` both register `YUM` and `slmicro`.

**`downgrade`'s transactional arm** — `checks/downgrade.rs:65-86` gates on `exitcode == -1` only, by design (its doc explains transactional-update's codes differ from zypper's). A `transactional-update` exiting 1 passes the check. #393's reboot gate now catches that for the *reboot* decision, but the check itself stays silent. `downgrade.rs` also still has no `("YUM", false)` entry.

## What happens with no check entry

Two different fallbacks, worth knowing because they behave differently:

**install/uninstall** go through the `PlanProvider` adapter — `update_workflow/mod.rs:344-356`:

```rust
None if a.exitcode != 0 => Err(format!("{op} command failed")),
None => Ok(()),
```

So they get exit-code-only classification: no locked-stack, dependency-error or RPM-error attribution.

**prepare/downgrade** do not go through that adapter. `reports/update_flow.rs:316-318` **silently skips** the host:

```rust
let Some(check): Option<CheckFn> = registry.check(role, &release, transactional) else {
    continue;
};
```

with only the coarser `host_command_failures` scan (`update_flow.rs:370-380`, non-zero exit **or** non-empty stderr) behind it.

## Constraint on any fix

Do **not** adopt a `!stderr.is_empty()` rule on slmicro/YUM. `transactional-update` and `yum` both write progress and warnings to stderr on a successful run, and this is pinned — `reports/update_flow.rs:1809-1838`:

```rust
async fn perform_install_ignores_stderr_on_a_clean_exit() {
    // `transactional-update` and `yum` write progress and warnings to stderr
    // on a successful run. Treating any stderr as failure would fail every
    // SL Micro install.
```

That test's host is SL-Micro/transactional, so it hits exactly the `None` fallback arm — any fix that makes the fallback stderr-sensitive breaks it.

## Suggested shape

Give prepare/install the same marker-based treatment `update` now has, factoring the shared classification. The `markers` fn added to `checks/update.rs` in #393 is the model: it reads only stdout/stderr, so it is valid on every key, with exit-code handling left to per-key wrappers. Note the `Error:`-in-stderr marker carries the sibling-command caveat documented there — worth re-reading that reasoning before reusing it on a non-zypper transcript.

plusky:
Closes #406.

## The bug

`("slmicro", true)` and `("YUM", false)` have had a preparer, an installer and an uninstaller since the port — and `("YUM", false)` a downgrader too — but no post-run check for any of those roles. Both check modules' own `table_lookup` tests asserted that hole rather than flagging it.

The consequences differ by path:

- **`prepare`** resolves its check through `run_checks`, which silently `continue`s when the table answers `None`, so such a host contributed no attributed verdict at all. Worse, the per-host reboot gate's skip set is seeded from check verdicts — so a `transactional-update` that reported `System management is locked` **and still exited `0`** was invisible to every rule the flow had. The exit-code rule sees `0`; the stderr rule is deliberately excluded from the gate, because `transactional-update` writes progress to stderr on a *successful* run and skipping such a host's reboot would strand a healthy staged snapshot. The host rebooted straight into the locked transaction while the flow reported success.
- **`install`/`uninstall`** fell through to the `PlanProvider` adapter's exit-code-only fallback, which answered `"install command failed"` for a locked update stack, a failed RPM transaction and a command that never ran alike.
- **`downgrade`** on the YUM key was the last role with a rollback command and no check: a `yum -y downgrade` that never ran to completion left the host on the update version while the flow ended looking done.

## The fix

Two commits; the second is independent and droppable.

**1. `fix(testreport): give prepare/install/uninstall a verdict on slmicro and YUM`**

New check functions for both keys in `checks/prepare.rs` and `checks/install.rs` (the install table also serves `uninstall`), registered in the two tables.

On the transactional key both new checks reuse `checks::update`'s shared classification — `markers` for prepare, `classified` for install/uninstall — rather than restating it, so the two `transactional-update` keys cannot drift apart; those two helpers become `pub(super)` for it. The `Error:`-from-a-sibling-command caveat the markers carry on the update *template* does not arise here: a prepare or install command is a single invocation, so the whole transcript belongs to the command being judged.

The prepare verdict is now taken after **every** fan-out rather than once at the end, which is what makes the reboot gate complete on the `--installed-only` path. `run_checks` reads the same `last*` snapshot `note_dispatch` does, so a single post-loop call judged only the last package — very often a clean no-op `rpm -q` skip — and an exit-`0` lock message on an earlier package was overwritten by it. `note_check` is `note_dispatch`'s companion: same call sites, scoped to that fan-out's own hosts, first-failure-wins per host. Same shape the downgrade loop already uses.

`aggregate_failures` also names each host once. A single host legitimately contributes two failures here (the stderr scan and the check that recognised the marker), which rendered as `prepare failed on h1, h1`; the detail list still carries both causes.

**2. `fix(testreport): give downgrade a verdict on the YUM key`**

The same narrow `-1`-only gate the transactional downgrade check carries, for the same reason: a non-zero status is a routine outcome on a refhost carrying only part of the update's package list, but a command mtui never got a status from is not a completed rollback.

## Judgement calls, argued

- **The YUM verdict is deliberately narrower than zypper's, and asymmetric between roles.** `install`/`uninstall` read the exit code; `prepare` reads only the `-1` sentinel. That is not an oversight: `prepare --installed-only` renders `rpm -q $package &>/dev/null && yum … install $package`, which exits `1` on the routine "this host does not carry the package" skip — the case the flag exists to serve — indistinguishably from a `yum` that failed. The zypper and slmicro preparers use `if …; then …; fi`, which exits `0` on the same skip, which is why they are not exposed to this.
- **zypper's informational band is not transferred to `yum`/`dnf`.** `100`–`107` are zypper's vocabulary; `dnf`'s only documented `100` belongs to `check-update`, which no doer runs, so accepting it would pass a genuinely failed transaction. Pinned in both directions.
- **No markers on the YUM key.** Three of the four marker strings are zypper-only vocabulary and the fourth (`Error:` on stderr) was written against zypper transcripts. A successful `yum install` that printed `Error:` must not fail. `"Unknown Error"` on the YUM install key is therefore an admission, not a diagnosis — naming the failure needs observed `yum`/`dnf` output from a real RHEL refhost, not inference.
- **No `!stderr.is_empty()` rule anywhere.** Both package managers write progress to stderr on a successful run. This is the issue's own hard constraint and it is pinned in both directions (a marker-carrying stderr fails; a progress-carrying stderr passes and still reboots).
- **No bare non-zero-exit arm in the slmicro *prepare* check.** `prepare_body` already owns exit codes: `note_dispatch` feeds them to the reboot gate and `host_command_failures` reports them. A second verdict on the same signal would name the host twice and push `aggregate_failures` out of its single-failure verbatim branch, where `host` survives. Adding the arm is not "completing" the check; it duplicates a verdict the flow already renders.
- **Scoping `note_check` to the fan-out's hosts changes one adjacent behaviour, on purpose.** A host outside the fan-out is no longer judged at all, where the post-loop call would have judged whatever its `last*` still held — for an empty package list, the `set_repo` transcript. Inventing a prepare verdict from another phase's record is worse than abstaining, and hosts no command was built for are already failed by name (#396).
- **The adapter's `None` fallback is now unreachable through any real key, and stays.** It is the safety net for a doer added before its check, and it gained direct unit coverage so it cannot silently rot.
- **Reason strings an operator or MCP client sees change.** The typed failure a check routes to — which is what decides whether a group is rolled back — is unchanged, and nothing in `mtui-mcp` matches on the old strings. Called out in the CHANGELOG.

## Testing

`mtui-testreport` lib goes 313 → 317; full workspace 2504 passed / 0 failed across 37 suites, `-F mcp` 286 / 0 across 4. fmt, clippy `-D warnings`, rustdoc `-D warnings`, both feature-matrix builds and `typos` all clean.

Every new assertion was hand-broken and observed RED, not assumed:

- **`perform_prepare_judges_the_markers_of_every_package_not_just_the_last`** — deleted the in-loop `note_check` call (exactly the pre-fix shape). RED with `h1's pkg-a hit a locked stack; it must not reboot into it: ["systemctl reboot"]`, i.e. the host really did reboot into the locked transaction. The reboot assert is placed *before* the reason assert so the reboot claim owns the red. Attribution holds because every command in that fixture exits `0` with empty stderr except pkg-a's, so the check verdict is the only mechanism that can seed the reboot gate — the `with_default` trap from previous sessions is defused by giving one command its own response. Clean h2 in the same group must still reboot, so an empty `fired` list cannot fake the pass.
- **`the_new_arms_resolve_to_their_own_check_not_a_sibling`** (prepare + install) — applied the substitution the review named, `("YUM", false) | ("slmicro", true) => Some(Box::new(transactional_update))`, in both files. 70 passed / 2 failed: only the two new tests caught it, confirming that `is_some()` plus direct-call behaviour tests could not.
- **`aggregate_failures_names_each_host_once`** — removed `hosts.dedup()`. RED with the literal `prepare failed on h1, h1 (h1: prepare command failed; h1: update stack locked)`.
- The pre-existing `--installed-only` exit-code test and both stderr-direction flow tests still pass unchanged, so the new in-loop placement did not quietly re-open the half that was already closed.

## Known limitations

- **The slmicro *downgrade* check is still marker-blind.** An exit-`0` `transactional-update … --oldpackage` that reported a locked stack still reboots on the rollback path. The exit gating there is deliberate (a non-zero downgrade status is routine on a refhost carrying part of the package list), but the marker half is the same hole this PR closed for `prepare` and deserves its own issue — possibly folded into the #451 rollback-path work. Deliberately out of scope here.
- **The YUM `--installed-only` template still exits `1` on the routine skip.** The check correctly abstains, but `host_command_failures`' exit-code scan at flow level still reports that host, so a RHEL host legitimately skipping a package can fail the prepare. The fix is in the template (use the `if …; then …; fi` shape the other two families use), not in the check; separate issue.
- `"Unknown Error"` on the YUM install/uninstall key names nothing useful. Improving it needs a real transcript from an RHEL refhost.
- Coverage of the new code is by unit + flow tests only; no live SL Micro or RHEL host was involved, so the transcripts are the templates the flow really renders plus documented marker strings — not captured output.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -380,6 +380,34 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   templates` count now reports templates actually completed instead of counting
   never-attempted ones as done — as does the `succeeded` log line, which no
   longer names templates the stop never reached.
+- `prepare`, `install` and `uninstall` now judge SL Micro (`transactional-
+  update`) and RHEL (`yum`/`dnf`) hosts (#406). Both keys had commands for all
+  three roles but no post-run check: `prepare` skipped such a host silently, so
+  only the coarse "non-zero exit or any stderr" scan spoke for it, and
+  `install`/`uninstall` fell back to the exit code alone, reporting every
+  failure as "install command failed" with no further attribution. On SL Micro
+  the checks now read the same output markers the `update` check uses, so a
+  locked update stack, an unresolved dependency prompt or an RPM error is named
+  — and, the headline case, a `prepare` that reported a locked stack **and
+  still exited `0`** no longer reboots the host into the failed transaction. On
+  RHEL the verdict stays deliberately narrow: `install`/`uninstall` judge the
+  exit code (zypper's informational `100`-`107` are *not* transferred to
+  `yum`/`dnf`), while `prepare` judges only whether the command ran, because
+  `prepare --installed-only` renders `rpm -q <pkg> && yum … install <pkg>`,
+  which exits `1` on the routine "this host does not carry the package" skip.
+  A command that never ran to completion (a timeout, a dropped connection, an
+  unconnected host) now gets its own verdict on every one of these keys instead
+  of passing on an empty transcript. Neither check treats stderr as a failure:
+  both package managers write progress there on a successful run. The verdict
+  is taken after **every** fan-out, so `prepare --installed-only` — which runs
+  one command per package — is judged on each package rather than only the
+  last, whose clean no-op used to mask an earlier one.
+  **Reason strings an operator or MCP client sees change accordingly** — the
+  typed failure a check routes to, which is what decides whether a group is
+  rolled back, is unchanged. A multi-failure summary also names each host once
+  now: a single host can contribute two failures (a lock message is reported
+  both by the stderr scan and by the check that recognised it), which used to
+  read "prepare failed on h1, h1".
 
 ### Security
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -408,6 +408,12 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   now: a single host can contribute two failures (a lock message is reported
   both by the stderr scan and by the check that recognised it), which used to
   read "prepare failed on h1, h1".
+- A `downgrade` on a RHEL host is no longer reported as completed when its
+  `yum -y downgrade` never ran (#406) — the last `(release, transactional)`
+  key with a downgrade command and no check. As on SL Micro, the check judges
+  only that: a rollback that timed out, lost its connection, or never reached
+  the host leaves the packages at the update version while the flow ends
+  looking done.
 
 ### Security
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -404,10 +404,12 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   last, whose clean no-op used to mask an earlier one.
   **Reason strings an operator or MCP client sees change accordingly** — the
   typed failure a check routes to, which is what decides whether a group is
-  rolled back, is unchanged. A multi-failure summary also names each host once
-  now: a single host can contribute two failures (a lock message is reported
-  both by the stderr scan and by the check that recognised it), which used to
-  read "prepare failed on h1, h1".
+  rolled back, is unchanged, and so is the failing host's name: where a check
+  and the coarse scan both speak for one host, `prepare` now reports the
+  check's sharper reason alone rather than summarising two, which would have
+  dropped the host field the summary form cannot carry. A summary over
+  genuinely distinct causes on one host also names it once now, instead of
+  reading "downgrade failed on h1, h1".
 - A `downgrade` on a RHEL host is no longer reported as completed when its
   `yum -y downgrade` never ran (#406) — the last `(release, transactional)`
   key with a downgrade command and no check. As on SL Micro, the check judges
--- a/crates/mtui-testreport/src/reports/update_flow.rs
+++ b/crates/mtui-testreport/src/reports/update_flow.rs
@@ -397,14 +397,42 @@ fn build_reboot_map(
 ///
 /// The check reads each host's `last*` snapshot after the command ran. Only the
 /// `update` check currently emits diagnostics; the other roles append nothing.
+///
+/// Every host in the group is judged. A caller that fans out to a *subset* —
+/// the per-package `prepare` and `downgrade` loops — must use
+/// [`run_checks_where`] instead of filtering the returned list: a check logs
+/// its own ERROR breadcrumb before returning `Err`, so a verdict discarded
+/// afterwards has already been printed, for a host whose `last*` snapshot
+/// belongs to some earlier phase.
 fn run_checks(
     targets: &HostsGroup,
     registry: &WorkflowRegistry,
     role: Role,
     diagnostics: &mut Vec<Diagnostic>,
 ) -> Vec<UpdateError> {
+    run_checks_where(targets, registry, role, diagnostics, |_| true)
+}
+
+/// [`run_checks`] restricted to the hosts `allowed` accepts.
+///
+/// The predicate is applied *before* the check runs, not to its verdict. That
+/// is the whole point: a check calls
+/// [`log_failed`](crate::update_workflow::checks) on the way to its `Err`, so a
+/// post-filter still emits an operator-facing ERROR for a host whose verdict is
+/// then thrown away — once per package under `prepare --installed-only`, each
+/// time against a snapshot from a fan-out that host was not part of.
+fn run_checks_where(
+    targets: &HostsGroup,
+    registry: &WorkflowRegistry,
+    role: Role,
+    diagnostics: &mut Vec<Diagnostic>,
+    allowed: impl Fn(&str) -> bool,
+) -> Vec<UpdateError> {
     let mut failures = Vec::new();
     for target in targets.targets() {
+        if !allowed(target.hostname()) {
+            continue;
+        }
         let Some((release, transactional)) = host_key(target) else {
             continue;
         };
@@ -924,7 +952,26 @@ async fn prepare_body(
 
     // Surface any per-host command failure from the install fan-out; the
     // prepare check's own failures were collected per fan-out above.
-    let mut failures = host_command_failures(targets, "prepare command failed");
+    //
+    // A host the check already named is dropped here: the two rules overlap on
+    // one signal. `Target::run` records exit `-1` for a timeout, a dropped
+    // connection or an unconnected host, which trips this scan's
+    // `lastexit() != 0` *and* the `("slmicro", true)` / `("YUM", false)`
+    // checks' never-ran gate; a lock message on stderr trips the stderr half
+    // and the marker check. Both would name one host twice, which pushes
+    // `aggregate_failures` out of its single-failure verbatim branch into the
+    // summary — where `host` becomes `None` and the MCP client loses the one
+    // field that says *which* refhost to go and look at. The check's verdict
+    // is the one kept because it is the more specific of the two ("timed out
+    // or failed to run" and "update stack locked" both say more than "prepare
+    // command failed"); dropping the check's instead would trade attribution
+    // back for a coarser diagnosis. `downgrade_body` resolves the identical
+    // overlap the identical way — its `failed_downgrade.insert` gates the
+    // exit-code entry behind the check's verdict for the same host.
+    let mut failures: Vec<UpdateError> = host_command_failures(targets, "prepare command failed")
+        .into_iter()
+        .filter(|e| !e.host.as_ref().is_some_and(|h| check_failed.contains(h)))
+        .collect();
 
     // A host whose prepare failed must not reboot into the failed transaction,
     // mirroring the install/uninstall template's per-host gate: activating the
@@ -1071,6 +1118,20 @@ fn note_dispatch(
 /// a package it was skipped for), and judging that as a prepare would invent a
 /// verdict. First-failure-wins per host keeps `aggregate_failures` in its
 /// single-failure verbatim branch, where `host` survives.
+///
+/// The scope is imposed through [`run_checks_where`], so an out-of-fan-out host
+/// is never *judged*, rather than judged and then filtered: a check logs its
+/// own ERROR breadcrumb before it returns `Err`, and a discarded verdict has
+/// still been printed to the operator — once per package under
+/// `--installed-only`.
+///
+/// A consequence worth naming because it is intended, not incidental: a host
+/// [`build_prepare_map`] dropped (unresolved release key, or a template with no
+/// `--installed-only` variant) is in no `cmd` map, so it is never check-judged
+/// on its stale snapshot. It is not thereby excused — `prepare_body`'s dispatch
+/// accounting fails it by name with "no prepare command could be built for this
+/// host" (#396), which is a truthful verdict where a check's would have been an
+/// invented one.
 fn note_check(
     targets: &HostsGroup,
     registry: &WorkflowRegistry,
@@ -1078,11 +1139,11 @@ fn note_check(
     check_failed: &mut BTreeSet<String>,
     failures: &mut Vec<UpdateError>,
 ) {
-    for e in run_checks(targets, registry, Role::Prepare, &mut Vec::new()) {
+    let judged = run_checks_where(targets, registry, Role::Prepare, &mut Vec::new(), |host| {
+        cmd.contains_key(host)
+    });
+    for e in judged {
         let Some(host) = e.host.clone() else { continue };
-        if !cmd.contains_key(&host) {
-            continue;
-        }
         if check_failed.insert(host) {
             error!(error = %e, "prepare check failed");
             failures.push(e);
@@ -1319,14 +1380,20 @@ async fn downgrade_body(
             targets.run(Command::PerHost(cmd.clone())).await;
             // Check only the hosts that actually ran this command: a host
             // outside `cmd` (e.g. a dead-probe host) still carries its previous
-            // record, whose stale -1 would trip the new timeout gate and cancel
-            // the healthy hosts' rollback.
-            for e in run_checks(targets, registry, Role::Downgrade, &mut Vec::new()) {
-                let host = e.host.as_deref().unwrap_or("");
-                if cmd.contains_key(host) && !transactional_hosts.contains(host) {
-                    error!(error = %e, "downgrade check failed");
-                    failures.push(e);
-                }
+            // record, whose stale -1 would trip the timeout gate and cancel
+            // the healthy hosts' rollback. Scoped through `run_checks_where`,
+            // so such a host is not judged at all — a post-filter would still
+            // have emitted its ERROR breadcrumb, once per package.
+            let judged = run_checks_where(
+                targets,
+                registry,
+                Role::Downgrade,
+                &mut Vec::new(),
+                |host| cmd.contains_key(host) && !transactional_hosts.contains(host),
+            );
+            for e in judged {
+                error!(error = %e, "downgrade check failed");
+                failures.push(e);
             }
         }
     }
@@ -1363,15 +1430,21 @@ async fn downgrade_body(
     let mut failed_downgrade: BTreeSet<String> = BTreeSet::new();
     if !combined.is_empty() {
         targets.run(Command::PerHost(combined.clone())).await;
-        // Same scoping as the per-package loop: only check the transactional
-        // hosts that ran this combined command.
-        for e in run_checks(targets, registry, Role::Downgrade, &mut Vec::new()) {
-            let host = e.host.as_deref().unwrap_or("");
-            if combined.contains_key(host) && transactional_hosts.contains(host) {
-                error!(error = %e, "downgrade check failed");
-                failed_downgrade.insert(host.to_owned());
-                failures.push(e);
-            }
+        // Same scoping as the per-package loop, and imposed the same way: a
+        // host outside `combined` is not judged, rather than judged and then
+        // dropped after its breadcrumb has already been logged.
+        let judged = run_checks_where(
+            targets,
+            registry,
+            Role::Downgrade,
+            &mut Vec::new(),
+            |host| combined.contains_key(host) && transactional_hosts.contains(host),
+        );
+        for e in judged {
+            let Some(host) = e.host.clone() else { continue };
+            error!(error = %e, "downgrade check failed");
+            failed_downgrade.insert(host);
+            failures.push(e);
         }
         // The transactional check only gates "timed out or failed to run"; the
         // downgrade template is a single command, so a non-zero exit is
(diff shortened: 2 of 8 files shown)


mimi1vx (changes requested):
Filling in the four missing `(slmicro, true)` / `("YUM", false)` check arms is
clearly right, and the reasoning for *how narrow* each verdict should be is the
strongest part of this — particularly refusing to transfer zypper's `100`–`107`
band to yum/dnf, and refusing to read the exit code on `prepare --installed-only`
because `rpm -q pkg && yum install pkg` exits `1` on the routine skip. Those are
the two places a less careful fix would have broken working refhosts.
`the_new_arms_resolve_to_their_own_check_not_a_sibling` is a good catch on a real
gap in `is_some()` assertions.

One blocking issue, one that should be fixed alongside it.

### 1. The `-1` gates reintroduce the double-naming the check was designed to avoid

`checks/prepare.rs::transactional_update` argues the point explicitly:

> **No bare non-zero-exit arm, on purpose.** `prepare_body` already owns exit
> codes […] A second verdict on the same signal would name the host twice,
> pushing `aggregate_failures` out of its single-failure verbatim branch into the
> summary — where `host` becomes `None` […] Adding the arm is not "completing"
> the check; it duplicates a verdict the flow already renders.

That reasoning is correct, but `-1` is a *subset* of non-zero, and
`host_command_failures` fires on it:

```rust
let bad_exit = target.lastexit().is_some_and(|c| c != 0);
```

`Target::run` records `Some(-1)` for a timeout, a dropped connection, or an
unconnected host. So on a slmicro or YUM host whose prepare never ran, the flow
now collects **two** failures for one host:

- `host_command_failures` → `"prepare command failed"`
- the new check → `"prepare command timed out or failed to run"`

which lands in the summary branch and collapses `err.host` from `Some("h1")` to
`None`. Before this PR those keys had no prepare check at all, so a timed-out
prepare returned a single verbatim error carrying its host. That is a regression
in the error shape MCP clients read, on exactly the path this PR adds.

`hosts.dedup()` fixes the *roll-call text* ("prepare failed on h1, h1") but not
the lost attribution — `failures.len()` is still 2, so line 439's
`else if failures.len() == 1` never fires.

Same applies to `checks/install.rs::transactional_update`'s `-1` gate, though
`install`/`uninstall` route through the `PlanProvider` adapter rather than
`host_command_failures`, so please double-check whether that path doubles too.

Two ways out, either fine:
- have `prepare_body` skip its own `host_command_failures` entry for a host the
  check already named (`check_failed` is right there), or
- narrow `host_command_failures` to exclude `-1` for roles whose check owns it.

Either way it wants a test asserting `err.host == Some("h1")` for a `-1` prepare
on both new keys — the current suite pins the reason strings but not the host,
so this regression is invisible to it.

### 2. `note_check` logs verdicts it then discards

`run_checks` iterates **every** target in the group, and the check functions call
`log_failed(args)` *before* returning `Err`. `note_check` filters on
`cmd.contains_key(&host)` only afterwards, so a host outside the current fan-out
still emits an ERROR breadcrumb for a verdict that is thrown away — and under
`--installed-only` that repeats once per package, against a stale `last*`
snapshot from `set_repo` or an earlier package.

The `check_failed.insert` dedup guards the outer
`error!(… "prepare check failed")` but not the check's own internal logging.

Filtering the target set *before* calling would fix both the noise and the
wasted work. `run_checks` currently takes the whole group; a variant taking an
allowed-host predicate would keep `note_check` honest by construction rather
than by post-filter.

---

### Non-blocking notes

**`aggregate_failures` dedup.** `hosts.sort(); hosts.dedup();` is correct and the
test covering "two distinct hosts stay two" is the right guard. No issue — just
flagging that it changes the summary text for any pre-existing multi-failure
path, not only the new ones.

**`run_checks` scope change.** Moving from one post-loop call to per-fan-out
calls scoped to `cmd`'s keys means a host dropped by `build_prepare_map` is no
longer check-judged on its stale snapshot. That is an improvement, and the
"no prepare command could be built" failure already covers it — worth a sentence
in the `note_check` doc so it reads as intended rather than incidental.

**Reason-string contract.** The `update_workflow/mod.rs` doc says "no code
branches on them, so a check is free to pick the most accurate one" — that still
holds here, and routing the new arms through the existing `UpdateFailure`
variants keeps the actual contract intact. Good.

**Merge order.** Conflicts textually with #457, which inserts an
`add_op_history` call into the same `prepare_body` hunk this PR rewrites. Also
brushes #454 in `update_flow.rs`. Worth agreeing an order.

plusky:
Both review items are addressed, in `0f8519a1` (the fix) and `721a0c48` (one stale comment the fix left behind). Neither rewrites `87f98303` or anything below it — the branch is still fast-forward from what you reviewed.

## 1. The `-1` gates double-named a host (blocking)

Confirmed, and fixed the way you suggested — through `check_failed`, not by narrowing `host_command_failures` by role.

`Target::run` records exit `-1` for a timeout, a dropped connection or a target that was never connected. That trips `host_command_failures`' `lastexit() != 0` *and* the new `("slmicro", true)` / `("YUM", false)` never-ran gate, so one host contributed two failures, `aggregate_failures` left its single-failure verbatim branch, and `err.host` collapsed from `Some("h1")` to `None` — on exactly the path this PR adds. The exit-`0` lock message overlapped the same way.

`prepare_body` now drops its own coarser entry for a host the check has already named (`crates/mtui-testreport/src/reports/update_flow.rs:975-978`):

```rust
let mut failures: Vec<UpdateError> = host_command_failures(targets, "prepare command failed")
    .into_iter()
    .filter(|e| !e.host.as_ref().is_some_and(|h| check_failed.contains(h)))
    .collect();
```

The check's verdict is the one kept, deliberately: `"prepare command timed out or failed to run"` and `"update stack locked"` both say strictly more than `"prepare command failed"`, so keeping the check buys attribution *and* diagnosis. Dropping the check's instead would have traded diagnosis back for attribution.

New test `perform_prepare_names_the_host_of_a_prepare_that_never_ran`, looped over both new keys (`SL-Micro`/`slmicro` and `rhel`/`YUM`), asserting `err.host == Some("h1")` and the exact reason. Observed red with the filter removed:

```
assertion `left == right` failed: SL-Micro: prepare failed on h1 (h1: prepare command failed; h1: prepare command timed out or failed to run)
  left: None
 right: Some("h1")
```

`perform_prepare_skips_the_reboot_of_an_exit_zero_lock_message_prepare` asserted through `err.to_string().contains(...)` precisely *because* the host used to be named twice. It now asserts `err.reason == "update stack locked"` and `err.host == Some("h1")` exactly, and its comment no longer documents the summary branch as intended. Also red without the filter:

```
  left: "prepare failed on h1 (h1: prepare command failed; h1: update stack locked)"
 right: "update stack locked"
```

`perform_prepare_skips_the_reboot_of_a_host_whose_prepare_failed` (bare non-zero exit) is unaffected: the check passes on that transcript, so the coarse entry survives alone and the error is still verbatim.

### Your question: does the install/uninstall path double too?

**No, and the reason is structural** — it collects failures from a different place, so no code change was needed there.

- `perform_operation` builds its failure list from **only** `report.check_failures` + `report.reboot_failures` (`update_flow.rs:648-654`). It never calls `host_command_failures`; no exit-code scan runs on that path at all.
- The template pushes exactly **one** `check_failures` entry per host plan — a single `push` inside the per-plan loop (`crates/mtui-hosts/src/target/operation.rs:406-420`).
- A host whose check failed is filtered **out of the reboot map before `group.reboot` runs** (`operation.rs:424-436`), so `reboot_failures` cannot contribute a second entry for the same host either.
- A `-1` does reach that check unchanged: `exitcode: target.lastexit().map_or(0, i32::from)` (`crates/mtui-hosts/src/target/hostgroup.rs:1700`).

Pinned rather than asserted: `perform_install_names_the_host_of_an_install_that_never_ran` drives `perform_install` on both new keys with a `-1` transcript and requires `err.host == Some("h1")` plus the exact reason `"command timed out or failed to run"` (role-neutral, since that table also serves `uninstall`). Observed red by injecting the doubling — adding `failures.extend(host_command_failures(targets, "install command failed"))` to `perform_operation`:

```
assertion `left == right` failed: SL-Micro: install failed on h1 (h1: command timed out or failed to run; h1: install command failed)
  left: None
 right: Some("h1")
```

## 2. `note_check` logged verdicts it then discarded

Fixed as suggested, by imposing the scope *before* the check runs rather than filtering its output. New `run_checks_where` takes an `allowed: impl Fn(&str) -> bool` and skips a host before `check(...)` is called (`update_flow.rs:428-431`). `run_checks` keeps its old signature and delegates with `|_| true`, so the `update` flow's caller is untouched and no other crate sees a signature change.

One correction to the report, worth stating rather than claiming a bug I did not fix: **`note_check`'s own site is not reachable today.** It scopes to `cmd`'s keys, `cmd` comes from `build_prepare_map`, and the prepare *check* table has exactly the same key set as the *preparer* table — `("11"|"12"|"15"|"16", false)`, `("YUM", false)`, `("slmicro", true)` (`checks/prepare.rs:133-135` vs `actions/prepare.rs:84-88`) — while all three prepare template families do carry an `installed_only` variant (`actions/prepare.rs:29, 50, 64`). So no host is currently judged-then-discarded there. The migration keeps it that way by construction.

The **identical shape in `downgrade_body` is reachable**, and that is where the symptom actually occurs: a dead-probe host is excluded from `cmd`/`combined` but keeps its probe's `-1` snapshot, and every downgrade check gates `-1` with `log_failed` (`checks/downgrade.rs:16-27` zypper, `:76-84` transactional, plus the YUM one this PR adds). The old post-filter printed an operator-facing `command failed` ERROR for such a host and then threw the verdict away — once per package. Both sites (`update_flow.rs:1378`, `:1427`) now use `run_checks_where`.

That is one hunk beyond what you asked for, deliberately: it is the same shape, it is the reachable instance of it, and this PR's new YUM downgrade check newly reaches those lines for RHEL hosts. Behaviour is otherwise unchanged — the predicate is the same condition as the old post-filter, and `run_checks` always sets `e.host`, so the old `unwrap_or("")` branch was equivalent.

New test `run_checks_where_never_judges_a_host_outside_the_predicate` gives both hosts the **same** failing transcript, so the predicate is the only thing that can separate them, and asserts through captured logs (the in-scope breadcrumb must fire, the out-of-scope one must not). Observed red with the gate degraded back to a post-filter:

```
an out-of-scope host must not be judged, so it must not be logged: command failed host="h1" ... 
command failed host="h2" stderr="System management is locked"
```

## Stale rationales retired

Your framing — the `-1` gate "reintroduces the double-naming the check was designed to avoid" — was argued from comments that are no longer true once the gate exists. Rather than leave them to contradict the fix:

- `checks/prepare.rs`'s **"No bare non-zero-exit arm, on purpose"** paragraph argued from double-naming, which `prepare_body` now prevents. Re-argued from what is still true: the prepare templates are a single command, so `"prepare command failed"` is *accurate*; this key's `transactional-update` returns only `0` or `1`, so there is no status to classify; and an arm would only reword a verdict while changing a pinned reason string. **The arm is still absent.** Same for the `slmicro_prepare_does_not_judge_a_bare_non_zero_exit` test comment.
- The `-1` bullet now says positively why it earns a verdict despite the overlap: `"command failed"` asserts the command ran and failed, `-1` asserts it did neither, and the operator's next step differs.
- `aggregate_failures`' `hosts.dedup()` **stays** (you called it fine), but its citation was the prepare stderr+check pair — the pair this fix makes unreachable. `721a0c48` re-argues it from the overlap that remains reachable: `downgrade_body` seeds its list from the issue-repo removal scan and later adds the same host's downgrade-check verdict, two genuinely distinct causes for one host. It also states the boundary this review turned on — one *signal* reported by two rules has to be resolved before `aggregate_failures` sees it, because arriving there at all means the verbatim branch was skipped and `host` is already lost. `aggregate_failures_names_each_host_once` was rewritten to the downgrade example, and is red without the `dedup()`:
  ```
  one host, named once: downgrade failed on h1, h1 (h1: failed to remove issue repo; h1: downgrade command timed out or failed to run)
  ```

`CHANGELOG.md`: edited this PR's own not-yet-merged `[Unreleased]` entry rather than appending a paragraph that contradicts it. Its last paragraph claimed the dedup existed for the prepare stderr+check overlap and said nothing about attribution; it now states that the failing host's name is preserved, with the roll-call dedup re-illustrated on the downgrade example.

## Trap audit on the new tests

The `-1` fixtures use `MockConnection::with_default`, which answers every command identically — so a test could pass with nothing having run. Guarded:

- `perform_prepare_names_the_host_of_a_prepare_that_never_ran` asserts the prepare command actually dispatched (`handle.commands()` contains `pkg-a`) before checking the verdict; a host no command was built for fails by a different name.
- Every reason assert is `assert_eq!`, not `contains` — a `contains("timed out")` would also match the summary that carries both entries, which is exactly the wrong state being pinned against.
- `NoopRepo::set_repo` runs no command, so the repo-stage `host_command_failures` early return cannot pre-empt the assertion.
- `run_checks_where_never_judges_a_host_outside_the_predicate` asserts the in-scope breadcrumb *fires* as well as the out-of-scope one not firing, so a mutant that logs nothing at all does not pass.

## Gate

Full DoD gate, observed green on `721a0c48`:

| gate | result |
| --- | --- |
| `cargo fmt --all --check` | clean |
| `cargo clippy --workspace --all-targets -- -D warnings` | clean |
| `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features --document-private-items` | clean |
| `cargo test --workspace` | 2507 passed, 0 failed, 0 ignored (37 suites) |
| `cargo test -p mtui-mcp -F mcp` | 286 passed, 0 failed (7 suites) |
| `cargo build --workspace --no-default-features` | ok |
| `cargo build --workspace --all-features` | ok |
| `typos .` | clean |

`mtui-testreport`'s lib suite goes 317 → 320: three new tests, none removed.

plusky:
Merge order (this review asked): proposed **#454 → #458 → #457** — full reasoning in https://github.com/openSUSE/mtui/pull/457#issuecomment-5306278290 (the comment after it). I'll rebase the survivors after each merge.

mimi1vx (dismissed):
LGTM. The headline defect is real and nasty: a `transactional-update` that
printed `System management is locked` and still exited `0` seeded no verdict at
all, so the per-host reboot gate rebooted straight into the locked transaction
while the flow reported success. Moving the prepare verdict inside the fan-out
loop (`note_check`, the companion to `note_dispatch`) is what makes the gate
complete on the `--installed-only` path, and scoping it to the fan-out's own
hosts — rather than filtering the verdict afterwards — is right, because a check
logs its ERROR breadcrumb on the way to `Err`, so a discarded verdict has already
been printed.

Two things I checked specifically:

- Reusing `checks::update::classified` for the slmicro install/uninstall key is
  not new risk: the slmicro *update* check already routes through it, so both
  `("slmicro", true)` keys now share one exit-code vocabulary instead of
  drifting. `pub(super)` on `markers`/`classified` is the minimal way to get it.
- `hosts.dedup()` is added after `hosts.sort()`, which is what makes it correct.

Non-blocking:

1. Dropping the coarse `host_command_failures` entry for a host the check has
   already named is the right precedence, but it means a host with a genuine
   non-zero exit on package 1 *and* a marker failure on package 2 now reports
   only the marker reason. Better diagnosis, but a distinct signal is silently
   lost — worth a line in Known limitations.
2. The YUM `--installed-only` hole is user-visible and stays open: the check
   correctly abstains, but the flow-level exit-code scan still fails a RHEL host
   that legitimately skips a package. Deferring the fix to the template is
   defensible, but #406 then closes while a user-facing failure the same analysis
   uncovered remains. Please file the follow-up and reference it from the
   CHANGELOG, not only from the PR body.
3. `"Unknown Error"` on the YUM install/uninstall key needs its own issue — it is
   exactly the string an operator will paste into a bug report.
4. The slmicro *downgrade* check is still marker-blind, on the rollback path.
   That is the same hole in the same family, in the function #454 is already
   rewriting; consider folding it in there rather than opening a third issue.

Reason strings change for MCP clients — the CHANGELOG covers it and nothing
matches on them programmatically; just confirm no downstream config
prompt-matches on `"install command failed"`.

Merge order: this rewrites the `prepare_body` per-package loop and
`downgrade_body`, both of which #454 and #457 also touch. Suggest landing it
fourth, rebased on #454; #457 goes last.

plusky:
Rebased onto main after #459's merge — the conflict was CHANGELOG-only (both entries kept, order preserved). The reviewed delta is unchanged: `git diff` of old-base..old-tip vs main..`d10d0b0e` differs only in the CHANGELOG hunk's line offset, zero code differences. The push dismissed the approval; re-requesting.

mimi1vx (approved):
Re-approving: the force-push was a pure rebase onto `main` after #459 landed —
I diffed the old and new heads and the only delta is the CHANGELOG hunk offset,
with no change to any source file. All checks green on the new head.

My earlier review stands unchanged, including the outstanding items in it.

PR #459, opened 2026-08-15
fix(core): let a body-level cancel be the fan-out's verdict

### LINKED ISSUE #404: a cancelled command body is reported as FanOut, not Cancelled
A `job_cancel` that lands inside a command body is reported as `CommandError::FanOut`, not `CommandError::Cancelled` — so the guard that exists to stop a cancel from burying a real failure ends up letting a cancel impersonate one.

AGENTS.md: "A flow that stops on a cancel must surface as `CommandError::Cancelled`, not a generic failure."

## Mechanism

`crates/mtui-core/src/command.rs:193-212` — the fan-out loop pushes **any** `Err` into `failures`, with no variant discrimination:

```rust
let mut cancelled = false;
for rrid in &resolved {
    if session.cancel_requested() {
        cancelled = true;
        break;
    }
    ...
    if let Err(exc) = self.call(session, args).await {
        tracing::error!(command = self.name(), rrid = %rrid, error = %exc, "command failed");
        failures.push((rrid.clone(), exc));
    }
}
```

`cancelled` is set **only** by the loop-top `session.cancel_requested()` check. It is never derived from the returned error, so a body-originated `CommandError::Cancelled` is indistinguishable from a real failure at this seam — and is logged at `error!` "command failed" on the way in.

Then `command.rs:215-227`:

```rust
if cancelled && failures.is_empty() {
    // Report the stop — but only when nothing actually failed. A real
    // per-template failure outranks it and falls through to the
    // `FanOut` aggregate below: burying a broken template behind a
    // bare "cancelled" is the one thing the caller must not be told.
```

with `command.rs:245-246`: `if !failures.is_empty() { return Err(CommandError::FanOut(failures)); }`

The comment's intent is right. The bug is that the thing sitting in `failures` may itself *be* the cancel, so the guard suppresses the verdict it should be producing.

## Reachable

`crates/mtui-core/src/commands/perform.rs:73-78` produces exactly that variant from a flow:

```rust
fn map_flow_error(e: &UpdateError) -> CommandError {
    if e.is_cancelled() {
        return CommandError::Cancelled(e.to_string());
    }
```

Two shapes, both wrong:

- **One template.** `cancelled` stays `false`; the result is `FanOut([(rrid, Cancelled(..))])`, rendered by `error.rs:41` as `"fan-out failed on <rrid> (<rrid>: cancelled: …)"`.
- **Several templates.** Template 1 succeeds; the operator fires `job_cancel` while template 2 is mid-flight; template 2's body returns `Cancelled` into `failures`; template 3's loop-top check sets `cancelled = true`. `failures` is non-empty, so `FanOut` wins.

Newly reachable for `install`/`uninstall` after the cancellation checkpoints were added to those flows.

## Fix

Discriminate in the loop, which preserves both halves of the rule:

```rust
Err(CommandError::Cancelled(_)) => { cancelled = true; break; }
Err(exc) => { tracing::error!(...); failures.push((rrid.clone(), exc)); }
```

## Test gap worth naming

Existing coverage (`crates/mtui-core/src/commands/mod.rs:1113-1160`) only exercises bodies that return `Ok(())` after cancelling the token — i.e. the path where `cancelled` comes from the loop-top check. **Nothing pins the body-returns-`Cancelled` case**, which is why this survived. A test needs 2+ templates with a body that actually returns `Cancelled`, asserting the verdict is `Cancelled` and not `FanOut`; and it should be mutation-checked, since a test asserting only "an error came back" passes either way.

plusky:
Closes #404.

## The bug

The multi-template fan-out driver pushed **every** per-template `Err` into `failures` without looking at the variant, and set its `cancelled` flag **only** from the loop-top token check — never from the returned error.

A command body that stops at one of its own cancellation checkpoints returns `CommandError::Cancelled` (`commands/perform.rs::map_flow_error`, reachable from `install`, `uninstall`, `prepare`, `downgrade`, `update`). That cancel therefore landed in the failure aggregate, logged at `error!` "command failed" on the way in.

The `cancelled && failures.is_empty()` guard then saw a non-empty list and fell through to `FanOut`. That guard exists so a cancel cannot bury a real failure. The bug let a cancel *impersonate* one: the operator got `fan-out failed on <rrid> (<rrid>: cancelled: …)` for a job they had just cancelled themselves.

## The fix

Discriminate the variant in the loop. `Err(CommandError::Cancelled(detail))` breaks at the template boundary — exactly what the loop-top check already does — records the flow's own detail, and never enters `failures`. Everything else is collected as before.

Both halves of the AGENTS.md rule survive:

- The verdict comes from the **returned error variant**, never from sniffing the session token. Sniffing would mask a real host failure that merely coincided with a cancel.
- A real failure banked **before** the cancel still outranks it and is still reported as the `FanOut` aggregate — now without the cancel entry padding the failure list.

The `Cancelled` payload carries the interrupted flow's own detail (`stopped after 1 of 3 templates; SUSE:Maintenance:2:2: flow stopped mid-body`), which is what that variant promises: report what completed before stopping.

## Judgement calls

**Counted, not derived — and this is not optional scope.** The completion figure was `resolved.len() - skipped.len() - failures.len()`. That was already wrong on any `break` path (a fan-out that stopped at the first of two reported "stopped after 2 of 2 templates"), and the fix above makes it *worse*: the body-cancelled template no longer lands in `failures`, so the derivation would newly count it as done. The same derivation also fed the `succeeded` summary log — so on the outranked-cancel path the log named the template that was interrupted mid-flight, plus every template the break never reached, as templates the update had been applied to. `completed` is now a `Vec<&str>` pushed only where the body returned `Ok`, and the count, the log, and the all-skipped `NoRefhostsDefined` guard all read that one list. The old `HashSet` + filter derivation is deleted rather than left alongside.

**The issue's one-template shape does not exist, and I did not "fix" it.** #404 lists a single-template case producing `FanOut([(rrid, Cancelled(..))])`. `Command::run` early-returns for `resolved.len() <= 1` and propagates the body's error verbatim, so that path already surfaced `Cancelled` correctly. Only the multi-template shape was broken. Saying so is cheaper than adding a redundant guard that would need its own justification later.

**No new checkpoint, no new signal handler.** The change is entirely in how an *already returned* error is classified. The loop-top `session.cancel_requested()` check is the sanctioned template-boundary checkpoint and is untouched; nothing was added to a parallel fan-out, and nothing polls the token past a point of no return.

**A dev-dependency for one log assertion.** Pinning the `succeeded` line needed `tracing-subscriber` in `mtui-core`'s dev-dependencies and a small capture helper. The helper installs the subscriber **globally and permanently**, with a thread-local sink for scoping, rather than using a thread-local default: `tracing` caches callsite interest process-wide, and the `succeeded` callsite is reached by many other tests in this binary that have no subscriber, so a thread-local default would have been cached as `Interest::never` and made the assertion pass or fail by test order. `set_global_default` rebuilds that cache. Verified by running the test alone, under the full parallel suite five times, and under `--test-threads=1` (which forces the callsite to be reached by other tests first).

## Testing

Three new driver tests in `commands::cancellation_seam`, all mutation-verified.

`fanout_body_cancel_reports_cancelled_not_fanout` — three templates, the second body stopping at its own checkpoint while the session token is **deliberately never fired**. It asserts the full composed message by exact equality, not `contains("cancelled")`: the broken rendering nests `…: cancelled: …` *inside* `fan-out failed on …`, so a substring assertion would have passed on the unfixed code. It also asserts the full `ran` vector from a genuinely shared `Arc<Mutex<…>>` the probe writes from inside `call`. Because the token is never fired, a token-sniffing implementation degrades to `FanOut` and fails here.

`real_failure_before_body_cancel_still_outranks_it` — a real failure on template A, a body cancel on B, and the token **is** fired, so a token-sniffing implementation would bury A's error. Asserts `FanOut`, exactly one entry, that the entry is A's `Other`, and that the fan-out still stopped at the boundary.

`succeeded_log_names_only_completed_templates` — four templates: A completes, B fails for real, C stops mid-body, D is never reached. Captures the `succeeded` field and asserts exact vector equality against `[A]`. Exact equality because the defect is *extra* names in that list, which `contains` cannot see.

**How RED was observed.** Nine mutants across two passes, each applied by unique-anchor replacement that aborts unless the anchor matches exactly once, and each run checked to have actually *run* a non-zero test count rather than trusting an exit code:

| mutation | result |
| --- | --- |
| `Err(Cancelled)` arm reverted to the old catch-all push | RED ×2 — `fan-out failed on SUSE:Maintenance:2:2 (…: cancelled: flow stopped mid-body)`, and `left: 2, right: 1` on "the cancel must not pad the aggregate" |
| `break` removed from that arm | RED ×2 — `stopped after 2 of 3 templates`, and the log gains `SUSE:Maintenance:4:4` |
| completed templates no longer recorded | RED — `stopped after 0 of 3 templates` |
| count derived from the lists again | RED — `stopped after 3 of 3 templates` |
| detail suffix dropped | RED — suffix missing from the message |
| outrank guard weakened to plain `if cancelled` | RED — `expected FanOut, got Cancelled(…)` |
| `succeeded` log derived from the lists again | RED — `left: ["…1:1, …3:3, …4:4"]` vs `right: ["…1:1"]` |
| capture layer records nothing (anti-vacuity) | RED — `left: []`; an empty capture cannot read as "nothing extra was named" |

Worth recording: on the first pass the `succeeded`-log mutant came back **green**. That is how the log defect was established as genuinely unpinned rather than argued about — the test was written afterwards, and the same mutant is red now.

- The `!detail.is_empty()` branch is defensive only. `map_flow_error` always produces a non-empty detail, so the bare `stopped after N of M templates` form is unreachable from today's flows and is not covered by a test.
- On the **pre-existing** loop-top-break path (token fired between templates, no body involvement) only the verdict variant is pinned, by the untouched `fanout_driver_stops_at_template_boundary_on_cancel`. The completion count on that path is pinned only indirectly, through the shared counter.
- No MCP client can distinguish a cooperative body stop from a driver-boundary stop: both arrive as one `cancelled: …` string. `mtui-mcp` records the richer payload through its existing cancel-claimed-record path, but the wire shape is unchanged and this PR does not touch any tool schema.
- The capture subscriber is now the one global subscriber in `mtui-core`'s lib test binary and stays installed for the process. It is inert (thread-local sink, empty by default), but a future test wanting tracing output must reuse this helper rather than install its own.
- A cancel arriving mid-`call` in a body that does **not** opt into the seam is still handled only by the MCP job layer's hard abort. Unchanged, and out of scope here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -346,10 +346,14 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   it had applied — instead of `fan-out failed on ...` with the cancel listed as
   a template failure (#404). A real template failure collected before the
   cancel still outranks it and is still reported as the fan-out aggregate, now
-  without the cancel entry padding the failure list; and the `stopped after N
-  of M templates` count now reports templates actually completed instead of
-  counting never-attempted ones as done — as does the `succeeded` log line,
-  which no longer names templates the stop never reached.
+  without the cancel entry padding the failure list — but the stop is no longer
+  dropped from that aggregate either: it is appended to the message
+  (`fan-out failed on A (A: boom); stopped after 0 of 3 templates; B: <what the
+  interrupted flow had done>`), so a caller is no longer left believing the
+  templates the stop never reached ran clean. The `stopped after N of M
+  templates` count now reports templates actually completed instead of counting
+  never-attempted ones as done — as does the `succeeded` log line, which no
+  longer names templates the stop never reached.
 
 ### Security
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -340,6 +340,16 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   non-determinism: `request_review --watch` used to arm the process-wide SIGINT
   handler as a side effect, after which Ctrl-C was silently swallowed for the
   rest of the session.
+- A cancel (MCP `job_cancel`, REPL Ctrl-C) that lands inside a command body
+  during a multi-template fan-out is again reported as `cancelled: stopped
+  after N of M templates` — carrying the interrupted flow's own detail of what
+  it had applied — instead of `fan-out failed on ...` with the cancel listed as
+  a template failure (#404). A real template failure collected before the
+  cancel still outranks it and is still reported as the fan-out aggregate, now
+  without the cancel entry padding the failure list; and the `stopped after N
+  of M templates` count now reports templates actually completed instead of
+  counting never-attempted ones as done — as does the `succeeded` log line,
+  which no longer names templates the stop never reached.
 
 ### Security
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2421,6 +2421,7 @@ dependencies = [
  "tokio",
  "tokio-util",
  "tracing",
+ "tracing-subscriber",
  "uuid",
  "wiremock",
 ]
--- a/crates/mtui-core/Cargo.toml
+++ b/crates/mtui-core/Cargo.toml
@@ -51,3 +51,7 @@ serial_test.workspace = true
 tempfile = "3"
 wiremock.workspace = true
 tokio = { workspace = true, features = ["macros", "rt", "rt-multi-thread"] }
+# Captures the fan-out driver's `succeeded` summary field, so the list of
+# templates it names is asserted rather than assumed
+# (`commands::cancellation_seam::succeeded_log`).
+tracing-subscriber.workspace = true
--- a/crates/mtui-core/src/command.rs
+++ b/crates/mtui-core/src/command.rs
@@ -155,7 +155,10 @@ pub trait Command: Send + Sync {
     /// boundary and keeping the flow's detail — rather than collecting it as a
     /// template failure. A real failure banked before the cancel still
     /// outranks it and is still reported as the [`CommandError::FanOut`]
-    /// aggregate.
+    /// aggregate; the stop then rides along as that aggregate's `stop` note
+    /// (`stopped after N of M templates`, plus the interrupted flow's detail),
+    /// so it neither pads the failure list nor leaves the templates the break
+    /// never reached looking as though they ran clean.
     async fn run(&self, session: &mut Session, args: &ArgMatches) -> CommandResult {
         session.check_cancelled()?;
         let resolved = resolve_templates(self.scope(), session, args)?;
@@ -247,23 +250,33 @@ pub trait Command: Send + Sync {
         }
 
         restore_active(session, restore);
+        // How far the fan-out got, in one shape for both verdicts: it is the
+        // `Cancelled` payload when the stop *is* the verdict, and the `FanOut`
+        // aggregate's `stop` note when a real failure outranks it. The cancel
+        // never becomes a failure entry — it is not a broken template — but a
+        // caller told only "template X failed" would otherwise read the
+        // templates the break never reached as having run clean.
+        let stop_summary = || {
+            let mut msg = format!(
+                "stopped after {} of {} templates",
+                completed.len(),
+                resolved.len()
+            );
+            if let Some((rrid, detail)) = &body_stop
+                && !detail.is_empty()
+            {
+                msg.push_str(&format!("; {rrid}: {detail}"));
+            }
+            msg
+        };
+
         if cancelled && failures.is_empty() {
             // Report the stop — but only when nothing actually failed. A real
             // per-template failure outranks it and falls through to the
             // `FanOut` aggregate below: burying a broken template behind a
             // bare "cancelled" is the one thing the caller must not be told.
             tracing::info!(command = self.name(), "fan-out cancelled");
-            let mut msg = format!(
-                "stopped after {} of {} templates",
-                completed.len(),
-                resolved.len()
-            );
-            if let Some((rrid, detail)) = body_stop
-                && !detail.is_empty()
-            {
-                msg.push_str(&format!("; {rrid}: {detail}"));
-            }
-            return Err(CommandError::Cancelled(msg));
+            return Err(CommandError::Cancelled(stop_summary()));
         }
 
         if !completed.is_empty() {
@@ -273,7 +286,10 @@ pub trait Command: Send + Sync {
             tracing::info!(command = self.name(), skipped = %skipped.join(", "), "no connected host");
         }
         if !failures.is_empty() {
-            return Err(CommandError::FanOut(failures));
+            return Err(CommandError::FanOut {
+                failures,
+                stop: cancelled.then(stop_summary),
+            });
         }
         if !skipped.is_empty() && completed.is_empty() {
             // Every resolved template was skipped: the command executed on
--- a/crates/mtui-core/src/command.rs
+++ b/crates/mtui-core/src/command.rs
@@ -149,7 +149,13 @@ pub trait Command: Send + Sync {
     /// multi-template job stops at the next template boundary instead of
     /// grinding through the rest. A cancel arriving *mid*-`call` is only
     /// observed if the body opts in ([`Session::cancel_requested`]); otherwise
-    /// the MCP job layer hard-aborts after its grace period.
+    /// the MCP job layer hard-aborts after its grace period. A body that does
+    /// opt in reports its stop as [`CommandError::Cancelled`], and the driver
+    /// treats that variant as the cancel itself — breaking at the template
+    /// boundary and keeping the flow's detail — rather than collecting it as a
+    /// template failure. A real failure banked before the cancel still
+    /// outranks it and is still reported as the [`CommandError::FanOut`]
+    /// aggregate.
     async fn run(&self, session: &mut Session, args: &ArgMatches) -> CommandResult {
         session.check_cancelled()?;
         let resolved = resolve_templates(self.scope(), session, args)?;
@@ -193,6 +199,17 @@ pub trait Command: Send + Sync {
         let mut skipped: Vec<String> = Vec::new();
 
         let mut cancelled = false;
+        // Templates whose body actually returned `Ok`, in fan-out order —
+        // counted, never derived. Deriving it (`resolved` minus `skipped` minus
+        // `failures`) claimed every template a `break` never reached, and the
+        // body-cancelled one that no longer lands in `failures`, as done: it
+        // reported "stopped after M of M templates" for a fan-out that stopped
+        // at the first one, and named never-attempted templates as `succeeded`.
+        let mut completed: Vec<&str> = Vec::new();
+        // Set when the break came from a body-level cancel: the flow's own
+        // verdict detail, preserved rather than flattened (`CommandError::
+        // Cancelled` promises the payload names what the flow managed to do).
+        let mut body_stop: Option<(String, String)> = None;
         for rrid in &resolved {
             // Template boundary = cancellation checkpoint: a job_cancel that
             // lands while template N runs stops the fan-out before N+1.
@@ -208,9 +225,24 @@ pub trait Command: Send + Sync {
             }
             session.activate(rrid);
             session.display.template_banner(rrid);
-            if let Err(exc) = self.call(session, args).await {
-                tracing::error!(command = self.name(), rrid = %rrid, error = %exc, "command failed");
-                failures.push((rrid.clone(), exc));
+            match self.call(session, args).await {
+                Ok(()) => completed.push(rrid.as_str()),
+                // A body that stopped at one of its own cancellation
+                // checkpoints (`commands::perform::map_flow_error`) *is* the
+                // cancel verdict, not a template failure: stop at the boundary
+                // exactly like the loop-top check does. Pushing it into
+                // `failures` would let a cancel impersonate a broken template
+                // and be reported as a `FanOut` aggregate.
+                Err(CommandError::Cancelled(detail)) => {
+                    tracing::info!(command = self.name(), rrid = %rrid, "template body cancelled");
+                    body_stop = Some((rrid.clone(), detail));
+                    cancelled = true;
+                    break;
+                }
+                Err(exc) => {
+                    tracing::error!(command = self.name(), rrid = %rrid, error = %exc, "command failed");
+                    failures.push((rrid.clone(), exc));
+                }
             }
         }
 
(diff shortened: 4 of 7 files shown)


mimi1vx (changes requested):
The central fix is right and the reasoning matches the seam's documented rule:
a body-level stop is the cancel verdict, not a template failure, and a real
failure banked before it still outranks it. Replacing the derived `done` count
with a counted `completed` is a strict improvement — the old
`resolved - skipped - failures` genuinely claimed never-attempted templates, and
`succeeded_log_names_only_completed_templates` picks exactly the one shape where
that lied. Good test.

One thing to reconsider before merge.

### The body's detail is dropped when a real failure outranks the cancel

In the fall-through path, `body_stop` is only consumed inside the
`cancelled && failures.is_empty()` branch:

```rust
if let Some((rrid, detail)) = body_stop
    && !detail.is_empty()
{
    msg.push_str(&format!("; {rrid}: {detail}"));
}
```

When a real failure outranks the cancel, `body_stop` is silently discarded and
the returned `FanOut` names only the broken template. The caller is told nothing
about the template that was interrupted mid-body, nor that the remaining
templates never ran. Before this change the cancelled template at least appeared
in the aggregate as a `Cancelled` entry.

`real_failure_before_body_cancel_still_outranks_it` pins
`failures.len() == 1`, so this is deliberate — and I agree the cancel must not
*pad the failure list*. But "not a failure entry" and "not mentioned at all"
are different. The `tracing::info!(… "template body cancelled")` breadcrumb is
the only surviving trace and it does not reach an MCP client.

Suggestion: append the stop summary to the aggregate reason rather than adding a
failure entry — e.g. `… (h1: boom); stopped after 1 of 4 templates`. That keeps
the `FanOut` variant, keeps the failure list at 1, and stops the caller
believing templates C and D ran clean.

### Contract note on `CommandError::Cancelled`

`Err(CommandError::Cancelled(_))` now breaks the fan-out **unconditionally**,
independent of the session token. Today that is equivalent in practice — every
producer (`perform.rs::map_flow_error`, `Session::check_cancelled`) derives from
a genuine session cancel, so the loop-top check would have broken anyway. But it
promotes `Cancelled` into a *fan-out-terminating* variant by contract.

The `Command::run` doc update covers it well. Worth mirroring one line onto
`CommandError::Cancelled` in `error.rs`, or a future per-template use of the
variant will silently abandon the remaining templates.

---

### Non-blocking: the global test subscriber

The interest-caching analysis in `succeeded_log`'s doc is correct and the
thread-local sink is the right scoping mechanism. Two residuals worth a line in
the comment, since this is the second copy of the pattern (`#454` adds the same
shape to `mtui-testreport`):

- An unfiltered `Registry` reports no `max_level_hint`, so `LevelFilter::current()`
  goes to TRACE for the **whole** `mtui-core` lib test binary once `start()` runs.
  Every previously-dead `debug!`/`trace!` in every other test now evaluates its
  `%`/`?` formatting arguments. Cheap in practice, but it is a behaviour change
  for tests that never opted in, and a `Display` impl that takes a lock could
  deadlock. A `LevelFilter`-bounded layer would keep the interest rebuild without
  the blast radius.
- The capture relies on `#[tokio::test]` defaulting to the current-thread flavour
  so the fan-out runs on the arming thread. #454's copy documents this explicitly;
  this one does not. Worth adding, or a later `flavor = "multi_thread"` will
  silently empty the capture and the exact-equality assertion will read as a
  product bug.

Confirmed there is no other `set_global_default` in `mtui-core`, so the
`Once`-guarded install cannot lose the race to a sibling today.

plusky:
All three points are addressed, in two commits on top of `eb79c48c`; nothing at or below the reviewed SHA was rewritten.

## 1. The body's detail is dropped when a real failure outranks the cancel

Implemented in the shape you suggested — the stop summary goes onto the **aggregate reason**, not into the failure list. `CommandError::FanOut` gained a second field, `stop: Option<String>`, rendered as a `; <summary>` suffix (`crates/mtui-core/src/error.rs:38-59`), and the summary the `Cancelled` payload already built is now built once and shared by both verdicts (`crates/mtui-core/src/command.rs:253-292`). Commit `2c04d9d1`.

Your scenario now reads:

```
fan-out failed on A (A: boom); stopped after 0 of 3 templates; B: flow stopped mid-body
```

`failures.len()` stays 1 — `real_failure_before_body_cancel_still_outranks_it` keeps that assertion and gained exact-equality pins on both the rendered string and the `stop` field.

On your point that the `tracing::info!` breadcrumb does not reach an MCP client: the suffix does. `EngineError::Command` is `#[error(transparent)]` over `CommandError` (`crates/mtui-core/src/engine.rs:57-58`), and `McpSession::run_command` puts `other.to_string()` straight into `McpCommandError.stderr` (`crates/mtui-mcp/src/session.rs:1033-1040`), which is what the tool-result envelope renders. It is the same string the REPL prints.

### Shape change you should know about

Keeping the `FanOut` *variant* meant turning it from a tuple into a struct variant: `FanOut { failures, stop }`. There is no way to append text to a `thiserror` `Display` without a field to hold it, and a synthetic `failures` entry is exactly what you ruled out. Blast radius, checked by grep: three match sites (`crates/mtui-core/tests/fanout.rs:185`, `crates/mtui-core/src/commands/mod.rs:1351` and `:1460`) plus the two `Display` tests in `error.rs`. Nothing outside `mtui-core` constructs or matches the variant, and no snapshot or doc carries the string. With `stop: None` the message is byte-identical to before — `fanout_display_has_stable_format` and `fanout_with_single_failure` are unchanged apart from constructor syntax and still pin it. It is your enum; if you would rather it stayed a tuple, say so and I will rework it.

### Scope extension (deliberate)

The drop had a sibling the review did not name: a *token* cancel observed at the loop-top checkpoint with a failure already banked. Before this PR and after it, that aggregate also said nothing about the stop. Because the summary is now built once and shared, both verdicts carry it; the token case gets the bare `stopped after N of M templates`, with no `; rrid: detail` clause, since no flow reported one there. Pinned by a new test, `real_failure_before_token_cancel_keeps_the_stop_note` (`crates/mtui-core/src/commands/mod.rs:1378-1416`).

### Message format

Reused the cancelled-branch format verbatim (`stopped after N of M templates`, plus `; {rrid}: {detail}` only when the detail is non-empty), so the aggregate can carry two semicolon-separated clauses. I considered `(rrid: detail)` parentheses to cut the semicolon noise and rejected it: one shared builder is what guarantees the two verdicts cannot drift, and that only holds while the format is literally the same string.

## 2. Contract note on `CommandError::Cancelled`

Added at `crates/mtui-core/src/error.rs:74-84`, same commit. It says what you asked: this variant coming *out of a command body* terminates the fan-out independent of the session token, so a future per-template use that is not a session-level stop would silently abandon the templates after it. It also names the alternative — such a condition wants `Other`, which is collected and lets the fan-out continue.

## 3. Non-blocking: the global test subscriber

Taken as doc lines only this round; no `LevelFilter`-bounded layer yet. Both residuals are named at `crates/mtui-core/src/commands/mod.rs:1114-1134` (commit `e8edc0b4`):

- the missing `max_level_hint` on an unfiltered `Registry`, so `LevelFilter::current()` is `TRACE` for the whole `mtui-core` lib test binary from the first `start()`, every previously-dead `debug!`/`trace!` evaluating its formatting arguments in every parallel test, and a `Display`/`Debug` impl taking a lock the emitting thread holds deadlocking rather than staying unevaluated — with the `LevelFilter`-bounded layer named as the remedy if a callsite ever pays for it, explicitly in preference to going back to a thread-local default;
- the `#[tokio::test]` current-thread reliance, and that a later `flavor = "multi_thread"` would move the event to a worker thread with no sink, silently emptying the buffer so the exact-equality assertion reads as "the driver named no templates" — a product bug that is not one.

That matches #454's copy of the pattern.

Your `set_global_default` observation checks out rather than being taken on trust: `grep -rn set_global_default crates/mtui-core` returns exactly one hit, ours at `commands/mod.rs:1185`. The other two in the tree are in `mtui-datasources`, a different test binary.

## Test-probe refactor

`ScriptedBody`'s `cancel_token_too: bool` became `fire_token_on: Option<&'static str>`, and the token now fires *before* the scripted verdict so it composes with the failing body as well as the cancelling one (`crates/mtui-core/src/commands/mod.rs:1225-1238`). Two overlapping knobs would have been worse. The existing tests keep identical semantics: `fire_token_on: Some(RRID_B)` is the old `cancel_token_too: true`.

## Observed red

Every new assertion is exact equality, never `contains` — the defect here is a *missing* suffix and its inverse a *spurious* one. Three mutations, each reverted and the suite re-run green afterwards:

1. `stop: cancelled.then(stop_summary)` → `stop: None`, i.e. exactly the behaviour you flagged: 3 passed, 3 failed. `real_failure_before_body_cancel_still_outranks_it` reports

   ```
   left:  "fan-out failed on SUSE:Maintenance:1:1 (SUSE:Maintenance:1:1: boom)"
   right: "fan-out failed on SUSE:Maintenance:1:1 (SUSE:Maintenance:1:1: boom); stopped after 0 of 3 templates; SUSE:Maintenance:2:2: flow stopped mid-body"
   ```

   with `real_failure_before_token_cancel_keeps_the_stop_note` and `succeeded_log_names_only_completed_templates` failing alongside it.
2. `; {stop}` suffix removed from the `Display`: `fanout_stop_note_is_appended_to_the_aggregate` fails at `error.rs:178`, while `fanout_display_has_stable_format` and `fanout_with_single_failure` stay green — that pair is the evidence that a `stop: None` message is unchanged.
3. `stop` attached unconditionally: `partial_failure_raises_fanout_aggregate` fails with `left: Some("stopped after 1 of 2 templates")`, `right: None`. That assertion is new but pins pre-existing behaviour, so it needed its own mutation to prove it can fail at all.

## CHANGELOG

Edited this PR's own not-yet-merged `[Unreleased]` bullet rather than appending a contradicting one: the old text stopped at "without the cancel entry padding the failure list", which is now half the story. The net diff against `main` is still a single added bullet.

## Gate

Local, on `e8edc0b4`: `fmt --check` clean; `clippy --workspace --all-targets -D warnings` clean; `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features --document-private-items` clean; `cargo test --workspace` 2491 passed / 0 failed; `cargo test -p mtui-mcp -F mcp` 286 passed / 0 failed; `cargo build --workspace --no-default-features` and `--all-features` both build; `typos` clean. CI on these two commits is the confirmation to wait for.

mimi1vx (approved):
Reviewed the driver change, the `CommandError` reshape and the new tests. LGTM.

The discrimination in the loop is the right fix, and turning `completed` into a
counted list rather than a derivation (`resolved` minus `skipped` minus
`failures`) fixes a second, latent defect that the first fix would otherwise have
made worse. Carrying the outranked stop as the aggregate's `stop` note instead of
dropping it is the right call: without it a caller reads the templates the break
never reached as clean.

Checked and satisfied:

- The verdict comes from the returned error variant, never from the session
  token, so a real host failure coinciding with a cancel is not masked.
- `cancelled.then(stop_summary)` is only reachable with a non-empty `failures`,
  since the `cancelled && failures.is_empty()` arm returns first.
- The `NoRefhostsDefined` guard reads the same `completed` list, so the
  all-skipped case is unchanged.

Two non-blocking notes:

1. `succeeded_log::start()` installs a permanent global `Registry` with no level
   bound. Your own rustdoc says this pushes `LevelFilter::current()` to `TRACE`
   for the whole `mtui-core` lib test binary, makes every `debug!`/`trace!`
   evaluate its arguments in every parallel test, and creates a deadlock hazard
   for a `Display` impl that takes a lock — and then says "bound the layer with a
   `LevelFilter` if that ever matters". Please just do it now:
   `.with_filter(LevelFilter::INFO)` is one line, keeps the interest-cache
   rebuild that is the whole point, and removes the hazard rather than
   documenting it.
2. `FanOut` going from a tuple to a struct variant is a semver break on a `pub`
   enum. All four in-workspace sites are updated and the crates are internal, so
   this is fine as-is — flagging only so it is a decision rather than an
   accident.

Minor: #454 adds an AGENTS.md list of the canonical copies of this
global-subscriber capture pattern; whichever of the two lands second should add
itself to that list.

Merging this first — it is disjoint from #454/#456/#457/#458.

PR #461, opened 2026-08-16
Four minor cleanups: two docs contradicting the code, a lossy seam, two weak tests

### LINKED ISSUE #408: Four minor cleanups: two docs contradicting the code, a lossy seam, two weak tests
Four small items collected from adversarial review of the reboot/history work. None is a live bug; each is a place where the code and its documentation disagree, or a test that cannot fail. Grouped because they are all two-line changes.

One item originally in this set — a doc comment saying history is recorded "before fan-out" — is **already fixed** (commit `c44765cd` rewrote it to "Called after the command has dispatched … but *before* any transactional reboot"). Verified before filing; not included.

## 1. `last_output`'s doc contradicts what the default actually does

`crates/mtui-hosts/src/target/operation.rs:200-205`:

```rust
/// Returns `None` for a host outside the group (should not happen for a
/// host with a [`HostPlan`]); the template treats that as an empty
/// snapshot rather than panicking.
fn last_output(&self, hostname: &str) -> Option<HostOutput>;
```

The use site, `operation.rs:386`: `let output = group.last_output(&plan.hostname).unwrap_or_default();`

`HostOutput` derives `Default` with `pub exitcode: i32`, so the "empty snapshot" is **exit 0** — which every check reads as success (`checks/install.rs:20-22`, and the no-table fallback `None if a.exitcode != 0` at `update_workflow/mod.rs:354`). Since the reboot map is filtered only by *failed* check hosts (`operation.rs:404-415`), an unknown host silently passes its check and stays in the reboot map.

A `None` there is a bug by construction — the host has a `HostPlan`. So either say that in the doc, or log at WARN, rather than describing a benign fallback that is actually a silent pass.

## 2. The check seam cannot represent a cancellation

`update_workflow/mod.rs:335-343` narrows the check result to `Err(e.reason)` — a bare `String`:

```rust
Err(e) => Err(e.reason),
```

and `reports/update_flow.rs:492-496` rebuilds an error from it:

```rust
let mut failures: Vec<UpdateError> = report
    .check_failures
    .into_iter()
    .map(|(host, reason)| UpdateError::new(reason, host))
    .collect();
```

`UpdateError::new` hardcodes `cancelled: false`, so `UpdateError.cancelled` cannot survive the seam — the seam type is `Result<(), String>` on the `mtui-hosts` side. Latent today, but the codebase leans on `cancelled` being distinguishable from a genuine failure (`checks/downgrade.rs` already special-cases exit `-1` for an SSH no-output timeout), and it is unrepresentable across this boundary rather than merely unused.

## 3. A test that passes for the wrong reason

`reports/update_flow.rs:2576-2585`:

```rust
async fn perform_uninstall_reports_a_transactional_host_that_never_reconnects() {
    ...
    assert_eq!(err.host.as_deref(), Some("h1"));
}
```

Asserting only the host name means the test passes if the uninstall fails for *any* reason — it does not check that the reason is the one the test is named after. Its install sibling (`:2550-2574`, same fixture) gets it right, including a discriminating negative:

```rust
assert!(err.reason.contains("did not come back after the reboot"), ...);
// Discriminating: the two reachable causes must not be claimed for a
// host that genuinely went away — they route the rollback differently.
assert!(!err.reason.contains("never rebooted") && !err.reason.contains("never received"), ...);
```

Bring the uninstall twin up to the sibling's standard.

## 4. `plan_provider_resolves_the_role_specific_table` pins difference, not mapping

`update_workflow/mod.rs:525-539` asserts only that the two doers differ:

```rust
assert_ne!(
    format!("{install:?}"),
    format!("{uninstall:?}"),
    "installer and uninstaller must not resolve to the same doer"
);
```

An adapter that **swapped** the two roles still passes, because they remain different. Its own comment states the intent it does not check: "The role string must actually select the table: an adapter that ignored it would make `uninstall` run the *install* command."

Lowest priority of the four, and I checked why before filing: a swap **is** caught one layer up — `tests/sl_report.rs:202-235` pins install to `zypper -n in -y -l pkg-a` and uninstall to `zypper -n rm pkg`, and `mod.rs:566-592` renders both roles across all six keys. So this is test hardening with no live hole behind it. Fix is to assert the rendered command text rather than inequality.

mimi1vx:
## Summary

- `crates/mtui-hosts/src/target/operation.rs`: `OperationGroup::last_output`'s doc called the `None` fallback benign, but `HostOutput::default()` is exit 0, which every check reads as success — the host silently passes and stays in the reboot map. Doc now names the hazard, the use site logs a WARN naming the host, and a `MockGroup` test pins the current fallback behaviour.
- Widened the check seam from a bare `String` to a typed `CheckFailure { reason, cancelled }` (mirroring `RebootFailure`'s shape), so a check that stops at a cancellation checkpoint can carry that fact across `mtui-hosts` into `UpdateError.cancelled` instead of being hardcoded to `false`.
- `perform_uninstall_reports_a_transactional_host_that_never_reconnects` only asserted the host name, so it passed for any reboot-failure cause. Brought it up to its install sibling's two discriminating assertions.
- `plan_provider_resolves_the_role_specific_table` asserted the two doers merely differ, which a role swap also satisfies. Now compares each resolved doer against one built from that role's own table entry.

## Testing

- `cargo fmt --all --check`
- `cargo test --workspace`
- `cargo build --workspace --no-default-features` and `--all-features`

Each new/strengthened test was observed failing against the unfixed/unswapped code, then restored, before landing.

No CHANGELOG entry — no user-visible change.

closes #408

### DIFF
--- a/crates/mtui-hosts/src/lib.rs
+++ b/crates/mtui-hosts/src/lib.rs
@@ -34,8 +34,8 @@ pub use prompter::{Prompter, Reader};
 #[cfg(test)]
 pub(crate) use target::POOL_LOCK_PATH;
 pub use target::{
-    Check, CheckArgs, Clock, Command, Doer, HostArbiter, HostOutput, HostPlan, HostsGroup,
-    InstallOperation, LockOutcome, Operation, OperationGroup, OperationReport, Owner,
+    Check, CheckArgs, CheckFailure, Clock, Command, Doer, HostArbiter, HostOutput, HostPlan,
+    HostsGroup, InstallOperation, LockOutcome, Operation, OperationGroup, OperationReport, Owner,
     PackageQuerier, PlanProvider, PoolLock, RebootFailure, RebootFailureCause, RemoteLock,
     RepoManager, RepoOp, SetRepo, Sink, SpinnerGuard, Suspend, SystemClock, TARGET_LOCK_PATH,
     Target, TargetLock, TtySpinner, UninstallOperation, get_arbiter, parse_system, set_test_sink,
--- a/crates/mtui-hosts/src/target/mod.rs
+++ b/crates/mtui-hosts/src/target/mod.rs
@@ -53,8 +53,9 @@ pub use hostgroup::{HostsGroup, LockOutcome};
 pub(crate) use locks::POOL_LOCK_PATH;
 pub use locks::{Clock, LockRow, PoolLock, RemoteLock, SystemClock, TARGET_LOCK_PATH, TargetLock};
 pub use operation::{
-    Check, CheckArgs, Doer, HostOutput, HostPlan, InstallOperation, Operation, OperationGroup,
-    OperationReport, PlanProvider, RebootFailure, RebootFailureCause, UninstallOperation,
+    Check, CheckArgs, CheckFailure, Doer, HostOutput, HostPlan, InstallOperation, Operation,
+    OperationGroup, OperationReport, PlanProvider, RebootFailure, RebootFailureCause,
+    UninstallOperation,
 };
 pub use package_querier::PackageQuerier;
 pub use parsers::parse_system;
--- a/crates/mtui-hosts/src/target/operation.rs
+++ b/crates/mtui-hosts/src/target/operation.rs
@@ -202,9 +202,12 @@ pub trait OperationGroup: Send {
 
     /// Reads back `hostname`'s post-run output, for the [`Check`] call.
     ///
-    /// Returns `None` for a host outside the group (should not happen for a
-    /// host with a [`HostPlan`]); the template treats that as an empty
-    /// snapshot rather than panicking.
+    /// A `None` here is a bug by construction: every caller only ever asks for
+    /// a hostname it just built a [`HostPlan`] for, so the host cannot be
+    /// outside the group. The template does not panic on it — it falls back to
+    /// [`HostOutput::default`], which is **exit 0**. Every [`Check`] reads exit
+    /// 0 as success, so a `None` here makes the host silently pass its check
+    /// and stay in the reboot map instead of failing loudly.
     fn last_output(&self, hostname: &str) -> Option<HostOutput>;
 
     /// Reboots the transactional hosts named in `reboot`.
@@ -405,7 +408,16 @@ pub trait Operation: Send + Sync {
 
         let mut check_failures: Vec<(String, String)> = Vec::new();
         for plan in &mut plans {
-            let output = group.last_output(&plan.hostname).unwrap_or_default();
+            let output = group.last_output(&plan.hostname).unwrap_or_else(|| {
+                // See `OperationGroup::last_output`: this host has a `HostPlan`,
+                // so it cannot really be missing from the group. Warn so the
+                // silent-pass is at least visible, rather than swallowing it.
+                tracing::warn!(
+                    host = %plan.hostname,
+                    "no post-run output for a planned host; treating as exit 0"
+                );
+                HostOutput::default()
+            });
             let args = CheckArgs {
                 hostname: &plan.hostname,
                 stdout: &output.stdout,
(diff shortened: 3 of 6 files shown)


plusky (changes requested):
All four items do what they claim, and the two judgement calls I checked hardest both hold: the `probe_failed` omission from `CheckFailure` is genuinely safe (the only producer is `UpdateError::probe_failure`, reached solely from the update path, which drives its own template and never crosses the `Operation` seam — and the doc says exactly that, as a decision rather than an oversight), and item 4's self-reference concern (expected doer built from the same table the adapter reads) is covered one layer up by `sl_report.rs` pinning the literal command text, as the issue itself already argued. The uninstall test now matches its install sibling assertion-for-assertion, including the discriminating negative. CI 16/16 on `024f2ad5`.

Two things to fix before merge — both, a little ironically, instances of this PR's own theme.

### 1. `aggregate_failures`'s `cancelled` comment now argues from a premise this PR removed

`reports/update_flow.rs`, the summary branch:

> `cancelled` is deliberately NOT propagated here. Every cancellation in this
> module is an early `return Err`, so no cancelled error ever reaches a
> `failures` vec — the line would be dead […]

That was true while the seam was `Result<(), String>`: a cancelled check was *unrepresentable*, so the vec provably never carried one. This PR's `perform_operation` now builds `UpdateError { cancelled: failure.cancelled, … }` straight from `report.check_failures` and hands the vec to `aggregate_failures` — the exact construction the comment says cannot happen. It stays dead *in practice* only because no production check emits `CheckFailure::cancelled` today, which is a statement about the current producers, not about the module's structure — one live producer away from false, and the comment gives a future reader a guarantee the types no longer make.

The behaviour itself is defensible either way (a lone cancelled check failure routes verbatim with the flag; a summary drops it, which for the mixed case is just the outranking rule — though note the all-cancelled case loses the flag too, unlike `probe_failed`'s `all()`). I am not asking for a behaviour change — just rewrite the comment to argue from the new premise: representable but unproduced, single-failure carries it, summary deliberately does not, and why. This is a "doc contradicting the code" being introduced by the PR that closes the doc-contradiction issue.

### 2. The headline seam fix is not pinned on the side that had the bug

`a_cancelled_check_failure_arrives_in_the_report_with_the_flag_intact` stops at `OperationReport` — the `mtui-hosts` side, which never had the bug. The lossy half was `perform_operation`'s map, and its fix is the five lines `cancelled: failure.cancelled`. Revert exactly that line to `cancelled: false` — the pre-PR behaviour — and nothing in the workspace goes red: the existing `is_cancelled` assertions all exercise flow-level checkpoint cancels, not the check seam. By this tree's own bar ("a pinning test that was never red pins nothing"), the one mutation this PR exists to prevent is currently uncaught.

Cheapest closure: drive `perform_install` (or `perform_uninstall`) with a provider whose check returns `CheckFailure::cancelled(…)` on the single-failure path and assert `err.is_cancelled()` — one test, and it red/greens on that exact line.

### Non-blocking

`missing_last_output_silently_passes_the_check_and_keeps_the_reboot` cannot see its own fixture wiring break: if `with_missing_output` stopped being consulted, `last_output` answers `Some(HostOutput::default())` — also exit 0 — and every assertion still passes, so the mutation the test names (fallback changed to fail the check) would silently stop being covered. A one-line liveness guard (`assert!(group.last_output("h1").is_none())` before the run, or asserting the new WARN) makes the fixture prove it is in the state the test's name claims. Same shape as the stub-sentinel guards the downgrade probe tests use.

plusky:
The three findings from my review are closed in follow-up #463 (they were in flight as commits on top of your head when this merged, so they land as their own PR; your four commits are untouched). The one design point worth your eye there: injecting the cancelling check for the new seam test needed `perform_operation` split into a thin delegator over `perform_operation_with(…, provider)` — production byte-identical, argued in the PR body.

Issue #401, opened 2026-07-30
CHANGELOG has no sections for released 26.0.2 and 26.0.3; blocks curated release notes


plusky:
`CHANGELOG.md` has no sections for two released versions, so the curated changelog and the published releases have drifted apart.

## State (checked against the tags in the tree)

| tag | `## [x.y.z]` section |
|---|---|
| 26.0.1 | present (`## [26.0.1] - 2026-07-22`) |
| 26.0.2 | **missing** |
| 26.0.3 | **missing** |
| 26.1.1 | present (`## [26.1.1] - 2026-07-27`) |

26.0.2's and 26.0.3's entries are still sitting under `[Unreleased]`, so `[Unreleased]` currently mixes shipped and unshipped work.

## Why it is worth more than tidiness

It blocks the better fix for release notes. #392 fixed the *base tag* selection for `--generate-notes`, which was picking a base that had already been released because releases are cut on per-minor branches (`26.0.x`, `26.1.x`) that fork from main and are never merged back, making consecutive tags siblings rather than ancestors. That fix works, but it is still deriving notes from commit ancestry.

If `chore(release): cut X.Y.Z` also moved `[Unreleased]` into a dated section, `gh release create --notes-file` on that section would give **curated** notes and make the base-tag inference moot. #392 deliberately left the process alone because that is a release-process change.

## Two pieces of work

1. Backfill sections for 26.0.2 and 26.0.3 from the git history between their tags. Note the CHANGELOG is append-only here — the diff for this must be purely additive to the released-history region.
2. Decide whether `chore(release): cut X.Y.Z` should cut `[Unreleased]` into a dated section, and whether release notes should then come from the CHANGELOG rather than `--generate-notes`.

(1) is mechanical. (2) is the one worth a maintainer's opinion.

mimi1vx:
sub releases in changelog are only in stable branches :D

plusky:
Part 1 is up as PR #462: it adds `## [26.0.2] - 2026-07-22` and `## [26.0.3] - 2026-07-23` on main between `[26.1.1]` and `[26.0.1]`, purely additively (36 insertions, 0 deletions), with every line copied byte-for-byte from the sections the 26.0.x branch itself carries at those tags.

[...]

PR #462, opened 2026-08-17
docs(changelog): backfill the released 26.0.2 and 26.0.3 sections

plusky:
Part 1 of #401.

Deliberately not "Closes": part 2 (release notes from the CHANGELOG instead of `--generate-notes`) is posted as a proposal comment on the issue and needs a maintainer decision, so #401 stays open after this merges.

## What this does

Adds the missing sections for every released tag that had none — `## [26.0.2] - 2026-07-22`, `## [26.0.3] - 2026-07-23`, and (per review) `## [26.1.2] - 2026-08-04` — and removes each entry from every section other than the release that first shipped it.

## How each section's content was established

Every inserted line is copied **byte-for-byte from the CHANGELOG the release branch itself carries at that tag** (`git show <tag>:CHANGELOG.md`), cross-checked against the commit ranges between tags (`git log 26.0.1..26.0.2`, `26.0.2..26.0.3`, `26.1.1..26.1.2`) with each release-branch commit matched to its main-branch counterpart. Section dates are the tag dates (`git log -1 --format=%cs <tag>`). A scripted verifier pins provenance: every added body line exists byte-identical in the corresponding tag's file; sub-ranges are compared explicitly.

## De-duplication (second commit, per review)

The 26.1.1 cut had swept the 26.0.2/26.0.3 entries into `[26.1.1]`, so the first commit's purely-additive backfill initially duplicated three bullets. On review direction the duplication is gone: `[26.1.1]` loses its copies of the `reboot_timeout` keys, the slow-reboot false-failure, and the refhosts-cache abort entries, and the 26.1.2 fix's differently-worded record leaves `[Unreleased]` (it described released behaviour). A whole-file scan shows zero duplicated entry blocks; `[Unreleased]` now holds only unshipped work. This deliberately edits a released section — maintainer-directed, superseding the issue's purely-additive constraint: one entry, one section, first shipment wins.

## Verification

- `typos` clean (the load-bearing check for this file, proven with a positive control: an injected known misspelling in the new region is caught at its exact line).
- Diff touches `CHANGELOG.md` only.
- `## [26.1.2]` byte-identical to tag `26.1.2`'s own section; 26.0.x sections byte-identical to their tag sub-ranges (26.0.3: tag lines 11–32; 26.0.2: tag lines 11–22 — the tags' whole sections include entries that first shipped in earlier releases, excluded here by the first-shipment rule).
- Zero duplicated entry blocks file-wide after the second commit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -127,10 +127,6 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
 - An unreachable openQA instance now reports the actual transport failure
   (e.g. `"...: Connection refused (os error 61)"`) instead of restating the
   request URL (`"...: error sending request for url (http://.../api/v1/jobs)"`).
-- MCP calls passing a list to a repeatable single-value flag (e.g. `approve`
-  with two `group`s) no longer fail to parse: the reconstructed argv now
-  repeats the flag before every element instead of emitting it once for the
-  whole list.
 - `approve` no longer approves a Gitea update whose PR head differs from the
   checked-out testreport without saying so. Interactively it now logs the
   expected/actual hash pair and asks for confirmation (default no); a missing
@@ -496,6 +492,16 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   underlying certificate/IO cause instead of `reqwest`'s error kind plus URL,
   which is both more actionable and URL-free by construction.
 
+## [26.1.2] - 2026-08-04
+
+### Fixed
+
+- MCP tool calls passing an array with more than one element (e.g. `add_host`
+  `target`, `openqa_overview` `aggregated_groups`, `list_refhosts` `arch`) no
+  longer fail with `unexpected argument '<second element>' found`: the
+  reconstructed argv now repeats the flag before every element instead of
+  emitting it once for the whole list.
+
 ## [26.1.1] - 2026-07-27
 
 ### Added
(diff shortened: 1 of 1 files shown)


mimi1vx:
@plusky, when updating the changelog, please ensure that entries are added for all release types—not solely patch releases—and remove any redundant records. 

I fully endorse any well-structured proposal for the release and changelog process. My preference is to maintain the main branch as the development line, with stable branches designated for major releases and their corresponding patch updates.

mimi1vx (changes requested):
needs remove duplicate records , … and same for 26.1.2

plusky:
Both points addressed in `c9b7485d`:

**All release types** — `## [26.1.2] - 2026-08-04` is added, copied byte-for-byte from the section the 26.1.x branch carries at the tag, same derivation as the 26.0.x sections. Its fix was also described by a differently-worded entry still under `[Unreleased]` (the `repeatable single-value flag` record); that described released behaviour, so it leaves `[Unreleased]` — which now holds only unshipped work.

**Redundant records** — the three bullets the backfill had duplicated between `[26.1.1]` and the 26.0.x sections now exist exactly once, in the section of the release that first shipped them; `[26.1.1]` loses its copies. A whole-file scan confirms zero duplicated entry blocks remain. This edits a released section, which the PR had previously avoided on the issue's purely-additive wording — your direction settles that: one entry, one section, first shipment wins.

The PR body is updated to match (the "duplication compromise" section it previously argued is gone).

On the process preference (main as development line, stable branches per major with their patch updates): that matches how the tree already works, and the proposal on #401 builds on exactly that — cut-time `[Unreleased]` → dated section on main, notes from the section. If the recommendation there works for you, I'll implement it as its own PR.

mimi1vx (approved):

PR #463, opened 2026-08-17
test,docs(hosts,testreport): close the #461 review findings

plusky:
Follow-up to #461, which merged while these commits were in flight; they close the three findings from [its review](https://github.com/openSUSE/mtui/pull/461#pullrequestreview-6482451190) plus two accuracy defects an adversarial pass found in my own first rewrite. #408 stays closed — this completes its cleanup rather than reopening it.

## What each commit does

- **`c4ff574c`** rewrites `aggregate_failures`'s `cancelled` comment from the premise #461 removed: a cancelled check failure is now *representable* across the seam (`CheckFailure.cancelled` → `UpdateError.cancelled`) and merely unproduced — the old text claimed the vec could never carry one, which the types no longer guarantee. Comment-only; behavior untouched (verified mechanically: every changed line is a comment line).
- **`cfe0d02a`** adds the test the seam fix was missing: a provider whose check returns `CheckFailure::cancelled(…)` driven through `perform_install`, single-failure verbatim path, asserting `err.is_cancelled()`. **Observed red exactly on the fixed line**: reverting `cancelled: failure.cancelled` → `cancelled: false` fails only this test (338 passed / 1 failed, panic at the new assertion), restoring goes green. Getting the provider injected needed a seam: `perform_operation` becomes a thin delegator over `perform_operation_with(…, provider)` — production byte-identical, the single `set_plan_provider` call site stays inside the shared body, so the inject-at-the-point-of-use rule holds. That split is the one judgement call here worth a look.
- **`0d359f42`** makes the missing-output fixture prove its own wiring: `assert!(group.last_output("h1").is_none())` before the run, so the test named after the `None` fallback can no longer pass while silently exercising `Some(default)`. Also observed red (dropping `.with_missing_output()` fails only this test).
- **`a9ead622`** fixes an overclaim my first rewrite introduced: the `cancelled` flag routes *reporting* (`map_flow_error` → `CommandError::Cancelled` — its only non-test consumer), **not** the rollback, which `perform_update_with_rollback` dispatches purely on the `UpdateFailure` variant. The comment now says exactly that and what a rollback-skipping cancel would additionally need.
- **`f7243e3a`** renames the four comment sites that still said `perform_operation` for the map that now lives in `perform_operation_with`.

## Gate

Full DoD gate observed green on this exact tree: fmt, clippy 0, rustdoc `--all-features` 0, workspace 2556 passed / 0 failed (36 suites), `-F mcp` 291 passed / 0 failed, both feature-matrix builds, typos clean, no MUTANT markers, clean status. The branch was rebased onto main from a base whose tree is byte-identical to the merge (`git diff` between the gated head and this head is empty), so the gate results describe these bytes.

### DIFF
--- a/crates/mtui-hosts/src/target/operation.rs
+++ b/crates/mtui-hosts/src/target/operation.rs
@@ -1245,6 +1245,14 @@ mod tests {
             check,
         }];
         let mut group = MockGroup::new(Ok(plans)).with_missing_output();
+        // Liveness guard: the fixture *is* the test. If `with_missing_output`
+        // stopped being consulted, `last_output` would answer
+        // `Some(HostOutput::default())` — also exit 0 — and every assertion
+        // below would keep passing while covering nothing.
+        assert!(
+            group.last_output("h1").is_none(),
+            "the fixture must really be scripting a missing snapshot"
+        );
 
         let op = InstallOperation::new(strs(&["pkg-a"]));
         let report = op.run(&mut group).await.expect("the run started");
--- a/crates/mtui-testreport/src/reports/update_flow.rs
+++ b/crates/mtui-testreport/src/reports/update_flow.rs
@@ -559,10 +559,19 @@ fn aggregate_failures(op: &str, mut failures: Vec<UpdateError>) -> Result<(), Up
         // The all-cancelled case loses the flag too, which `probe_failed`'s
         // `all()` above would have kept. That asymmetry is deliberate: with no
         // live producer there is no all-cancelled population to summarise, and
-        // minting a group-wide cancel verdict — which routes rollback and
-        // reporting differently — for a case that cannot yet occur would be a
-        // behaviour decision taken blind. Make it here, with `all()`, if a
-        // check ever does cancel.
+        // minting a group-wide cancel verdict for a case that cannot yet occur
+        // would be a behaviour decision taken blind. Make it here, with
+        // `all()`, if a check ever does cancel — but know what that verdict
+        // does and does not reach. The flag routes *reporting*:
+        // `commands/perform.rs::map_flow_error` is its only non-test reader,
+        // and it turns the error into `CommandError::Cancelled`. It does not
+        // route the rollback — that follows the `UpdateFailure` variant
+        // `update_run_phase` picks from `repairable`/`probe_failed`, which
+        // never inspects `cancelled`. Setting it here alone would tell the
+        // operator the run was cancelled while the group-wide downgrade ran
+        // on every healthy host anyway; skipping that rollback the way
+        // `UpdateFailure::Cancelled` does needs a matching change at the
+        // `wrap` site.
         Err(aggregate)
     }
 }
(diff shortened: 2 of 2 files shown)


mimi1vx (changes requested):
Behaviour-neutral and correct as far as I can check it: I fetched the head, read both files in full, and every factual claim in the new prose holds — `commands/perform.rs::map_flow_error` really is the only non-test reader of `is_cancelled`, no `update_workflow::checks` fn emits `CheckFailure::cancelled` today (only this test and `crates/mtui-hosts/tests/operation_group.rs` construct one), and the rollback really does route off the `UpdateFailure` variant at the `wrap` site rather than the flag. `fmt`, `clippy --all-targets -D warnings` and the two touched test modules are green on my machine too, and the `perform_operation` → `perform_operation_with` split is the right seam: `set_plan_provider` overwrites unconditionally, so there is genuinely no other way in, and the single call site stays inside the shared body.

Two things to fix before merge, both in the prose rather than the code.

**1. The same argument now lives at three sites.** "No production check emits `CheckFailure::cancelled`" is asserted in the `aggregate_failures` inline comment (`update_flow.rs` ~538-575), in `perform_operation_with`'s doc (~662-676), and in the `aggregate_failures_carries_the_typed_flags_through_the_summary` comment (~3462-3468). It is true today, but the first check that *does* cancel invalidates three paragraphs across two crates, and whoever writes that check will find at most one of them. Please keep the authoritative statement at the omission site — `aggregate_failures` — and reduce the other two to a pointer at it. One place to be wrong is the point of the comment.

**2. 36 comment lines for one absent assignment, four of them prescriptive.** The load-bearing content is the first paragraph (representable but unproduced — do not read the emptiness as a structural guarantee) plus the outranking rule (a cancel must not mask a real failure collected beside it). The all-cancelled asymmetry paragraph, and specifically "Make it here, with `all()`, if a check ever does cancel — but know what that verdict does and does not reach", prescribes a design for a state that cannot occur; that is the comment form of handling an impossible case, and it will be stale advice by the time anyone needs it. Two lines noting that the all-cancelled case has no producer to summarise is enough — the rollback-vs-reporting distinction from `a9ead622` is worth keeping, just not the instruction.

Nits, take or leave: `perform_operation_with`'s doc says production *isn't* reached with a non-registry provider; it would be stronger as "must not be" — an invariant rather than an observation. And `CancellingProvider` spells `mtui_hosts::` out five times while the test module already has `use mtui_hosts::{HostsGroup, MockConnection, Target};`.

I did not independently reproduce the two observed-red mutations; taking those on the description.

plusky:
Both points addressed in `a0dffcc7`: the producer claim now lives only at `aggregate_failures`'s cancelled comment with the other two sites reduced to pointers, and that comment loses its prescriptive all-cancelled paragraph — what stays is representable-but-unproduced, the outranking rule, and the reporting-vs-rollback distinction. Both nits taken: "must not be" invariant phrasing, and `CancellingProvider` on the module's existing imports (the trait impl line stays qualified — importing `PlanProvider` makes every `reg.doer()`/`reg.check()` in the module ambiguous). Comment-only except the import line; workspace re-run green (37 suites / 2556 / 0).

mimi1vx (approved):

PR #470, opened 2026-08-20
test(testreport): make an empty probe.ran diagnosable

### LINKED ISSUE #464: Flaky liveness sentinel in rows_filtered_to_nothing_is_a_noop (slmicro leg): stub zypper recorded no invocations on CI
One occurrence so far, on an unrelated PR's CI run (#463, whose diff does not touch this file): [run 32023998502](https://github.com/openSUSE/mtui/actions/runs/32023998502), `MTUI tests`:

```
thread 'update_workflow::actions::downgrade::tests::rendered_script::rows_filtered_to_nothing_is_a_noop' panicked at crates/mtui-testreport/src/update_workflow/actions/downgrade.rs:962:
slmicro: the script never reached the stub zypper at all: []
```

## What is known

- The assertion that fired is the harness's own **liveness sentinel**: the stub `zypper`'s first act is `printf '%s\n' "$2" >> "$MTUI_STUB_DIR/probe.ran"`, and `probe_invocations()` came back empty — so either `/bin/sh` never executed the stub, or the append itself failed. This is the sentinel doing exactly what it was added for: before it, this state would have been an invisible vacuous pass.
- The **zypper leg of the same test passed seconds earlier in the same process** with an identical fresh `Stubs` (own `tempfile::tempdir`, own `probe.ran`), so the failure is not systematic to the harness shape — the two legs differ only in the `ActionCommands` they render, and both share `LIST_COMMAND`.
- **Not an in-process env race**: `grep -rn 'set_var\|remove_var' crates/mtui-testreport/{src,tests}` is empty, so nothing in the one-binary suite mutates `PATH` (captured at `Stubs::new`) or the stub variables concurrently.
- The workspace run was otherwise green (338 passed / 1 failed), and the same tree passed `cargo test --workspace` locally in full.

## Candidates, none proven

- A transient failure of the `>>` append on the runner's filesystem (the stub keeps going after a failed `printf` — the script has no `set -e` — so the run can look healthy while the log stays empty; on this occurrence the sentinel fired before any other assertion could say).
- A transient spawn/exec hiccup of the stub under `/bin/sh -c` on the shared runner.

## If it recurs

Two cheap hardenings that would discriminate: (a) have the stub `exit 97` when the append fails (`printf ... || exit 97`), so "recording broke" separates from "never ran"; (b) on sentinel failure, dump `ran.code`/`ran.stdout` in the panic message — this occurrence discarded them, and they would have said whether the script ran at all.

plusky:
Closes #464.

The one CI occurrence of `rows_filtered_to_nothing_is_a_noop` (slmicro leg) discarded the only evidence: `probe.ran` was empty, and the sentinel panic printed only the empty invocation list. The stub's first act is an unguarded `>>` with no `set -e`, so a failed append still looks like a healthy probe. `run_script` also dropped stderr, which is where `zypper: not found` / `Is a directory` would have been.

This does not retry the unknown runner glitch. It makes the next occurrence diagnosable, and pins "recording broke" so it cannot look like "never ran".

## What changed (test harness only)

- Stub appends now `|| exit 97` (`downgrade.rs` `ZYPPER_STUB`; `update.rs` `ZYPPER_STUB` `probe.ran`/`cleanup.ran`/`patch.ran` and `TU_STUB`). 97 is outside zypper's space and not in the probe-guard accepted set (`0`/`104`/`106`).
- `Ran` keeps `stderr`. The nine copy-pasted "never reached the stub" asserts go through `assert_reached_stub`, which dumps invocations plus `code`/`stdout`/`stderr`.
- `a_failed_sentinel_append_exits_97` in both `rendered_script` modules: `probe.ran` is pre-created as a directory, healthy knobs, both templates; asserts code 97 and empty invocations.

No CHANGELOG (test-only). Production `LIST_COMMAND` / templates are untouched.

## Proof

Red-then-green of the pinning tests (AGENTS.md): with `|| exit 97` dropped, both tests failed `left: 0` / `right: 97` (healthy stdout, stderr `Is a directory`). Restored, green.

Local gate observed green (not predicted):

- `cargo fmt --all --check`
- `cargo test --workspace`
- `cargo build --workspace --no-default-features`
- `cargo build --workspace --all-features`

Adversarial review of the diff: 0 issues.

### DIFF
--- a/crates/mtui-testreport/src/update_workflow/actions/downgrade.rs
+++ b/crates/mtui-testreport/src/update_workflow/actions/downgrade.rs
@@ -478,7 +478,8 @@ mod tests {
         /// Every arm records its invocation, which is the liveness signal the
         /// negative assertions rest on.
         const ZYPPER_STUB: &str = r#"#!/bin/sh
-printf '%s\n' "$2" >> "$MTUI_STUB_DIR/probe.ran"
+# A failed append must not look like a healthy probe (no set -e in the stub).
+printf '%s\n' "$2" >> "$MTUI_STUB_DIR/probe.ran" || exit 97
 case "$2" in
 se)
     if [ -n "$MTUI_STUB_SE_ROWS" ]; then printf '%s\n' "$MTUI_STUB_SE_ROWS"; fi
@@ -575,6 +576,8 @@ exit 0
             /// The script's stdout — what the flow parses into the version map,
             /// and what `show_log` shows the operator.
             stdout: String,
+            /// The script's stderr, captured for sentinel-miss diagnostics.
+            stderr: String,
         }
 
         impl Ran {
@@ -627,7 +630,20 @@ exit 0
                     .code()
                     .unwrap_or_else(|| panic!("script died on a signal: {:?}", out.status)),
                 stdout: String::from_utf8_lossy(&out.stdout).into_owned(),
+                stderr: String::from_utf8_lossy(&out.stderr).into_owned(),
             }
+        }
+
+        fn assert_reached_stub(name: &str, stubs: &Stubs, ran: &Ran, expected: &str) {
+            let invocations = stubs.probe_invocations();
+            assert!(
+                invocations.iter().any(|c| c == expected),
+                "{name}: the script never reached the stub zypper at all: {invocations:?}\n\
+                 code: {}\nstdout: {}\nstderr: {}",
+                ran.code,
+                ran.stdout,
+                ran.stderr,
+            );
         }
 
         #[test]
@@ -653,11 +669,7 @@ exit 0
                 );
                 // Liveness first: the negative below is also what an empty
                 // `PATH` produces.
-                assert!(
-                    stubs.probe_invocations().iter().any(|c| c == "se"),
-                    "{name}: the script never reached the stub zypper at all: {:?}",
-                    stubs.probe_invocations()
-                );
+                assert_reached_stub(name, &stubs, &ran, "se");
                 assert_eq!(
                     ran.marker(),
                     Some(format!("{PROBE_MARKER}: zypper -n se exited 7").as_str()),
@@ -783,11 +795,7 @@ exit 0
                         ..Knobs::default()
                     },
                 );
-                assert!(
-                    stubs.probe_invocations().iter().any(|c| c == "se"),
-                    "{name}: the script never reached the stub zypper at all: {:?}",
-                    stubs.probe_invocations()
-                );
+                assert_reached_stub(name, &stubs, &ran, "se");
                 assert_eq!(
                     ran.marker(),
                     Some(format!("{PROBE_MARKER}: awk exited 2").as_str()),
@@ -838,11 +846,7 @@ exit 0
                         ..Knobs::default()
                     },
                 );
-                assert!(
-                    stubs.probe_invocations().iter().any(|c| c == "se"),
-                    "{name}: the script never reached the stub zypper at all: {:?}",
-                    stubs.probe_invocations()
-                );
+                assert_reached_stub(name, &stubs, &ran, "se");
                 assert_eq!(
                     ran.marker(),
                     None,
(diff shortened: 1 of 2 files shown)


mimi1vx (changes requested):
Reviewed the diff, ran the affected tests, and reproduced the red/green
mutation independently (dropped `|| exit 97` from both `ZYPPER_STUB`s → both
`a_failed_sentinel_append_exits_97` cases failed `left: 0` / `right: 97` with
stderr `Is a directory`; restored → green). `cargo test -p mtui-testreport
--lib update_workflow::actions` is 45/45 here, and CI is green.

Scope is right: test-only, production templates and `LIST_COMMAND` untouched,
no contract touched, no CHANGELOG needed. Two things to change first.

## Requested changes

**1. `assert_reached_stub` needs `#[track_caller]`**
(`downgrade.rs:637`, `update.rs:826`)

Each of the nine inline `assert!`s used to report its own call site. Extracted
into a plain helper, all of them now panic at the helper's line, so the
failure no longer says *which* assertion in the test body fired — a
diagnosability regression inside a diagnosability PR. One attribute restores
it.

**2. The second assertion in `a_failed_sentinel_append_exits_97` cannot fail**
(`downgrade.rs:998`, `update.rs:1358`)

`probe.ran` is a directory for the whole test, and `probe_invocations()` /
`sentinel()` swallow the resulting `EISDIR` with `unwrap_or_default()` — so an
empty vec comes back unconditionally, independent of the stub *and* of the
template. It would still pass with `|| exit 97` removed, with the stub
removed, and with the guard removed.

That also means the PR's "pins *recording broke* so it cannot look like
*never ran*" is carried entirely by the `ran.code == 97` assertion above it;
this one adds no evidence. Per AGENTS.md ("a test that cannot fail is not
evidence") please drop it, or make it discriminating — e.g. assert the guard's
marker line names the probe (`… : zypper -n se exited 97`), which does depend
on the propagation path.

## Suggestions (non-blocking)

- **The residual #464 ambiguity is in `sentinel()`'s `unwrap_or_default()`**
  (`update.rs:715`, `downgrade.rs:514`): "read failed" and "nothing was
  recorded" still collapse to the same observation — precisely what discarded
  the evidence on the slmicro leg. `|| exit 97` only helps where the stub is
  reached *and* its status survives to the script's exit; the unchecked
  `zypper -n lr -puU`, `refresh` and the `patches | grep` pipeline still
  swallow it. Having `assert_reached_stub` report the sentinel's fs state
  (exists / is_dir / the `io::Error`) would close the rest cheaply.
- **`97` is a bare literal at five stub sites and two assertions across two
  files.** A named const interpolated into the stub bodies would stop the two
  copies drifting, and documents the one latent collision: a future
  `Knobs { patch_exit: 97 }` would be indistinguishable from a sentinel
  failure.
- **Description nit:** the accepted set `0`/`104`/`106` is *downgrade*'s guard.
  `update`'s accepts `0`, plus `106` only when `mtui_issue_repo_present`.
  Harmless — 97 is outside both — but the two guards do not share that shape.
- `assert_reached_stub` is duplicated verbatim in both modules, the same
  copy-paste being removed within each file. Fine as-is (no shared test-util
  module in this crate), just noting it.

plusky:
Addressed the two requested changes in f1853a24:

1. **`#[track_caller]`** on `assert_reached_stub` in both modules, so a missed stub still names the test line rather than the helper.
2. **Dropped the vacuous empty-`probe.ran` assertion** (`EISDIR` → `unwrap_or_default()`). Kept `ran.code == 97` and now assert the guard marker that names the probe and status 97 (`zypper -n se` / `zypper -n patches`). That is the path `mtui_probe_ok` must have seen; it fails if the stub or the guard is removed.

Red-then-green: dropping `|| exit 97` from each `ZYPPER_STUB` failed both `a_failed_sentinel_append_exits_97` cases with code 0 and no marker; restored, green. `cargo test -p mtui-testreport --lib update_workflow::actions` is 45/45.

Left the non-blocking suggestions (named `97` const, sentinel fs-state in the helper, unifying the two copies) for a follow-up if you still want them.

mimi1vx (approved):

PR #471, opened 2026-08-21
refactor: name repeated magic-number literals

mimi1vx:
## Summary

Names the repeated numeric literals in `mtui-testreport`'s zypper
exit-code checks, plus a handful of smaller repeated magic numbers
elsewhere in the workspace, and folds the zypper `install` check onto
`update`'s exit-code/marker classifier so the two cannot silently
diverge on the same transcript.

## Changes

- **Exit-code constants** (`checks::mod`): `ZYPPER_EXIT_*` and
  `EXIT_NOT_RUN` replace bare `100`-`107`/`4`/`5`/`8`/`-1` literals
  across the update/downgrade/prepare/install checks — pure
  literal-to-name, no behaviour change.
- **`install::zypper` now delegates to `update::classified`** instead
  of restating the same three exit-code sets inline in a different
  order. This is a real behaviour change on the zypper release keys
  (`11`/`12`/`15`/`16`, `install`/`uninstall`): reason strings for
  `4`/`5`/`8` and for a command that never ran now match `update`'s;
  see the CHANGELOG entry for the exact deltas.
- **Small repeated constants**: `MAX_ERROR_SOURCE_LINKS`
  (mtui-datasources), `DEFAULT_SSH_PORT` (mtui-hosts),
  `DEFAULT_TERM_COLS`/`DEFAULT_TERM_ROWS` (mtui-core, shared with
  mtui-cli's shell resize fallback).
- **One host-close timeout**: `quit`'s, `template_registry`'s and
  `mtui-mcp`'s separately-declared 45s `Duration` constants are now a
  single `Session::HOST_CLOSE_TIMEOUT`.

## Testing

- `cargo fmt --all --check`
- `cargo test --workspace`
- `cargo build --workspace --no-default-features`
- `cargo build --workspace --all-features`
- New/changed behaviour in the install-check fold was verified red
  before green (a role-neutral timeout gate, a both-markers ordering
  case, and the install/update reason-string convergence).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,16 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
 
 ### Changed
 
+- The zypper `install`/`uninstall` check now shares the `update` check's exit
+  code/marker classifier instead of restating it inline, so the two cannot
+  drift apart. Reason-string changes on `install`/`uninstall` for the
+  `("11"|"12"|"15"|"16", false)` keys: exit `4`/`5`/`8` now let a locked
+  update stack, dependency prompt or RPM error name the failure instead of
+  always answering "package not found" (only `104` still short-circuits to
+  it); a transcript carrying both a dependency prompt and an `Error:` line now
+  answers "Dependency Error" instead of "RPM Error"; and a command that never
+  produced an exit status now answers "command timed out or failed to run"
+  (previously "Unknown Error").
 - The REPL line editor moved to reedline 0.50. User-facing deltas: the
   terminal painter no longer misplaces the cursor once a line fills the full
   terminal width, a bottom-flush prompt no longer reuses a stale anchor after
--- a/crates/mtui-testreport/src/update_workflow/checks/mod.rs
+++ b/crates/mtui-testreport/src/update_workflow/checks/mod.rs
@@ -91,6 +91,20 @@ pub struct CheckArgs<'a> {
 /// stable reason string otherwise.
 pub type CheckFn = Box<dyn Fn(CheckArgs<'_>) -> Result<Vec<Diagnostic>, UpdateError> + Send + Sync>;
 
+/// zypper's documented exit codes, by their `zypper(8)` names.
+pub(crate) const ZYPPER_EXIT_ERR_ZYPP: i32 = 4;
+pub(crate) const ZYPPER_EXIT_ERR_PRIVILEGES: i32 = 5;
+pub(crate) const ZYPPER_EXIT_ERR_COMMIT: i32 = 8;
+pub(crate) const ZYPPER_EXIT_INF_UPDATE_NEEDED: i32 = 100;
+pub(crate) const ZYPPER_EXIT_INF_SEC_UPDATE_NEEDED: i32 = 101;
+pub(crate) const ZYPPER_EXIT_INF_REBOOT_NEEDED: i32 = 102;
+pub(crate) const ZYPPER_EXIT_INF_RESTART_NEEDED: i32 = 103;
+pub(crate) const ZYPPER_EXIT_INF_CAP_NOT_FOUND: i32 = 104;
+pub(crate) const ZYPPER_EXIT_INF_REPO_SKIPPED: i32 = 106;
+pub(crate) const ZYPPER_EXIT_INF_RPM_SCRIPT_FAILED: i32 = 107;
+/// mtui's own "the command never produced a status" sentinel — *not* zypper's.
+pub(crate) const EXIT_NOT_RUN: i32 = -1;
+
 /// What a package manager's exit status says about the run.
 ///
 /// zypper's status is not a boolean. Quoting `man zypper` (verified against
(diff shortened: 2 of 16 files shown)


CHANGELOG.md
```diff
@@ -30,6 +30,16 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
 
 ### Changed
 
+- The zypper `install`/`uninstall` check now shares the `update` check's exit
```
plusky, line 33: **[suggestion]** The listed deltas are all remaps of cases that already failed (`4`/`5`/`8` + marker, dual markers, `-1`). They omit a pass→fail: old `install::zypper` returned `Ok(Vec::new())` for `0|100|101|102|103|106|107` without reading the transcript; `classified` runs `markers` on `ExitClass::Success`, so a lock line, `(c): c`, or `Error:` on those exits now fails the install/uninstall. That is reachable on 106 — zypper's skip-repo path (`ZYPPER_EXIT_INF_REPOS_SKIPPED`) calls `out().error(...)`, and `markers` treats stderr `Error:` as "RPM Error". `success_exit_codes_pass` only feeds empty streams, so it cannot catch this. New Diagnostics on a clean success are correctly kept out of the changelog (`PlanProvider` info-logs them and never reaches `perform::render_diagnostics`).

**Suggestion:** If the success-class marker run is intended (matching update), add it to this entry so operators know a skipped-repo / dependency-prompt transcript can now fail an install that previously passed; pin it with a zypper-install case analogous to `slmicro_lock_message_on_a_clean_exit_is_stack_locked` (empty-transcript success is not enough). If 106 should still pass, do not send Success through `markers` on this table.

crates/mtui-testreport/src/update_workflow/checks/mod.rs
```diff
@@ -160,7 +174,7 @@ pub(crate) enum ExitClass {
 /// bent into this shape (see the note there).
```
plusky, line 174: **[suggestion]** `classify_exit`'s docs still say the install check "still spells the same three sets out inline, interleaved with its stderr markers, and is left alone rather than being bent into this shape". After `809877b5`, `install::zypper` is `classified`. The next sentence ("What a check then *says* about the three inaccurate members is its own decision: the `update` check lets the transcript name them") is the same leftover: both tables now share the message as well as the sort.

**Suggestion:** Drop the "left alone" paragraph and the "its own decision" split. State once that `classify_exit` + `classified` is the shared verdict for the zypper install/update keys (install still gates `-1` itself so uninstall does not inherit "update command timed out…").

crates/mtui-testreport/src/update_workflow/checks/mod.rs
```diff
@@ -91,6 +91,20 @@ pub struct CheckArgs<'a> {
...
+/// zypper's documented exit codes, by their `zypper(8)` names.
+pub(crate) const ZYPPER_EXIT_ERR_ZYPP: i32 = 4;
+pub(crate) const ZYPPER_EXIT_ERR_PRIVILEGES: i32 = 5;
+pub(crate) const ZYPPER_EXIT_ERR_COMMIT: i32 = 8;
+pub(crate) const ZYPPER_EXIT_INF_UPDATE_NEEDED: i32 = 100;
+pub(crate) const ZYPPER_EXIT_INF_SEC_UPDATE_NEEDED: i32 = 101;
+pub(crate) const ZYPPER_EXIT_INF_REBOOT_NEEDED: i32 = 102;
+pub(crate) const ZYPPER_EXIT_INF_RESTART_NEEDED: i32 = 103;
+pub(crate) const ZYPPER_EXIT_INF_CAP_NOT_FOUND: i32 = 104;
+pub(crate) const ZYPPER_EXIT_INF_REPO_SKIPPED: i32 = 106;
```
plusky, line 103: **[nit]** The block is documented as "zypper's documented exit codes, by their `zypper(8)` names". `zypper-main.h` and the EXIT CODES section of `zypper(8)` spell 106 as `ZYPPER_EXIT_INF_REPOS_SKIPPED` (plural *REPOS*). The value is correct; the identifier is not the man-page name.

**Suggestion:** Rename to `ZYPPER_EXIT_INF_REPOS_SKIPPED` (or note the deviation next to the const). No behaviour change.

plusky (commented):
## Summary

The four-commit series is a clean refactor: `ZYPPER_EXIT_*` / `EXIT_NOT_RUN` match zypper 1.14.98 / `zypper(8)` numeric values at every comparison site, `HOST_CLOSE_TIMEOUT` is the single 45s budget at all four close paths (with the test-only override accessors kept), and the install/update fold does what the PR claims for exits 4/5/8, dual-marker order, and the role-neutral `-1` gate (uninstall still shares `install_check`). No MCP / lock / RRID / export contracts move. The remaining gaps are a CHANGELOG hole around success-class exits now running markers (a pass→fail, not a reason remap), a now-false `classify_exit` doc that still says install restates the sets inline, and a one-letter mismatch with zypper's `REPOS_SKIPPED` name.

## Issue counts by severity

- bugs: 0
- suggestions: 2
- nits: 1

plusky (approved):
Approving: no blocking issues. The COMMENT review (https://github.com/openSUSE/mtui/pull/471#pullrequestreview-4989944557) has three non-blocking notes (CHANGELOG success-class marker pass→fail, stale `classify_exit` docs, `REPOS_SKIPPED` identifier). CI is green.

PR #472, opened 2026-08-21
feat(testreport): list_metadata reports RRID, Rating and Gitea PR

mimi1vx:
## Summary

- Drop the `Hosts` row from `list_metadata`'s shared row set — it read `TestReportBase::systems`, a field no production code ever wrote (only `remove_host` cleared it), so the row was always empty and always dropped anyway.
- Delete `TestReportBase::systems` entirely now that nothing reads it.
- Add `ReviewRequestID`, `Rating` and `Gitea PR` rows, sourced from `base.rrid`/`base.rating`/`base.giteapr`. These are parsed from every metadata envelope but were previously displayed nowhere — for a git-served template that meant `list_metadata` showed neither its RRID nor the Gitea PR that distinguishes it from an OBS template.

No per-report override or `update_source` dispatch needed: the existing empty-value-drop behavior already does the right thing per template (Null report has no RRID → omitted; OBS/PI report has no Gitea PR → omitted).

## Test plan

- `cargo test -p mtui-testreport show_yourself` (extended fixture + assertions, observed red before the fix, green after)
- `cargo test -p mtui-core listmetadata`
- `cargo fmt --all --check`
- `cargo test --workspace`
- `cargo build --workspace --no-default-features` / `--all-features`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -88,6 +88,9 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
   update-repository URL layout for a git-served `SLFO:1.1` update (previously
   always treated as OBS-served) and restores the approve-time Gitea
   checkout-hash comparison for it (previously always skipped).
+- `list_metadata` now reports `ReviewRequestID`, `Rating` and the Gitea PR
+  where the template carries one, and no longer emits the always-empty
+  `Hosts` row.
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


plusky (approved):
Looks correct. `TestReportBase::systems` has no production writer (only `remove_host` cleared it), so the `Hosts` row was always empty and dropped. Putting `ReviewRequestID` / `Rating` / `Gitea PR` in that slot is the right user-visible gap, and the existing empty-value drop already omits them on Null/OBS templates. CHANGELOG is in the right Unreleased/Changed place. MCP schema is unchanged (output text only).

Non-blocking: `show_yourself_data_drops_empty_and_sorts` asserts the new labels exist but not their values, and does not pin `Hosts` as gone (`assert!(!has("Hosts"))`). Worth adding if you touch the test again.

Holding merge until CI finishes.

PR #474, opened 2026-08-21
docs: reconcile the README with the code, and regenerate the stale bash completion

mimi1vx:
A documentation pass over the README, which last moved 144 commits ago
(`f2798759`, spanning 26.1.2 through 26.3.0) and had drifted. Every claim was
checked against the code; only the ones that were actually wrong were touched.

## `docs(readme)` — three corrections

**The MCP build line did not work.** `mtui-mcp` declares no default features,
so `cargo run -p mtui-mcp -- --help` prints the *"built without the `mcp`
feature"* stub rather than the server's help. Every other doc —
`installation.md`, `developer.md`, `mcp.md`, `CONTRIBUTING.md`, `AGENTS.md` —
already carried the `--features mcp` form; the README was the sole outlier.
Both spellings were run to confirm which reaches the parser.

**`parallel`/`serial` host modes do not exist.** `TargetState` is
`enabled`/`disabled` only, no command exposes a serial flag, and the token
appears nowhere in the generated `cli.md`.

**The locking bullet named only `/var/lock/mtui.lock`.** Both lock files are
cross-process wire-format contracts, and the pool claim is what actually lets
concurrent testers share a fleet — the property the bullet claims. Wording
follows `locks.rs`.

## `chore(dist)` — regenerate the `mtui-mcp` bash completion

`cargo xtask gen` should be reproducible against the checked-in tree, and for
every other artifact it is: both man pages, the zsh and fish completions, and
the two generated mdBook pages all came back byte-identical. Only this file
did not — it predates a `clap_complete` change that mangles the binary's
hyphen when deriving the shell function name, so regenerating `dist/` always
surfaced an unrelated diff. No CI job checks generated-artifact freshness, so
this drifted silently; `dist/*` is shipped by `release.yml`, so it does reach
users.

The function is now `_mtui__mcp`; the definition and both `complete -F`
registrations move together, so completion behaves identically.

## Verification

- `cargo run -p mtui-mcp --features mcp -- --help` — reaches the parser; the
  bare form reproduces the stub.
- `cargo xtask gen-docs` + `cargo xtask gen` — tree now regenerates clean.
- `bash -n` on the completion; no stale `_mtui-mcp` reference remains.
- `cargo test -p xtask` — passes (its assertions pin generated *filenames*,
  not the function identifier).

No Rust source changed, so the compile/clippy/test gates are unaffected.

## Deliberately out of scope

- **`docs/src/**`** — actively maintained across the change set (`faq.md`
  covers the Ctrl-C contract #441, `mcp.md` the in-band `get`/`put` override
  #434 and forced-cancel lock release #405). Spot-checked `concepts.md`,
  `introduction.md`, `installation.md`; no drift found.
- **`updates` / openQA-`ruoqa` flag detail** — the README's coarse
  descriptions remain accurate; flag-level detail belongs in the generated
  reference, where it already is.
- **`vim-plugin` naming** — the README's phrasing versus the
  `mtui-vim-plugin` subpackage name is ambiguous rather than wrong.

## Changelog

No entry on either commit: documentation-only, and the completion
regeneration carries no user-visible behaviour change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ authenticated boundary trusted to operate the remaining maintenance tools.
 
 - Parallel SSH command execution across reference hosts (`run`, `update`,
   `install`, `prepare`, `downgrade`, …) with per-host `enabled`/`disabled`
-  states and `parallel`/`serial` modes. **Pubkey auth only.**
+  states. **Pubkey auth only.**
 - OBS/IBS and Gitea maintenance-request workflow (`assign`, `approve`, `reject`,
   `comment`, …) via the native OBS/IBS API (no `osc` subprocess).
 - Optional Slack review-request integration (`request_review`, off by default):
@@ -57,8 +57,10 @@ authenticated boundary trusted to operate the remaining maintenance tools.
   regeneration (`regenerate`).
 - Reference-host discovery via `refhosts.yml` (HTTPS- or filesystem-resolved,
   cached) and offline inventory search (`list_refhosts`).
-- Cooperative reference-host locking (`/var/lock/mtui.lock`) so concurrent
-  testers can share a fleet.
+- Cooperative reference-host locking so concurrent testers can share a fleet: a
+  PID-based operation lock (`/var/lock/mtui.lock`) serializing repository
+  transactions, and an RRID-based pool claim (`/var/lock/mtui-pool.lock`)
+  reserving a host for a template.
 - Test-report lifecycle: `load_template`, `checkout`, `commit`, `edit`, `export`
   (SVN and Gitea backends).
 - File transfer (`put`/`get`) over SFTP.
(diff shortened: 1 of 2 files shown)


plusky (approved):
Looks correct.

- `mtui-mcp` has no default features (`mcp` is optional), so the README's `cargo run -p mtui-mcp -- --help` was the odd one out versus AGENTS.md / installation.md / developer.md.
- `TargetState` is only `Enabled`/`Disabled`; `serial` is absent from `cli.md`.
- Both lock files are the wire-format contracts; naming only `mtui.lock` understated the pool claim.
- Bash completion: definition and both `complete -F` lines move to `_mtui__mcp` together. zsh staying `_mtui-mcp` is consistent with it already regenerating clean.

No CHANGELOG is right (docs + generated-artifact freshness, no behaviour change). Holding merge until CI finishes.

